PR #102, opened 2013-08-03
Gtk UI: Fix UnicodeDecodeError after downloads are finished (bug 1834)

pluton8:
When all downloads/syncs are finished, gPodder displays a summary of
downloaded episodes, having cut the titles if they are too long.
However, Russian descriptions are regular, non-unicode python strings,
and gPodder may cut only a part of a multi-byte UTF-8 sequence. It
causes an exception like this:
UnicodeDecodeError: 'utf8' codec can't decode byte 0xbe in position 51:
invalid start byte

This patch fixes that by converting the title to a unicode string if
it's not unicode.

This bug is similar to bug 1825, commit
e1ce9b0551d6380514d255ec03d3609c574860c8.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -1428,6 +1428,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         result = []
         for title in episode_list[:min(len(episode_list), max_episodes)]:
+            # Bug 1834: make sure title is a unicode string,
+            # so it may be cut correctly on UTF-8 char boundaries
+            title = util.convert_bytes(title)
             if len(title) > MAX_TITLE_LENGTH:
                 middle = (MAX_TITLE_LENGTH/2)-2
                 title = '%s...%s' % (title[0:middle], title[-middle:])


alexweissman:
I'm having a related problem in v3.9.0, but with the "sync device" feature - specifically, when generating the playlist.  See [bug 1368](https://bugs.gpodder.org/show_bug.cgi?id=1368).  It would appear the gPodder gets stuck when trying to write a UTF8 filename to the playlist file:

```
ERROR: Uncaught exception: Traceback (most recent call last):
  File "C:\Program Files (x86)\gPodder\src\gpodder\gtkui\desktop\sync.py", line 282, in check_free_space
    resume_sync([],[],None)
  File "C:\Program Files (x86)\gPodder\src\gpodder\gtkui\desktop\sync.py", line 193, in resume_sync
    playlist.write_m3u(episodes_for_playlist)
  File "C:\Program Files (x86)\gPodder\src\gpodder\gtkui\desktop\deviceplaylist.py", line 110, in write_m3u
    filename = self.get_absolute_filename_for_playlist(current_episode)
  File "C:\Program Files (x86)\gPodder\src\gpodder\gtkui\desktop\deviceplaylist.py", line 89, in get_absolute_filename_for_playlist
    filename = os.path.join(util.relpath(self.mountpoint, self._config.device_sync.device_folder), filename)
  File "C:\Python27\Lib\ntpath.py", line 84, in join
    result_path = result_path + p_path
UnicodeDecodeError: 'utf8' codec can't decode byte 0x97 in position 53: invalid start byte
```

PR #115, opened 2014-02-28
signature is in the url itself now, not as an extra parameter any longer

pieska:
youtube changed the API again but this time for the better. they now include the signature in the URL, before this change it was an extra parameter.


snehalparmar:
thanks a lot, it worked for me on my ubuntu 14.04

PR #126, opened 2015-02-10
Show covers directly after adding feeds (bug 1987)

niklasf:
When adding a new podcast feed a cover will get downloaded in the
background. Once it is finished gPodder.cover_download_finished() calls
PodcastListModel.add_cover_by_channel(), however by then a placeholder
cover will already be in the cache and the new cover will not be shown.

cover_download_finished() is the only caller of add_cover_by_channel(),
so generally clearing the cache when adding a new cover should be fine.

I reported this as https://bugs.gpodder.org/show_bug.cgi?id=1987, earlier.


thp:
Great, I had this issue also a few times, but never bothered to investigate. Good work!

PR #127, opened 2015-02-11
Detect The Escapist IP Bans

somini:
As I said in the mailing list, The Escapist banned my IP (but they lifted it a few days ago, 2 weeks tops), so I was seeing some strange results.

This PR fixes that issue by raising an error when the JSON they provide contains only ads. The ideal thing was to redirect the request through a proxy of some kind, because I'm pretty sure the MP4 link works.

### DIFF
--- a/src/gpodder/escapist_videos.py
+++ b/src/gpodder/escapist_videos.py
@@ -68,7 +68,7 @@ def get_real_download_url(url):
 
     data_config_url = data_config_frag.group(1)
 
-    logger.debug("Config URL: "+data_config_url)
+    logger.debug('Config URL: %s', data_config_url)
 
     data_config_data = util.urlopen(data_config_url).read().decode('utf-8')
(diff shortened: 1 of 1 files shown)


thp:
Hm, why did they block the IP? Maybe we shouldn't be trying to work around their restrictions?

src/gpodder/escapist_videos.py
```diff
@@ -68,6 +68,8 @@ def get_real_download_url(url):
 
     data_config_url = data_config_frag.group(1)
 
+    logger.debug("Config URL: "+data_config_url)
```
thp: If you need this log call, don't use string concatenation, but use:

```
logger.debug('Config URL: %s', data_config_url)
```

somini:
I fixed that logger call.

As for the IP ban, it's probably because I've been testing and made a little too much requests. As it stands it just detects the ban, doesn't really circumvents it in any form.

PR #128, opened 2015-05-17
Escapist: Fixes for compatibility with the new site version

somini:
This makes it work again for the April-May 2015 update.
They tried to do crypto in JS, what fools.


somini:
This requires urllib as a dependency, not sure how this affect packaging.

thp:
Merged, thanks. `urllib` is included in the standard library, so should be good to go.

PR #129, opened 2015-05-17
New Plugin: RTP Play

somini:
This allows to use gpodder to subscribe and download RTP programmes, the portuguese public TV. I've been running this branch as default for a couple months now and couldn't find any outstanding bugs.

I created this purely for myself, not sure if it's useful to anyone else but me, but it has a cache and parses HTML with the `lxml` package.

Both `lxml` and `urllib` are required, in case there's need to change the packaging somehow.


src/gpodder/util.py
```diff
@@ -218,6 +218,7 @@ def normalize_feed_url(url):
             # YouTube playlists. To get a list of playlists per-user, use:
             # https://gdata.youtube.com/feeds/api/users/<username>/playlists
             'ytpl:': 'http://gdata.youtube.com/feeds/api/playlists/%s',
+            'rtp:' : 'http://www.rtp.pt/play/podcast/%s'
```
thp: There's an RTP URL scheme, and why we don't yet support it (and may never support it), I'd rather not have this in here. Maybe "rtppt:" if you really want to have a shortcut? Or just leave it out.

somini: It went right over my head, I'll just remove it.

thp:
The file `src/gpodder/plugins/rtp_play.py` isn't PEP-8 compliant. Please use 4-spaces indentation and otherwise follow the guidelines in PEP-8 (one empty line between member function definitions, two empty lines between global class definitions, etc...). After that, this can be merged.

somini:
I only get `E501 line too long` errors now, that requires some (very much needed) general code clean up. I'll try to have it completed by the weekend.

src/gpodder/__init__.py
```diff
@@ -206,6 +206,7 @@ def fixup_home(old_home):
 # Plugins to load by default
 DEFAULT_PLUGINS = [
     'gpodder.plugins.soundcloud',
+    'gpodder.plugins.rtp_play',
```
thp: Don't load that by default - it depends on `lxml`, which we don't depend on.

thp:
No need to worry about line lengths, it's okay to have some longer lines. PEP-8-wise, it looks good now to me. As for loading as default: If you manage to remove the `lxml` dependency, we can load it as default. Maybe you can use [xml.dom.minidom](https://docs.python.org/2/library/xml.dom.minidom.html) or some module that's shipped with Python by default? Otherwise, we can just remove it from the list of plugins to load by default and add it as-is with the `lxml` dependency.

somini:
I only use it for the XPath really, should be easy to replace.

I'm pretty sure this is included in standard Python:
https://docs.python.org/2/library/xml.etree.elementtree.html#xpath-support
If so, I think I can work it out.

thp:
Yes, that's in the standard library. If you can port it to that, this would be great :)

somini:
I'll have a go this weekend.

Thanks for taking the time to babysit this PR. :D

somini:
After reading the docs and doing some experimenting, we're gonna need a bigger boat.

There are several parsers, but none are direct ports of `lxml`. The XML parsers have XPath, which should have made for an easy port, but they die on the ugly HTML I need to parse.
The `HTMLParser` is my best bet. It parses the pages correctly, but it's really low level.
I searched around and found this:
https://github.com/marmelo/python-htmlparser
It's for Python 3, but the API looks very similar. I'll dive into this this week.

thp:
For parsing bad HTML, my tool of choice so far has been BeautifulSoup.

thp:
Any updates here? :)

somini:
Not yet I'm afraid, I've been kinda busy.
Since it Just Works (TM) for me, it also lack the urgency of other updates. :D
I'll try to have something to show this weekend.

thp:
Yeah, no worries. Just want to kick this forward every now and then so that we can still get it merged eventually and not have it fall on the wayside ;)

somini:
Progress!
It can almost work, the "business logic" is done, but the parser chokes on the ugly HTML. I think the main problem is that it uses UTF-8 when it should use Latin1, but I can't see how to change that.

somini:
The kind of errors it spews:
`'utf8' codec can't decode byte 0xe1 in position 1: unexpected end of data`
On this page:
http://www.rtp.pt/play/bg_l_ep/?stamp=194877&listDate=&listQuery=&listProgram=1809&listcategory=&listchannel=&listtype=recent&page=6&type=all

thp:
Did you get around to look into the errors?

somini:
Not really, sorry.

I did found out that the page is being sent with non-UTF8 encoding and the HTTP reader is not picking up on that. I think I can investigate this further in the next few weeks.

thp:
Closing for now. Please feel free to reopen this pull request when you have updated it :)

PR #131, opened 2015-09-23
Added date to the pattern of "rename episode"

rybak:
Post download extension rename_download.py:
* Added date (in ISO 8601 format) to the new filename pattern.

Reasoning: if podcast maintainer doesn't add a sequenced number to the title, the episodes won't be in chronological order in the filesystem. ISO 8601 makes it easily sortable.

I reported this as feature request https://bugs.gpodder.org/show_bug.cgi?id=2031.

### DIFF
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 _ = gpodder.gettext
 
 __title__ = _('Rename episodes after download')
-__description__ = _('Rename episodes to "<Episode Title>.<ext>" on download')
+__description__ = _('Rename episodes to "<Date>-<Episode Title>.<ext>" on download')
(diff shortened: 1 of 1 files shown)


thp:
Hi, not sure if you were the other person requesting this as well, but I've now merged a similar thing in a slightly different way:

https://github.com/gpodder/gpodder/commit/bf31b911b6e9717ea3c7d42a36fd71f8899c193a

rybak:
Cool, thanks.
No, Rob is not me.

rybak:
Where will those settings for the plugin appear in the application?

thp:
The settings will appear in "Preferences", "Edit config" (button), then search for "rename_downloads" and there you should then see two checkboxes that you can use to enable. You might have to restart gPodder once after enabling the extension for the checkboxes to appear, but then enabling/disabling the checkboxes will instantly change these settings.

PR #132, opened 2015-11-28
Adding another pattern for youtube feed url list

daneoshiga:
I tried to use the Update Youtube Subscriptions but it didn't work with most of my feeds, looking around I noticed that them had a slightly different url, so I'm adding it here, it could be made part of the previous pattern but I believe that it's better to keep it separated to easier understading of each pattern.


thp:
Looks good; merged - thanks :)

PR #133, opened 2016-01-14
auto enqueue episodes

boyska:
(basically copied from `enqueue_in_mediaplayer.py`)

This allows the user to automatically enqueue episodes while they're being downloaded.


thp:
Instead of creating a completely new extension and duplicating lots of code, maybe you can add a configuration option to the original extension that would make the enqueueing after download opt-in?

thp:
Merged it into the original extension now, thanks for the suggestion!

boyska:
But isn't it counter intuitive that it is not in the post-download category? As a user, how do I understand that there is such a functionality?

thp:
True, we should probably change its name and put it into two categories or something.

boyska:
Is it possible? I understood that a single extension can be in just one category, which is why I created a different one.
I don't know if it's possible to share code between extensions.

thp:
Well not yet, we'd have to change the code that parses the categories to also make it available there. But that shouldn't be too hard to do?

PR #140, opened 2016-08-07
Use html5lib to find hyperlinks in the HTML of podcast show notes

Mortal:
Some podcasts use HTML hyperlinks in the show notes, but currently gPodder strips away all HTML, including the hyperlinks, leaving just uninteractive text.

This patch adds HTML hyperlink handling by parsing the HTML with html5lib and walking the DOM tree to produce a plain text string with markers to indicate where the hyperlinks are and what they point to. When the TextView rendering the show notes is clicked, the cursor position is used to determine if the user has clicked on a link, in which case the web browser is launched using the [webbrowser.open](https://docs.python.org/2/library/webbrowser.html) from the Python standard library.

The style of Visitor class (ExtractHyperlinkedText in the patch) should be familiar to anyone who has used the [ast.NodeVisitor class](https://docs.python.org/2/library/ast.html#ast.NodeVisitor) in the Python standard library.

### DIFF
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -17,6 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
+import webbrowser
 import gtk
 import gtk.gdk
 import gobject
@@ -114,19 +115,40 @@ class gPodderShownotesText(gPodderShownotes):
         self.text_view.set_wrap_mode(gtk.WRAP_WORD_CHAR)
         self.text_view.set_border_width(10)
         self.text_view.set_editable(False)
+        self.text_view.connect('button-release-event', self.on_button_release)
         self.text_buffer = gtk.TextBuffer()
         self.text_buffer.create_tag('heading', scale=pango.SCALE_LARGE, weight=pango.WEIGHT_BOLD)
         self.text_buffer.create_tag('subheading', scale=pango.SCALE_SMALL)
+        self.text_buffer.create_tag('hyperlink', foreground="#0000FF", underline=pango.UNDERLINE_SINGLE)
         self.text_view.set_buffer(self.text_buffer)
         self.text_view.modify_bg(gtk.STATE_NORMAL,
                 gtk.gdk.color_parse('#ffffff'))
         return self.text_view
 
     def update(self, heading, subheading, episode):
+        self.hyperlinks = [(0, None)]
         self.text_buffer.set_text('')
         self.text_buffer.insert_with_tags_by_name(self.text_buffer.get_end_iter(), heading, 'heading')
         self.text_buffer.insert_at_cursor('\n')
         self.text_buffer.insert_with_tags_by_name(self.text_buffer.get_end_iter(), subheading, 'subheading')
         self.text_buffer.insert_at_cursor('\n\n')
-        self.text_buffer.insert(self.text_buffer.get_end_iter(), util.remove_html_tags(episode.description))
+        for target, text in util.extract_hyperlinked_text(episode.description):
+            self.hyperlinks.append((self.text_buffer.get_char_count(), target))
+            if target:
+                self.text_buffer.insert_with_tags_by_name(
+                    self.text_buffer.get_end_iter(), text, 'hyperlink')
+            else:
+                self.text_buffer.insert(
+                    self.text_buffer.get_end_iter(), text)
+        self.hyperlinks.append((self.text_buffer.get_char_count(), None))
         self.text_buffer.place_cursor(self.text_buffer.get_start_iter())
+
+    def on_button_release(self, widget, event):
+        if event.button == 1:
+            pos = self.text_buffer.props.cursor_position
+            i = 0
+            while i+1 < len(self.hyperlinks) and self.hyperlinks[i+1][0] < pos:
+                i += 1
+            target = self.hyperlinks[i][1]
+            if target is not None:
+                webbrowser.open(target)
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -115,6 +115,7 @@ class gPodderShownotesText(gPodderShownotes):
         self.text_view.set_border_width(10)
         self.text_view.set_editable(False)
         self.text_view.connect('button-release-event', self.on_button_release)
+        self.text_view.connect('key-press-event', self.on_key_press)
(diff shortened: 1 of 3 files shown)


src/gpodder/gtkui/shownotes.py
```diff
@@ -114,19 +115,40 @@ def init(self):
...
+
+    def on_button_release(self, widget, event):
+        if event.button == 1:
+            pos = self.text_buffer.props.cursor_position
+            i = 0
+            while i+1 < len(self.hyperlinks) and self.hyperlinks[i+1][0] < pos:
+                i += 1
+            target = self.hyperlinks[i][1]
+            if target is not None:
+                webbrowser.open(target)
```
thp: Use `util.open_website()` instead of `webbrowser.open()` directly to avoid blocking the UI (depends on browser).

thp:
Can this be used to replace/extend `util.remove_html_tags()`, which already does some rudimentary HTML-to-plaintext transforms?

https://github.com/gpodder/gpodder/blob/master/src/gpodder/util.py#L585

Mortal:
I tried to make the output of `extract_hyperlinked_text` close to the output of `remove_html_tags`, and indeed it is easy to reimplement `remove_html_tags` in terms of `extract_hyperlinked_text`:

```python
def remove_html_tags(html):
    return ''.join(text for target, text in extract_hyperlinked_text(html))
```

However, I left the old `remove_html_tags` as is in case the user has not installed `html5lib`. If we make `html5lib` a hard dependency, then we can do it.

thp:
Do you think you could use one of the HTML parsing libraries included in the standard library?
https://docs.python.org/2.7/library/markup.html

Possibly HTMLParser:
https://docs.python.org/2.7/library/htmlparser.html#example-html-parser-application

Then we could use that instead of `html5lib`, and we can replace `remove_html_tags()` with your code.

Mortal:
I don't think the modules in the standard library parse HTML well enough (and a [StackOverflow answer](http://stackoverflow.com/a/2680724) backs up my gut feeling), and I think it is pretty important that HTML hyperlinks are located correctly. I'm 99% sure that html5lib will handle all show notes in the wild correctly (as it follows the same modern parsing standard as modern web browsers), but with the offerings of the standard library I would need to collect many samples to be just as certain.

thp:
The stackoverflow answer deals with explicitly "malformed HTML". I expect most feeds to have proper HTML. And if the HTML is malformed, it's okay for the links to be missing or wrong (undefined behavior territory), and html5lib would also only guess/apply heuristics (even if guessing the same way as browsers).

If it isn't too much work, go with HTMLParser from the standard library. Anything that isn't parsed correctly by it is probably also not parsed correctly by the current version of `remove_html_tags()`, so using your new method in combination with HTMLParser will only improve the current situation.

I agree that html5lib is probably more robust than HTMLParser. However, using features of the standard library when possible is helpful in this situation to avoid new dependencies.

If/when we run into shownotes' HTML that HTMLParser cannot parse, we can deal with that in the future.

Mortal:
While I'm not intimately familiar with the [WHATWG HTML parsing specification](https://html.spec.whatwg.org/multipage/syntax.html#parsing), I do believe it's written to be deterministic, meaning that syntactically incorrect HTML will be parsed in the same way by html5lib and modern web browsers -- no undefined behavior, guessing or heuristics. Consequently, any HTML that has a wrong meaning when parsed by html5lib (e.g. a link spanning the rest of the document) will have the same wrong meaning when parsed by a modern web browser, so there's a good chance the author of said HTML fixes the problem.

HTMLParser only tokenizes the HTML and does not actually build a DOM tree for the input, so in this regard it's only a little better than a regex-based solution such as `remove_html_tags()`. It should be easy to reimplement `remove_html_tags()` using HTMLParser and then add the link handling, but the resulting implementation will have little to nothing to do with the `html5lib`-based approach in my patch.

thp:
I'll give it some thought..

Mortal:
Here's an implementation that uses HTMLParser. I don't subscribe to many podcast feeds but the one I have that has a lot of links in the show notes (Hello Internet) seems to be handled well by both implementations.

thp:
Added some commits here, maybe you can cherry-pick them?
https://github.com/gpodder/gpodder/commits/html5lib-shownotes-hyperlinks

The commits are: https://github.com/gpodder/gpodder/commit/b61072311b6f804545c6342ca0181940660bfe35 https://github.com/gpodder/gpodder/commit/d4304be36c8f19d2fa9a50cfce746b2ab5db98ce https://github.com/gpodder/gpodder/commit/874494063a722b8212ee13baaec862dfdb1548bd

thp:
Of course, now the coolest thing would be if there would be a `ExtractHyperlinkedText` class implementing the logic (handle tag - create hyperlink data and handle stripping of whitespace, etc..) and two subclasses, one using `html5lib` and one using `HTMLParser` and switching between using those depending on whether html5lib is available or not.

This way, if html5lib is available/installed, it's used, but if it's not available, we fall back to HTMLParser to at least get some of that functionality (and users can opt to install html5lib manually to get better HTML parsing).

Mortal:
Interesting idea! I got carried away with a different project tonight, but I might find the time to implement that another evening this week.

thp:
Proposed solution for combining `HTMLParser` and `html5lib` parsing: https://github.com/gpodder/gpodder/commit/3c113a4bfab2e6dac1fa96a6634fe29f6dc0668b

This builds just enough of a document tree to use with the current parsing code we have, so the parser can handle both a document tree from `html5lib` or from `HTMLParser` (preferring `html5lib` if available).

thp:
Of course,  it would be cool if we could make it so that when hovering over a hyperlink the cursor turns into a pointer, but failing that, with https://github.com/gpodder/gpodder/commit/3c113a4bfab2e6dac1fa96a6634fe29f6dc0668b applied, this is ready to merge, what do you think?

Mortal:
I think it is better to translate the html5lib tree into start/end tags than to translate the HTMLParser start/end tags into a sort-of tree.

Mortal:
I have pushed [a first attempt](https://github.com/gpodder/gpodder/pull/140/commits/81cbe1fba70342e23290c4bd93a5b7ef37c9ad9e), but as I don't have more time now I'll have to test it later.

thp:
I'll give it a try soon. So far it looks good, so hopefully once I've tested it it can be merged :)

thp:
Merged, thanks!

PR #150, opened 2016-10-08
Convert gPodder gtkui to Gtk3

elelay:
Target gtk would be 3.10 (maybe 3.14, but even Debian Jessie is at 3.14).
Everything I've tested works. The ui looks very close to gtk2.
Still expect unicode decode errors (with non ASCII filenames, maybe).
I'll give a link to an OSX build later.

Main changes:
 - pygtk to Gtk,Gdk,... changes (like pack_start() argument count change)
 - GtkTable to GtkGrid when GtkTable looked bad, Gtk{H,V}{Box,Paned,Scale} to Gtk{Box,Paned,Scale}
 - UIManager to Gio.Menu (no need for GTKOSXApplication anymore)


thp:
Looks good, thanks for your hard work! I'll merge this into the `gtk3-new` branch and see what's left to do.

vossad01:
One minor usability issue with this concerns selection of text in the "Episode details" frame.

Under the gtk2 implementation, selection has causes the text to turn white and have a dark-colored background, preserving sufficient contrast to read what is highlighted.

Under the gtk3 implementation, selection still causes the text to turn white (or light-gray), but the background does not change.  This effectively causes the selected text to disappear due to insufficient contract.

I suppose this could be a theme issue.  I am just using the default theme from Linux Mint Cinnamon.

I still consider the application usable and am grateful for the change.  However, I imagine not many users are running this PR already so thought I would report this.

thp:
Thanks for the fix @elelay and thanks for testing this branch @vossad01 -- let us know if you find any other issues.

vossad01:
I'm fine just commenting here, but let me know if you want me creating new issues.

The "Gtk Status Icon" under Extensions > Desktop integration does not work on the gtk3 branch.

> Extension cannot be activated

> 'gi.repository.Gtk' object has no attribute 'status_icon_new_from_icon_name'

When checking that was a regression, I noticed that all of the menus are text only on the gtk3 branch, the images previously associated with the menu items are not displayed.  Clearly not something I had missed, but I am sure many like the icons.

elelay:
Good catch!
I've not ported it yet. Gtk.StatusIcon is [deprecated](http://lazka.github.io/pgi-docs/#Gtk-3.0/classes/StatusIcon.html#Gtk.StatusIcon.new) since Gtk 3.14, but I don't see what to replace it with (notifications are not the same thing!).

About icons in menus, apparently they're out of fashion. I've kept some in the context menu, but removed them in menus. They can be added back I think...

Please open an issue "Gtk3 migration issues" for next bugs (always an optimist ;-) ).

PR #157, opened 2016-11-05
GtkBox to GtkGrid in gpodder.ui

vossad01:
There were two or three instances where the deprecated `GtkVBox` or `GtkHBox` was being used in **gpodder.ui**.  `GtkBox` could have been used, though the documentation indicates that, eventually it will be deprecated as well, so `GtkGrid` is more future proof.

> If you don’t need first-child or last-child styling, and want your code to be future-proof, the recommendation is to switch to GtkGrid instead of nested boxes. For more information about migrating to GtkGrid, see Migrating from other containers to GtkGrid.

https://developer.gnome.org/gtk3/stable/GtkHBox.html

While I was at it I just changed all instances within **gpodder.ui** to `GtkGrid` as it seemed like as good of time as any to do so.  If you would rather just have the `GtkVBox` and `GtkHBox` instances changed, I can easily provide a different pull request.

Also added a .gitignore file to help avoid accidentally checking in *.pyc files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.pyc
+src/feedparser.py
+src/mygpoclient
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -30,9 +30,9 @@
     <property name="urgency_hint">False</property>
     <signal handler="on_gPodder_delete_event" name="delete-event"/>
     <child>
-      <object class="GtkVBox" id="vMain">
+      <object class="GtkGrid" id="vMain">
         <property name="visible">True</property>
-        <property name="homogeneous">False</property>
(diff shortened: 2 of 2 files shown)


thp:
Please only create a minimal `.gitignore` file and don't just import a big generic gitignore file from somewhere.

thp:
The commit https://github.com/gpodder/gpodder/pull/157/commits/b7bf5f36998e806b4d98b51c0626a8d1098fb45d looks good, though.

vossad01:
Thanks, done.  Force-pushed new commits.

If you ever do need a .gitignore file for something, I have always found those maintained at https://github.com/github/gitignore to be of a high quality.  They are what GitHub sources if you make a repo through its UI and I have seen them used by other programs providing scaffolding.

thp:
Thanks for the quick update :) Merged!

PR #167, opened 2016-11-15
Allow reordering of the download queue

vossad01:
gPodder does not allow reordering of the download queue. If a user wants to prioritize the download of a particular episode, they must do one of the following:

 - Queue the downloads in the desired download order
    - This is tedious and does not work if the user wants to use the "Download new episodes" dialog
 - Pause the downloads and resume them in the desired order
    - This is tedious and the set download order is not visible to the user.
 - Force start the desired priority downloads
    - This goes around the concurrent download limit so does not scale well.

This provides the following behavior changes in an effort to provide a better way for users to prioritize or de-prioritze a download:
 - Allows a user to reorder the download queue via the context menu (allows reordering a selection) or by clicking and dragging a row (only one row at a time, NB. on my system, GTK seems to make it really easy to accidentally do a multi-select rather than a click-and-drag for reorder on a row).
 - Makes gPodder respect the queue order in the UI when starting downloads.

### DIFF
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -371,7 +371,7 @@
                             <property name="headers_visible">False</property>
                             <property name="rules_hint">False</property>
                             <property name="rubber-banding">True</property>
-                            <property name="reorderable">False</property>
+                            <property name="reorderable">True</property>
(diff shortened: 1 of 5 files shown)


vossad01:
Found an issue with this.  Will revise.

PR #177, opened 2016-11-26
remove pre-python 2.6 compat code

elelay:
Also removed unused cgi imports.

By the way, cgi.escape is deprecated in favor of html.escape, because it escapes quotes also, to prevent html injection in attributes. None of our uses is affected, but I'd recommend moving to html.escape. Thoughts?


thp:
Yes, let's go with `html.escape()`

Issue #180, opened 2016-11-30
Pagination in Soundcloud


mprangenberg:
gPodder currently only supports the Standard-API Limit of 50 tracks. Could you maybe raise that to the maximum of 200? I follow some podcasts on there, that are already at over 150 tracks. Even better would be to use pagination to always get all tracks: ttps://developers.soundcloud.com/docs/api/guide#pagination
Sadly i can't do it myself as i am a total noob to Python right now.
Cheers.

thp:
Pull request merged, thanks a lot!

Issue #179, opened 2016-11-29
Allow user to configure API key for Sound Cloud


f03el:
Sometimes when trying to play or download an episode, nothing happens (at least in the Sailfish OS version). If running from the terminal I see that Sound Cloud complains about "Too many requests". The API key doesn't seem to allow enough requests. Would it be possible to ask them for a higher limit?

thp:
Wonder if it would help if the users have to enter their own API keys similar to what we do with YouTube?

elelay:
Soundcloud doesn't let one register apps anymore.

PR #184, opened 2016-12-06
Update SoundCloud-Feeds with Pagination

mprangenberg:
This uses Pagination as per the SoundCloud API: https://developers.soundcloud.com/docs/api/guide#pagination to get all episodes of a feed. See Issue #180.
I don't know if this needs to respect the max_episodes_per_feed-setting. If so, could someone point me to how i can get the value of that setting?

### DIFF
--- a/src/gpodder/plugins/soundcloud.py
+++ b/src/gpodder/plugins/soundcloud.py
@@ -136,7 +136,7 @@ class SoundcloudUser(object):
             json_url = 'http://api.soundcloud.com/users/%(user)s/%(feed)s.json?filter=downloadable&consumer_key=%(consumer_key)s&limit=50&linked_partitioning=1' \
                     % { "user":self.username, "feed":feed, "consumer_key": CONSUMER_KEY }
 
-            while not json_url == '':
+            while json_url != '':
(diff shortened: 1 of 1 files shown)


src/gpodder/plugins/soundcloud.py
```diff
@@ -133,33 +133,37 @@ def get_tracks(self, feed):
...
-                    'link': track.get('permalink_url') or 'http://soundcloud.com/'+self.username,
-                    'description': track.get('description') or _('No description available'),
-                    'url': url,
-                    'file_size': int(filesize),
-                    'mime_type': filetype,
-                    'guid': track.get('permalink', track.get('id')),
-                    'published': soundcloud_parsedate(track.get('created_at', None)),
-                }
+
+            while not json_url == '':
```
thp: `while json_url != '':`

thp:
Holding off merging this for now, since it increases the API requests. We might want to let the user provide their soundcloud API key like we do for YouTube so we don't hit API request limits.

vossad01:
Depends on #179

elelay:
Soundcloud doesn't let one register applications anymore.
So I'm rejecting this PR as it could increase API requests.

PR #217, opened 2017-01-29
Fix unicode renaming problems

matoro:
I'm assuming that this is not the correct way to fix this, but I'm submitting it as a pull request rather than an issue because this was the hack I found on stackoverflow that fixed the problem for me.  The download renaming extension fails >75% of the time due to unicode errors, even when there are no obviously unusual characters.  The traceback follows the format:

```
[gpodder.extensions] ERROR: Error in /usr/share/gpodder/extensions/rename_download.py in on_episode_downloaded: 'ascii' codec can't decode byte 0xe2 in position 11: ordinal not in range(128)
  File "/usr/lib/python2.7/site-packages/gpodder/extensions.py", line 83, in handler
    cb_res = callback(*args, **kwargs)
  File "/usr/share/gpodder/extensions/rename_download.py", line 38, in on_episode_downloaded
    episode.sortdate, episode.channel.title)
  File "/usr/share/gpodder/extensions/rename_download.py", line 56, in make_filename
    new_basename = ' - '.join(new_basename)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 11: ordinal not in range(128)
```

Again, I know minimal python, but I do know that this technically fixes the problem, so if this PR could be done the right way and added it would be greatly appreciated.

### DIFF
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -3,6 +3,10 @@
 # Copyright (c) 2011-04-04 Thomas Perl <thp.io>
 # Licensed under the same terms as gPodder itself
 
+import sys
+reload(sys)
+sys.setdefaultencoding("utf-8")
+
 import os
 
 import gpodder


thp:
The proper fix is probably to set your environment variables / local to an UTF-8 local, then Python should pick it up and work correctly. Here are some Wiki pages how to set it up for some Linux distros:

https://help.ubuntu.com/community/Locale
https://wiki.archlinux.org/index.php/locale
https://docs.fedoraproject.org/en-US/Fedora/22/html/System_Administrators_Guide/ch-System_Locale_and_Keyboard_Configuration.html

What does `sys.getdefaultencoding()` give you before you configure your system with the correct locale setting?

thp:
And/or have you tried replacing `' - '` with `u' - '` in the code snippet above? (`new_basename = u' - '.join(new_basename)`)

matoro:
Hmm, on python 3 it gives me ```'utf-8'``` but on python2 it gives me ```'ascii'```.

Using ```u'-'``` does not fix the problem.

thp:
On the system where you get `'ascii'` for Python 2, what is the output of the following command (in the same session/shell that you open gPodder):

```
% set | egrep -a '^(LC_.*|LANG)=.*'
```

matoro:
```
LANG=en_US.UTF-8
```

thp:
Please try the following:

```
export LANG= en_US.UTF-8
export LC_ALL= en_US.UTF-8
gpodder --verbose
```

Then try to reproduce the problem.

thp:
@matoro Please check my last comment.

matoro:
@thp Sorry for not seeing this earlier, I will try this solution and see if it works.  I was able to get one episode renamed fine but it was from a feed from which I had not had problems in the past.  From what I remember, the issue primarily affected particular feeds, though I couldn't find any pattern to them like non-ASCII characters (as far as I could tell, every character was fairly normal).  I will post details here if I encounter another issue using the original ```rename_download.py``` and the verbose option.  If I see nothing after a week or so, I'll close this PR and revert my changes locally.

matoro:
@thp I was able to capture a verbose log of the issue occurring.  The env variables you specified are set.

```
STEAL THIS SHOW
Downloading ‘How The Swarm Will Beat The Cloud’, with Storj.io’s Shawn...1488055417.715010 [gpodder.model] INFO: Setting download filename: s02e11_How_The_Swarm_Will_Beat_The_Cloud_with_Storj.mp3
Downloading ‘How The Swarm Will Beat The Cloud’, with Storj.io’s Shawn...[100%]1488055485.755631 [gpodder.download] INFO: Download was redirected (http://content.blubrry.com/stealthisshow/s02e11_How_The_Swarm_Will_Beat_The_Cloud_with_Storj.mp3). New filename: s02e11_How_The_Swarm_Will_Beat_The_Cloud_with_Storj.mp3
1488055485.800977 [gpodder.extensions] ERROR: Error in /usr/share/gpodder/extensions/rename_download.py in on_episode_downloaded: 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
  File "/usr/lib/python2.7/site-packages/gpodder/extensions.py", line 83, in handler
    cb_res = callback(*args, **kwargs)
  File "/usr/share/gpodder/extensions/rename_download.py", line 38, in on_episode_downloaded
    episode.sortdate, episode.channel.title)
  File "/usr/share/gpodder/extensions/rename_download.py", line 56, in make_filename
    new_basename = ' - '.join(new_basename)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
Downloading ‘How The Swarm Will Beat The Cloud’, with Storj.io’s Shawn...[DONE]
```

The rest of the log was normal.

thp:
What about the following patch?

```
diff --git a/share/gpodder/extensions/rename_download.py b/share/gpodder/extensions/rename_download.py
index 8f2af405..68a3cff6 100644
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -48,12 +48,12 @@ class gPodderExtension:
         basename, ext = os.path.splitext(filename)
 
         new_basename = []
-        new_basename.append(util.sanitize_encoding(title) + ext)
-        if self.config.add_podcast_title:
-            new_basename.insert(0, podcast_title)
+        new_basename.append(util.to_unicode(title) + util.to_unicode(ext))
         if self.config.add_sortdate:
-            new_basename.insert(0, sortdate)
-        new_basename = ' - '.join(new_basename)
+            new_basename.insert(0, util.to_unicode(sortdate))
+        if self.config.add_podcast_title:
+            new_basename.insert(0, util.to_unicode(podcast_title))
+        new_basename = u' - '.join(new_basename)
 
         # On Windows, force ASCII encoding for filenames (bug 1724)
         new_basename = util.sanitize_filename(new_basename,
diff --git a/src/gpodder/util.py b/src/gpodder/util.py
index 7f3ecbcc..e7890c7f 100644
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1469,6 +1469,17 @@ def sanitize_encoding(filename):
     return filename.encode(encoding, 'ignore')
 
 
+def to_unicode(s):
+    """Convert a string or unicode object to unicode
+
+    >>> to_unicode('hello')
+    u'hello'
+    >>> to_unicode(u'world')
+    u'world'
+    """
+    return s if isinstance(s, unicode) else s.decode(encoding, 'ignore')
+
+
 def sanitize_filename(filename, max_length=0, use_ascii=False):
     """
     Generate a sanitized version of a filename that can
```

matoro:
@thp I will test this, only a minority of renames seem to have the issue and it is difficult to reproduce reliably but I will update with results.

matoro:
@thp Scratch that, had an issue right off the bat.  Traceback:

```
1492578451.939707 [gpodder.download] ERROR: Download failed: 'ascii' codec can't decode byte 0xe3 in position 42: ordinal not in range(128)
  File "/usr/lib/python2.7/site-packages/gpodder/download.py", line 820, in run
    self.filename = self.__episode.local_filename(create=True, force_update=True)
  File "/usr/lib/python2.7/site-packages/gpodder/model.py", line 487, in local_filename
    old_file_name = os.path.join(self.channel.save_dir, self.download_filename)
  File "/usr/lib/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 42: ordinal not in range(128)
```

elelay:
Forcing system encoding to utf8 is not acceptable.

PR #219, opened 2017-01-29
python 3.6 SQLite compat vacuum in transaction

elelay:
fixes a blocking exception on exit:
sqlite3.OperationalError: cannot VACUUM from within a transaction.
Solution comes from https://bugs.python.org/issue28518


thp:
Thanks!

Issue #235, opened 2017-02-27
Soundcloud feeds returning 404


christofdamian:
I think something must have changed at soundcloud. 
Gpodder shows a 404 error for my soundcloud feeds.
If you look up the feeds with http://getrssfeed.com/ they are also either empty or stop at the end of 2016.

![gpodder-soundcloud-errors](https://cloud.githubusercontent.com/assets/175254/23370102/e4964f78-fd13-11e6-8d71-c83cdf07243a.png)

one example: http://soundcloud.com/tanith
the getrssfeed feed for this is: http://feeds.soundcloud.com/users/soundcloud:users:3833/sounds.rss

Strubbl:
maybe this has something to do with soundcloud redirecting to https from http.

vatterspun:
Also seeing this with gPodder Portable v.3.9.2

vossad01:
I fixed the https issue locally, but it seems the SoundCloud API may also have changed.

gPodder is using a `consumer_key` to identify itself, while now they want a [`client_id`](https://developers.soundcloud.com/docs/api/introduction) which may require a new registration.  Or at least I don't see mention of `consumer_key` in the documentation.  [This may be a related StackOverflow question about the change](http://stackoverflow.com/q/34023373/1072626).

Here is an example url produced by gPodder that is now failing:

> https://api.soundcloud.com/users/geedee215/tracks.json?filter=downloadable&consumer_key=zrweghtEtnZLpXf3mlm8mQ&limit=200

I am not likely to look into this further since I do not have any podcasts that give me updates via SoundCloud.

vatterspun:
@vossad01 - thanks for looking into that

vossad01:
Thanks to @christofdamian this is now fixed in `master`.  If you need it before the next release, [you can run gPodder from git](https://github.com/gpodder/gpodder/wiki/Run-from-Git).

christofdamian:
You are welcome! Looking forward to the next release.

PR #240, opened 2017-03-06
Fix soundcloud feeds by using user id for track lists

christofdamian:
This fixes at least part of the issue #235 

There seems to be a bug in the soundcloud api that you have to use the numeric user id and not the user name: http://stackoverflow.com/questions/42214138/soundcloud-users-id-tracks-url-not-working

Excuse my python skills.


vossad01:
Looks good to me.  This resolved #235 in my testing.

PR #242, opened 2017-03-10
Use https urls for SoundCloud requests

vossad01:
SoundCloud has switched to 'https' by default.  'http' urls currently redirect to 'https' so gPodder continues to function.  This changes gPodder to make https requests directly.

This doesn't fix anything but does avoid the extra network traffic related to following the redirect.


thp:
This needs to be rebased now that #240 is merged, then it can go in.

PR #246, opened 2017-03-25
store episode's description_html (schema change)

elelay:
This way we can display the `description` in one-line description and `description_html` in show notes.

This is a schema change. I hope I've not forgotten any use of description.

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -149,10 +149,9 @@ class PodcastEpisode(PodcastModelObject):
             episode.description_html = entry['description_html']
         # XXX: That's not a very well-informed heuristic to check
         # if the description already contains HTML. Better ideas?
+        # TODO: This really should be handled in podcastparser and not here.
         elif '<' in entry['description']:
             episode.description_html = entry['description']
-        else:
-            episode.description_html = entry['description'].replace('\n', '<br>')
         episode.total_time = entry['total_time']
         episode.published = entry['published']
         episode.payment_url = entry['payment_url']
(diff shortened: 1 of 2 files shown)


elelay:
Remember to merge in gtk3 (apply 2to3 fix also)

src/gpodder/model.py
```diff
@@ -145,6 +145,14 @@ def from_podcastparser_entry(cls, entry, channel):
...
         episode.link = entry['link']
         episode.description = entry['description']
+        if entry.get('description_html'):
+            episode.description_html = entry['description_html']
+        # XXX: That's not a very well-informed heuristic to check
+        # if the description already contains HTML. Better ideas?
+        elif '<' in entry['description']:
+            episode.description_html = entry['description']
+        else:
+            episode.description_html = entry['description'].replace('\n', '<br>')
```
vossad01: Do I understand correctly that this will result in the description for every episode being stored twice?  Once under `description` and once under `description_html`?

elelay: Yes. Some podcasts provide an short (1 paragraph) description and a longer html_description with complete show notes and links. It seems bad to ignore their hand-crafted description and put a substring of the html_description when we can keep both.

vossad01: I understand storing both when they are separate things.  When we have only one, I was surprised that both get stored.  Other than that, this PR looks good.  You have the big picture on what this is for, if that storage decision is the right one I'd say lets get this merged.

elelay: ok. I can do better ;-)

vossad01: Thanks :)

PR #247, opened 2017-03-25
store episode's description_html (schema change)

elelay:
same as #246, but for gtk3 branch

### DIFF
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -201,7 +201,14 @@ class gPodderShownotesHTML(gPodderShownotes):
         else:
             self._base_uri = episode.channel.url
         self._loaded = False
-        self.html_view.load_html(episode.description_html, self._base_uri)
+
+        description = episode.description_html
+        if not description:
+            description = self.coerce_text_to_html(episode.description)
+        self.html_view.load_html(description, self._base_uri)
+
+    def coerce_text_to_html(self, text):
+        return text.replace('\n', '<br>')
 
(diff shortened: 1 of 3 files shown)


vossad01:
I understand how you arrived at having separate commits for the `gtk3` and `master` branch and I think it is a rational approach.  However, I am concerned about the risk of `gtk3` and `master` diverging, and this example of two independent commits, one for each branch, suggests that could be happening.

I think maintainability would be better if we had one commit that could be merged into both branches.  (Two PRs would still be welcome [especially in the case of conflicts], but they just need to share a common ancestor related to the change).

To try to facilitate this, I have opened PR #251.  Do you think that, if that is merged, you could close this PR and have 1a4b381 merged into `gtk3` instead?

elelay:
I agree with your proposal in #251 and have applied it.
It suggested following change:
```
-        for k, v in d.items():
+        for k, v in list(d.items()):
```
which seems ok.

So I'm closing this PR and adding a note to merge it in gtk3 in the other one...

vossad01:
Found out that I could push commits to this after I reopened it.  I tried before opening elelay/gpodder#2 but it failed, so I had opened that PR.

I did a force push of #246 being merged into the `gtk3` branch remedying https://github.com/gpodder/gpodder/pull/247#issuecomment-289258433.

Still outstanding would be changes to remove the coercion to HTML and try using `self.html_view.load_plain_text` instead, as recommenced by @elelay in https://github.com/elelay/gpodder/pull/2#pullrequestreview-31694896.

PR #251, opened 2017-03-26
Bring `gtk3` branch up to date with `master`

vossad01:
Thomas has occasionally has been merging `master` into `gtk3`; however, at any given point in time there are commits there are likely to be commits in `master` that have not been merged into `gtk3`.

The goal of this PR is three-fold:
 1. Bring `gtk3` up-to-date with master
 2. Establish merging process that prevents things in `master` being absent in `gtk3`
 3. Not place the burden keeping the `gtk3` branch up to date on a single maintainer

To this end, I utilized [`2to3`](https://docs.python.org/2/library/2to3.html) to establish a baseline of Python 3 compatibility.  This is certainly will not catch 100% of issues, but it is at least a bit of a safety to help in preventing incompatible Python 2 code from being merged into Python 3.  (I know there are some projects out there aiming to target a common subset of Python 2 and Python 3 so to be able to run under either, but I think we are past that point.)

I then merged every unmerged commit existing on `master` into `gtk3` (hopefully not screwing anything up along the way) and checked the result with `2to3` to ensure the merge did not introduce new compatibilities.

## Proposal going forward
1. Pull requests directly into `gtk3` do not need to backported to `master`
    They can is someone wants to.  This establishes gtk3/python3 as the eventual successor.  This has been the current status quo.
2. After accepting a Pull Request into `master`, use the following process to merge it into `gtk3` (this assumes all branches are up to date
    1. ```git checkout gtk3```
    2. ```git merge --no-commit master```
    3. for every FILE modified: ```2to3 FILE```
    4. If any of the diff suggestions seem relevant/correct, make the correction.
    5. ```git commit```
        indicate what was merged
    6. Verify the fix the same way as you did for the original merge into master
    7. Push the merge to GitHub

@thp @elelay Thoughts?


elelay:
Changed code looks good to me.
Proposal also looks good. I'll try it on store_html_description as you suggest.
Thanks!

PR #257, opened 2017-03-29
Propagate "old" state to other gpodder.net clients

bestlibre:
Unchecking new status in UI trig a 'mygpo_client.on_playback_full' to propagate the read state to other gpotter.net clients

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2686,15 +2686,17 @@ class gPodder(BuilderWidget, dbus.service.Object):
     def mark_selected_episodes_new(self):
         for episode in self.get_selected_episodes():
             episode.mark_new()
+        if self.mygpo_client.can_access_webservice():
+            self.mygpo_client.on_mark_new(self.get_selected_episodes())
+            self.mygpo_client.flush()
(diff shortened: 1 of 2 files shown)


elelay:
This has the advantage of marking episodes as played in AntennaPod but the drawback of marking the episode as fully played in gPodder.
We would rather not send a full played event, but unfortunately, AntennaPod [ignores 'played' events](https://github.com/AntennaPod/AntennaPod/blob/27bbe3ae80d9773a53e8e56dbb8990a8003489d7/core/src/main/java/de/danoeh/antennapod/core/gpoddernet/model/GpodnetEpisodeAction.java#L108) when position is not given or is 0.
What I'd like would be an implementation of the 'new' action, really corresponding to checking or unchecking the 'New' checkbox.
@stefankoegl do you remember what the 'new' action is for and why there is no way to sync 'mark as old'  / 'mark as new'?

vossad01:
@bestlibre Are you interested in pursuing this with gpodder.net/mygpo?

adamvoss:
Closing due to inactivity.  @bestlibre, please open a new pull request if more progress is made on this.  Thanks!

PR #258, opened 2017-04-01
python3 mpris listener fix

elelay:
I noticed exceptions when activating mpris-listener and playing an episode in VLC.
It now works :-)

### DIFF
--- a/share/gpodder/extensions/mpris-listener.py
+++ b/share/gpodder/extensions/mpris-listener.py
@@ -42,7 +42,7 @@ TrackInfo = collections.namedtuple('TrackInfo',
                         ['uri', 'length', 'status', 'pos', 'rate'])
 
 def subsecond_difference(usec1, usec2):
-    return abs(usec1 - usec2) < USECS_IN_SEC
+    return usec1 is not None and usec2 is not None and abs(usec1 - usec2) < USECS_IN_SEC
     
 class CurrentTrackTracker(object):
     '''An instance of this class is responsible for tracking the state of the
@@ -124,6 +124,7 @@ class CurrentTrackTracker(object):
                 self.notify_stop()
 
             if (    (kwargs['pos']) == 0
+                and self.pos is not None
                 and self.pos > (self.length - USECS_IN_SEC)
                 and self.pos < (self.length + 2 * USECS_IN_SEC)
             ):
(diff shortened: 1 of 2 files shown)


vossad01:
I confirmed this happening and this indeed would seem to prevent the exception.  Do you have any idea or did you look at all into why this occurs?  From your commit message, should I read that that you found it is just a timing thing?

This change seems valid because the code elsewhere shows that `pos` can be `None`.  My concern is that, since this was not occurring under Python 2, there could be an underlying issue on Python 3 that could pop up elsewhere and this only works around.  Since this extension is not enabled by default it gets less testing than other components.

elelay:
My understanding is that it is python3 being more picky when subtracting or comparing with `None`
```
~ ❯ python2
Python 2.7.13 (default, Dec 21 2016, 07:16:46) 
[GCC 6.2.1 20160830] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> None < 0
True
>>> 
~ ❯ python3
Python 3.6.0 (default, Jan 16 2017, 12:12:55) 
[GCC 6.3.1 20170109] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> None < 0
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'NoneType' and 'int'
>>> 
```

vossad01:
Of course, despite seeing the exception yesterday, now I am not reproducing it under Python 3 either :confused: 

I had been running with a debugger and I wasn't hitting the condition of either argument being `None` under **master** but was under **gtk3**.  The place this change protects is related to subtraction which fails under both Python 2 and 3:

``` shell
$ python2
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 5 - None
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
>>> None - 5
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

The difference you point out makes me wonder if the underlying issue isn't the inequality checks in `update`, but since trying to reproduce this is failing for me now, I'm not sure there is much more that can be done.

PR #262, opened 2017-04-01
extension: entry in context-menu to resume playback

elelay:
Uses the MPRIS2 dbus interface to enqueue and seek the episode
in supported media players.

### DIFF
--- a/share/gpodder/extensions/enqueue_in_mediaplayer.py
+++ b/share/gpodder/extensions/enqueue_in_mediaplayer.py
@@ -97,7 +97,7 @@ class MPRISResumer(FreeDesktopPlayer):
         self.title = '/'.join((_('Resume in'), application))
         self.bus_name = bus_name
         self.player = None
-        self.position = None
+        self.position_us = None
         self.url = None
 
     def is_installed(self):
@@ -122,13 +122,12 @@ class MPRISResumer(FreeDesktopPlayer):
             return False
 
         self.player = bus.get_object(self.bus_name, self.OBJECT_PLAYER)
-        self.signal_match = self.player.connect_to_signal(
-            self.SIGNAL_PROP_CHANGE,
+        self.signal_match = self.player.connect_to_signal(self.SIGNAL_PROP_CHANGE,
             self.on_prop_change,
             dbus_interface=self.INTERFACE_PROPS)
         return True
 
-    def wait_for_player(self, filename, pos):
+    def prepare_wait_for_player(self, filename, pos):
         def name_owner_changed(name, old_owner, new_owner):
             logger.debug('name_owner_changed "%s" "%s" "%s"',
                          name, old_owner, new_owner)
@@ -137,14 +136,9 @@ class MPRISResumer(FreeDesktopPlayer):
                 cancel.remove()
                 util.idle_add(lambda: self.do_enqueue(filename, pos))
 
-        logger.debug('waiting for player %s to enqueue %s',
-                     self.bus_name, filename)
         bus = gpodder.dbus_session_bus
         obj = bus.get_object(self.NAME_DBUS, self.OBJECT_DBUS)
-        cancel = obj.connect_to_signal(
-            'NameOwnerChanged',
-            name_owner_changed,
-            dbus_interface=self.NAME_DBUS)
+        cancel = obj.connect_to_signal('NameOwnerChanged', name_owner_changed, dbus_interface=self.NAME_DBUS)
 
     def do_enqueue(self, filename, pos):
         def on_reply():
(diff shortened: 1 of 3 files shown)


share/gpodder/extensions/enqueue_in_mediaplayer.py
```diff
@@ -81,6 +81,122 @@ def open_files(self, filenames):
...
+
+        def on_error(exception):
+            logger.error('MPRISResumer error %s', repr(exception))
+            self.signal_match.remove()
+
+        if filename.startswith('/'):
+            try:
+                import pathlib
+                self.url = pathlib.Path(filename).as_uri()
+            except:
```
thp: Is the `except` for `ImportError` (Python 2 and no `pathlib`)? If so, please make it explicit, so that we don't unnecessarily hide other exceptions.

share/gpodder/extensions/enqueue_in_mediaplayer.py
```diff
@@ -81,6 +81,122 @@ def open_files(self, filenames):
...
+    INTERFACE_PROPS = 'org.freedesktop.DBus.Properties'
+    SIGNAL_PROP_CHANGE = 'PropertiesChanged'
+    NAME_DBUS = 'org.freedesktop.DBus'
+
+    def __init__(self, slug, application, command, bus_name):
+        super(MPRISResumer, self).__init__(slug, application, command)
+        self.title = '/'.join((_('Resume in'), application))
+        self.bus_name = bus_name
+        self.player = None
+        self.position = None
```
thp: To make it more obvious that the position is measured in microseconds (according to your comment below), rename this variable to `position_us`.

share/gpodder/extensions/enqueue_in_mediaplayer.py
```diff
@@ -81,6 +81,122 @@ def open_files(self, filenames):
...
+            if name == self.bus_name:
+                logger.debug('MPRISResumer player %s is there', name)
+                cancel.remove()
+                util.idle_add(lambda: self.do_enqueue(filename, pos))
+
+        logger.debug('waiting for player %s to enqueue %s',
+                     self.bus_name, filename)
+        bus = gpodder.dbus_session_bus
+        obj = bus.get_object(self.NAME_DBUS, self.OBJECT_DBUS)
+        cancel = obj.connect_to_signal(
```
thp: The parameters to `connect_to_signal()` should fit into a single line.

share/gpodder/extensions/enqueue_in_mediaplayer.py
```diff
@@ -81,6 +81,122 @@ def open_files(self, filenames):
...
+
+    def on_prop_change(self, interface, props, invalidated_props):
+        def on_reply():
+            logger.debug('MPRISResumer has SetPosition for %s', self.url)
+
+        def on_error(exception):
+            logger.error('MPRISResumer SetPosition error %s', repr(exception))
+            self.signal_match.remove()
+
+        url = props.get('Metadata', {}).get('xesam:url')
```
thp: Extract common subexpression:

    metadata = props.get('Metadata', {})
    url = metadata.get('xesam:url')
    track_id = metadata.get('mpris:trackid')

elelay:
Changed according to review.
Thanks,

share/gpodder/extensions/enqueue_in_mediaplayer.py
```diff
@@ -158,37 +152,36 @@ def on_error(exception):
...
+        self.position_us = pos * 1000 * 1000  # pos in microseconds
         if self.init_dbus():
             # async to not freeze the ui waiting for the application to answer
             self.player.OpenUri(self.url,
                                 dbus_interface=self.INTERFACE_PLAYER,
                                 reply_handler=on_reply,
                                 error_handler=on_error)
         else:
-            self.wait_for_player(filename, pos)
+            self.prepare_wait_for_player(filename, pos)
```
vossad01: I only looked at this long enough to confirm you made the changes thp requested, and this is only an observation and not anything I am saying needs to be changed.  The name `prepare_wait_for_player` makes me expect somewhere there will be a "wait for player" and it is not immediately obvious where this happens.  I'm guessing `open_files` has that side-effect.

elelay: Yes: it is `open_files`. `wait_for_player` is also a bad name because it's not blocking. I'll change it to `observe_player_launch`.

elelay: Changed to `enqueue_when_ready` in the end. Thanks!

PR #221, opened 2017-01-29
gtk3: Html shownotes

elelay:
Working implementation of html shownotes rendered using webkit2.
 - headline and subheadline distinct component
 - no navigation  (maybe have to refine this for anchors in same page)
 - context-menu  item 'Open Link in Web Browser' on links
 - context-menu item 'Open Shownotes in Web Browser' on page background.

PR #263, opened 2017-04-02
store episode's HTML description in descr_html (schema change)

elelay:
Keep the description_html property to fallback to description if descr_html is not set.

Contrary to previous PR, a copy of description is not stored in descr_html if episode has no HTML description.

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -145,6 +145,7 @@ class PodcastEpisode(PodcastModelObject):
         episode.title = entry['title']
         episode.link = entry['link']
         episode.description = entry['description']
+        episode.descr_html = entry.get('description_html', '')
         episode.total_time = entry['total_time']
         episode.published = entry['published']
         episode.payment_url = entry['payment_url']
@@ -202,6 +203,7 @@ class PodcastEpisode(PodcastModelObject):
         self.mime_type = 'application/octet-stream'
         self.guid = ''
         self.description = ''
+        self.descr_html = ''
         self.link = ''
         self.published = 0
         self.download_filename = None
@@ -335,6 +337,8 @@ class PodcastEpisode(PodcastModelObject):
 
     @property
     def description_html(self):
+        if self.descr_html:
+            return self.descr_html
         # XXX: That's not a very well-informed heuristic to check
         # if the description already contains HTML. Better ideas?
         if '<' in self.description:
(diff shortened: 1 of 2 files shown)


vossad01:
Do you think Thomas will mind that this is the first instance of a name shortening in the database?  It looks like sqlite does not have a rename command so changing it later would be a bit more work.

I looked into mapping a [class member to a differently named db column](https://github.com/gpodder/gpodder/blob/master/src/gpodder/dbsqlite.py#L203).  It wouldn't be a large refactor, but it would introduce complexity that is not already there, for relatively small returns.

I then looked at the other side of things to gauge the impact of removing the property.  Textual search makes it appear that property is not currently in use within gPodder itself.  More interestingly though, is that [the textual description may derive from the HTML description](https://github.com/gpodder/podcastparser/blob/master/podcastparser.py#L678).  I wonder if that will change as part of your work and then if the `description_html` should even be around.

Sorry for being wishy washy, I'm still feeling out what my role here is supposed to be.

elelay:
Thanks for the feedback!

Naming was hard:
The first PR was indeed directly mapping the `description_html` column to a `description_html` member. But it had the issue of storing the same content twice. So I went back to a computed property, which can't be named the same as the member field. I was torn between keeping `description_html` as property vs column and chose to keep the proper name in code (better compatibility + it's the one we see more often). I discarded using description_html/html_description because I thought it would be confusing.

Yes, podcastparser ensures `description` will always be there in an entry, but `description_html` might be missing (e.g.  if there is no `<content:encoded />` element in an RSS feed. In turn, gPodder assumes there will always be a `description` and will derive `description_html` from it if necessary. Does it make sense to you?

`description_html` is used in the gtk3 branch, for html shownotes display with webkit. And it's the reason why I would like to store both description (plain text short description) and description_html (full shownotes with links) when they differ. So description_html is there to stay :-).

vossad01:
Thanks for getting back to me.  That makes sense.  I guess I had not really realized #221 had been merged so I could look at the implementation and how this is being used :S (and could test it if I installed the right library).

I had been meaning to say the existing `description_html` _property_ may not need to be around, though I recognize the description as html needs exposed in some form.  I see now my wonderings about the podcastparser behavior is well out of scope.

As I see it, an episode may have a description, a html_description, both[, or neither].  If I put aside all legacy considerations, I think the least surprising API would be using that that representation.  The caller could then chose what was appropriate for its use, coercing if necessary.

What do you think of elelay/gpodder#2?

elelay:
elelay/gpodder#2 seems better, indeed. To sum it up:
 - only make available the native `description_html` in `model.Episode` (keep the '<' heuristic to consided description html for now), via a field, not a property.
 - in `gPodderShownotesHTML`, if no `description_html` is available, load description as plain text.

Great!

vossad01:
Replaced by #246 and #247.  Thanks for sticking with me through this ☺️.

PR #264, opened 2017-04-08
Easier self-hosting support

elelay:
- allow restoring subscriptions from another server than gpodder.net (depends on  [gpodder/mygpoclient/pulls/11](https://github.com/gpodder/mygpoclient/pull/11))
 - configurable server in gpodder.net preferences UI (was already modifiable via edit-config).

### DIFF
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -155,7 +155,7 @@ class BuilderWidget(GtkBuilderWidget):
             return None
 
     def show_login_dialog(self, title, message, root_url=None, username=None, password=None,
-            username_prompt=None, register_callback=None, register_text=None):
+            username_prompt=None, register_callback=None, register_text=None, ask_server=False):
         if username_prompt is None:
             username_prompt = _('Username')
 
@@ -177,18 +177,18 @@ class BuilderWidget(GtkBuilderWidget):
         if register_callback is not None:
             dialog.add_button(register_text, gtk.RESPONSE_HELP)
 
-        url_entry = gtk.Entry()
-        url_entry.set_tooltip_text(_('hostname or root URL (e.g. https://my.gpo.me)'))
+        server_entry = gtk.Entry()
+        server_entry.set_tooltip_text(_('hostname or root URL (e.g. https://gpodder.net)'))
         username_entry = gtk.Entry()
         password_entry = gtk.Entry()
 
-        url_entry.connect('activate', lambda w: username_entry.grab_focus())
+        server_entry.connect('activate', lambda w: username_entry.grab_focus())
         username_entry.connect('activate', lambda w: password_entry.grab_focus())
         password_entry.set_visibility(False)
         password_entry.set_activates_default(True)
 
         if root_url is not None:
-            url_entry.set_text(root_url)
+            server_entry.set_text(root_url)
         if username is not None:
             username_entry.set_text(username)
         if password is not None:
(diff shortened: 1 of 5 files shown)


src/gpodder/gtkui/interface/common.py
```diff
@@ -154,7 +154,7 @@ def on_text_changed(editable):
         else:
             return None
 
-    def show_login_dialog(self, title, message, username=None, password=None,
+    def show_login_dialog(self, title, message, root_url=None, username=None, password=None,
```
vossad01: It looks like this can be hit when trying to restore from gPodder.net (can this only be done when there are no local podcasts?) and when encountering a podcast requiring authentication.

Do you know a simple/easy way to test each of these paths?

src/gpodder/gtkui/interface/common.py
```diff
@@ -177,50 +177,63 @@ def show_login_dialog(self, title, message, username=None, password=None,
         if register_callback is not None:
             dialog.add_button(register_text, gtk.RESPONSE_HELP)
 
+        url_entry = gtk.Entry()
+        url_entry.set_tooltip_text(_('hostname or root URL (e.g. https://my.gpo.me)'))
```
vossad01: Would it be better to use https://gpodder.net/ as the example? https://my.gpo.me/ does not load for me.

src/gpodder/gtkui/main.py
```diff
@@ -2921,33 +2921,34 @@ def on_goto_mygpo(self, widget):
         self.mygpo_client.open_website()
 
     def on_download_subscriptions_from_mygpo(self, action=None):
+        def after_login():
+            dir = gPodderPodcastDirectory(self.gPodder,
+                                           _config=self.config,
+                                           custom_title=_('Subscriptions on gpodder.net'),
```
vossad01: Could we here substitute the actual url used instead of always showing "gpodder.net"?

src/gpodder/gtkui/interface/common.py
elelay: Woops: good catch. This would have been bad.

To test:
 - restore from gPodder, I start with an empty GPODDER_HOME:
   `GPODDER_HOME=/tmp/GPO_HOME bin/gpodder -v`
 - authenticate podcast, I start the test-auth-server and subscribe to it:
   `python2 tools/test-auth-server.py&`

src/gpodder/gtkui/main.py
elelay: definitely!

elelay:
Do you think a checkbox _Self-hosting_ revealing the server line would be better? In the end I chose not to implement it to not add too much code, but it's feasible I guess.

vossad01:
If you went that route, I think a radio toggle between "gpodder.net' and "Custom host" (or similar language) would be a better pattern.

I agree that I don't think that complexity is needed at the moment.  I mentioned it because it was a consideration and worth being aware of, but when I saw it I felt like I had seen enough other programs with similar configuration options that it probably was not going to be a problem.

This is all opinion and conjecture, so I may be wrong.  However, if that is the case or if someone later disagrees this can always be changed easily enough. The rest of the PR functionality seems like it is still useful even if this one aspect ends up changed in the future.

elelay:
Let's roll it out, then ;-)

vossad01:
@vossad01 merged commit 709d3377e190151853938903a8e771eb919a8810 into `gpodder:gtk3`

vossad01:
Small cosmetic oddity on Gtk3:

When you access restore on a new environment, the server name __gpoddder.net__ is highlighted even though focus is correctly in the user name field.
If you previously assessed restore and set the server, accessing the restore again does not have the entry highlighted.

vossad01:
@elelay Thanks for this!

PR #266, opened 2017-04-09
if description is not html, show it as plain-text in webkit

elelay:
Also the html shownotes header is now selectable.

This works fine for me. Before adding about:blank to on_decide_policy, the plain text shownotes where not displaying. Alternating between html  and plain text shownotes displays fine.
I'm using http://radiofrance-podcast.net/podcast09/rss_11901.xml to test plain text episodes.

### DIFF
--- a/src/gpodder/schema.py
+++ b/src/gpodder/schema.py
@@ -28,6 +28,8 @@ import shutil
 import logging
 logger = logging.getLogger(__name__)
 
+from gpodder import util
+
 EpisodeColumns = (
     'podcast_id',
     'title',
@@ -108,6 +110,9 @@ UPGRADE_SQL = [
         # Version 7: Add HTML description
         (6, 7, """
         ALTER TABLE episode ADD COLUMN description_html TEXT NOT NULL DEFAULT ''
+        UPDATE episode SET description_html=description WHERE is_html(description)
+        UPDATE episode SET description=remove_html_tags(description_html) WHERE is_html(description)
+        UPDATE podcast SET http_last_modified=NULL, http_etag=NULL
         """),
 ]
 
@@ -193,6 +198,9 @@ def upgrade(db, filename):
     if not list(db.execute('PRAGMA table_info(version)')):
         initialize_database(db)
         return
+
+    db.create_function('is_html', 1, util.is_html)
+    db.create_function('remove_html_tags', 1, util.remove_html_tags)
 
     version = db.execute('SELECT version FROM version').fetchone()[0]
     if version == CURRENT_VERSION:
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -604,8 +604,8 @@ def is_html(text):
     >>> is_html('a < b < c')
     False
     """
-    e = re.compile('<[a-zA-Z][a-zA-Z0-9]*(\\s.*)?>')
-    return e.search(text) is not None
+    html_test = re.compile('<[a-z][a-z0-9]*(?:\s.*?>|\/?>)', re.IGNORECASE | re.DOTALL)
+    return bool(html_test.search(text))
 
 
 def remove_html_tags(html):
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -595,6 +595,19 @@ def delete_file(filename):
         pass
 
 
+def is_html(text):
+    """Heuristically tell if text is HTML
+
+    By looking for an open tag (more or less:)
+    >>> is_html('<h1>HELLO</h1>')
+    True
+    >>> is_html('a < b < c')
+    False
+    """
+    e = re.compile('<[a-zA-Z][a-zA-Z0-9]*(\\s.*)?>')
+    return e.search(text) is not None
+
+
 def remove_html_tags(html):
     """
     Remove HTML tags from a string and replace numeric and
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -182,6 +182,7 @@ class gPodderShownotesHTML(gPodderShownotes):
         self.header.set_halign(Gtk.Align.START)
         self.header.set_valign(Gtk.Align.START)
         self.header.set_property('margin', 10)
+        self.header.set_selectable(True)
         self.status = Gtk.Label.new()
         self.status.set_halign(Gtk.Align.START)
         self.status.set_valign(Gtk.Align.END)
@@ -202,13 +203,11 @@ class gPodderShownotesHTML(gPodderShownotes):
             self._base_uri = episode.channel.url
         self._loaded = False
 
-        description = episode.description_html
-        if not description:
-            description = self.coerce_text_to_html(episode.description)
-        self.html_view.load_html(description, self._base_uri)
-
-    def coerce_text_to_html(self, text):
-        return text.replace('\n', '<br>')
+        description_html = episode.description_html
+        if description_html:
+            self.html_view.load_html(description_html, self._base_uri)
+        else:
+            self.html_view.load_plain_text(episode.description)
 
     def on_mouse_over(self, webview, hit_test_result, modifiers):
         if hit_test_result.context_is_link():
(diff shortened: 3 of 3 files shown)


vossad01:
The good:  This works now (when my attempt didn't), good job!

The bad: We has an issue here with existing data.  Prior to #246 `html_description` did not exist so was not being populated.  However, we still may have HTML in `description.  Since this discriminates on `html_description` it results in pre-v7 schema entries displaying raw HTML to the user if the episode provided if the episode had an HTML description.

There are largely two options:
 1. Detect these cases in code
 2. Migrate the data

(1) is simpler, easier , and safer
(2) is cleaner and better in the long term, but more risky and permanent.  If this approach would be taken, I would recommend it be done after #267 (if it is accepted).  That way instead of copying and duplicating the data, we could move the data leaving `description` empty when HTML is available.

elelay:
I've seen it to, but considered it would be solved with time (new episodes coming and older going). This was oversimplifying. I'd go for (1).

vossad01:
@elelay I like your thinking.  If resolving the issue were sufficiently complex/costly, I might even agree with you.  However, I think a non-negligible number of users treat their gPodder downloads as a podcast archive.  I have seen reports about hundreds of podcasts or many, many GBs of data.  So we would be likely to get bug reports if we break old episodes on upgrade (unless they are like me and never look at the description).  Since it is simple enough to avoid the problem, I would prefer to keep users happy.

thp:
This case should only be hit for situations where we migrate from database schema versions before 7 to version 7 (when `description_html` was introduced), right?

In that case, would it be enough for the upgrade from version 6 to version 7 to also initialize the `description_html` field with the value from `description` and either initialize the description column with a stripped version or clear it (and make sure that code can detect this and use the HTML description, possibly stripped)?

Something like this (note that this can be done in one `UPDATE` statement, but [let's play it safe](http://stackoverflow.com/questions/2203202/sql-update-order-of-evaluation)):

    UPDATE episode SET description_html=description
    UPDATE episode SET description=NULL

Or alternatively, just the first `UPDATE` (to initialize the HTML description), and then for the second one, instead of setting it to `NULL`, expose a Python function that does the HTML stripping to SQLite (see https://docs.python.org/2/library/sqlite3.html#sqlite3.Connection.create_function for an example) and then do:

    UPDATE episode SET description_html=description
    UPDATE episode SET description=strip_html_tags(description_html)

In the exposed `strip_html_tags` we can then detect if it's HTML and create a stripped version. The advantage of that would be that it's only done once (at DB migration time), only done for old installations (so new DBs are not affected) and wouldn't cause issues when the feed is updated, since if the feed provides both plaintext and HTML, the next update will overwrite the values.

We could also add this into the 6-to-7 migration to force feed updates (so that new descriptions will be loaded):

    UPDATE podcast SET http_last_modified=NULL, http_etag=NULL

vossad01:
@thp Your understanding of when this occurs is correct and migrating the data is indeed an option.

With this PR as-is, it needs `description` to only have plain text and `description_html` to only have HTML.  On schema version 6, `description` could be either text or HTML, so initializing `description_html` with the value from `description` won't cover all cases correctly.

A python function to check if it contains HTML could be used, yielding something like this:
```
UPDATE episode SET description_html=description WHERE is_html(description)
UPDATE episode SET description=NULL where description_html IS NOT NULL
```

Whether that second line sets it to null or sets it to `strip_html_tags(description_html)` would depend on the outcome of gpodder/podcastparser#12

elelay:
@vossad01 OK, I'll commit a version setting description to `UPDATE episode SET description=strip_html_tags(description) where description_html IS NOT NULL` and we can later make another commit to set it to NULL.

elelay:
Committed the schema update.
I used: 
 - http://feeds.feedburner.com/linuxoutlaws to test description vs description_html (activate Menu / episode description) and HTML shownotes
 - http://radiofrance-podcast.net/podcast09/rss_11901.xml to test text shownotes

src/gpodder/schema.py
```diff
@@ -194,6 +199,9 @@ def upgrade(db, filename):
         initialize_database(db)
         return
 
+    db.create_function('is_html', 1, util.is_html)
+    db.create_function('remove_html_tags', 1, util.remove_html_tags)
```
vossad01: I have confirmed `util.remove_html_tags` has the same logic as is used by **podcastparser** so no risk of inconsistency there.

src/gpodder/util.py
```diff
@@ -595,6 +595,19 @@ def delete_file(filename):
         pass
 
 
+def is_html(text):
```
vossad01: Unless you see something otherwise, I will prefer the [implementation submitted to **podcastparser**](https://github.com/gpodder/podcastparser/pull/12/files#diff-e8db52e77ac4eba974cc9019d91ca9f0R847).  I have not done comparative testing, but just in reading the RegExes that one seems more comprehensive.

elelay: Sorry: I missed your PR.
I started from the regexp you pointed to in a previous comment (`r'<[a-z][\s\S]*>', re.IGNORECASE`), but found `[\s\S]*` equivalent to `.*` (isn't it?) and found no reason to not restrict the element name, only accepting unspecified characters after a space (`(\\s.*)?`). This way, `a<b<c` is rejected, for instance.
It can use the ignore case flag: `'<[a-z][a-z0-9]*(\\s.*)?>', re.IGNORECASE`. It's a gross over approximation anyway, so I'm OK with using `r'<[a-z].*>', re.IGNORECASE`.

vossad01: > but found [\s\S]* equivalent to .* (isn't it?)

There is a difference when the element spans multiple lines: http://stackoverflow.com/a/35248940/1072626.  This may be important if the HTML was pretty printed.

An alternative option would be the dot-matches-all flag.

> r'<[a-z][a-z0-9]\*(\s.\*)?>'

Matches **<** followed by an alpha letter, followed by any number of alpha letter and numbers, optionally followed by a space then any number of characters on that line, followed by **>**

> r'<[a-z][\s\S]*>'

Matches **<** followed by an alpha letter, followed by any number of whitespace and non-whitespace characters (spanning lines), followed by **>**

> found no reason to not restrict the element name

I admit to being a little confused by the double negative, you "found reason to restrict the element name"?  Reading the regex, the difference seems to be in restricting to element name to (roughly) the ASCII alpha-numeric range.  I have been able to confirm that that is a restriction in Chrome on custom elements, but I have not easily verified whether HTML allows for Unicode in element names.  Especially if this can be confirmed, you may be right in that being a good restriction to add.


**tl;dr:**  Need to decide if elements should be allowed unicode characters, and decide whether to use [\s\S] or pass a flag to change dot behavior.

src/gpodder/schema.py
vossad01: I looked for a way to delete these functions when we are done and could not find anything.  I am assuming they are transient.

vossad01:
I pushed this trough so I could direct dannn-o to test the fix to #274 without him missing these migrations.  Also, I should not have merged this, because I realized the migration commits were added directly to the commits incoming to the gtk3, I should have had them merged from master.  I fixed with #277, but there are now duplicate commits in history.

In hindsight, I should have just made the schema migration v8, instead of letting it be added to the already released v7 and avoided the rush entirely.

vossad01:
As far as I am concerned, the RegEx is still open for discussion and I would be interested in your thoughts @elelay.  For the sake of not delaying an implementation I wrote a new one that I think captures all the cases mine supported plus imposes your ASCII element name limitation, but it can still be changed.

vossad01:
I don't know that it is something that gPodder will encounter soon, but Chrome at least does allow unicode characters in custom elements:
```js
window.customElements.define('custom-中文', class extends HTMLElement {});
```

and [the ANTLR HTML grammar also shows some unicode characters as vaild](https://github.com/antlr/grammars-v4/blob/master/html/HTMLLexer.g4#L119-L128).

However, neither of those are authoritative sources.

elelay:
@vossad01 your RE is fine, thanks! I indeed forgot the multiline aspect when looking into it.
While I don't expect shownotes consisting only of custom elements, there might be the case of xhtml with non-default prefix (<html:p>) in atom feeds.
So maybe we go to your RE in the end?

vossad01:
I pursued this largely because I didn't know, and was thus curious, what is actually allowed for html.  In reality, none of the RegExes are perfect and I think the probability that the differences between them will ever manifest in an actual podcast is near zero.

Since the one in gpodder/podcastparser#12 and gpodder/podcastparser#13 are largely "just as good" as the one here, but also simpler, I figure I will leave them as-is for simplicity sake.

My main motivation in changing the gPodder one again would be if it were possible to have it delegate to the _podcastparser_ implementation, but I am not sure if that would break the dependency model.  Something to worry about at a later date.

PR #271, opened 2017-04-14
rename_download: Handle new basename as unicode string

thp:
As alternative fix for #217. Note that I wasn't able to reproduce the original issue on macOS, so can't verify that this actually fixes the original issue.

### DIFF
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -48,12 +48,12 @@ class gPodderExtension:
         basename, ext = os.path.splitext(filename)
 
         new_basename = []
-        new_basename.append(util.sanitize_encoding(title) + ext)
-        if self.config.add_podcast_title:
-            new_basename.insert(0, podcast_title)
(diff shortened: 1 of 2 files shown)


vossad01:
For documentation completeness, here is the failure reported by @matoro after applying this fix:

```
1492578451.939707 [gpodder.download] ERROR: Download failed: 'ascii' codec can't decode byte 0xe3 in position 42: ordinal not in range(128)
  File "/usr/lib/python2.7/site-packages/gpodder/download.py", line 820, in run
    self.filename = self.__episode.local_filename(create=True, force_update=True)
  File "/usr/lib/python2.7/site-packages/gpodder/model.py", line 487, in local_filename
    old_file_name = os.path.join(self.channel.save_dir, self.download_filename)
  File "/usr/lib/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 42: ordinal not in range(128)
```

Issue #272, opened 2017-04-14
Cannot Turn MPRIS Listener On


southvinement:
In preferences, extensions, desktop integration, when I check mpris listner, I get error dialog box:

extension cannot be activated

'module' object  has no attribute 'signal'

First Time user

thp:
Which version of gPodder are you using?

southvinement:
gPodder 3.9.3
Python 2.7.12
Win 7 SP1

vossad01:
Have you installed a D-Bus implementation?  [MPRIS is a D-Bus interface](https://specifications.freedesktop.org/mpris-spec/latest/) and Windows does not come with D-Bus.  Due to this, I would not expect the extension to work out of the box.

Even if it is not going to work for you, I recognize the error message is not helpful.

southvinement:
What is a D-bus implementation?

The wiki, https://github.com/gpodder/gpodder/wiki/_Extensions-MprisListener, doesn't mention this.

Is there an explanation anywhere of how to get this to work?

vossad01:
You are right, the wiki does assume the reader know what [D-Bus](https://www.freedesktop.org/wiki/Software/dbus/) ([wikipedia](https://en.wikipedia.org/wiki/D-Bus)) is or at least looks it up.  Please feel free to edit it to be how you would expect.

[Freedesktop.org indicates there is a Windows port](https://www.freedesktop.org/wiki/Software/dbus/#windowsport), but I have no experience nor advice regarding it.

Just to fully set expectations, even if you do get D-Bus it working with gPodder on Windows, I don't know how many media players would use it.  No Windows-specific players would and I don't know if cross-platform players would look for it when compiled for Windows.

southvinement:
The link says there is a port but doesn't say where to find it.  Maybe the reference to compilers means you are supposed to compile it for Windows?  I found a file at https://sourceforge.net/projects/windbus/.  I downloaded it.  7Zip said unexpected end of data.  In the extraction I did not find anything that looked like an install file.  I haven't found anything more about it.

Having a restart time in the duration column would be very valuable for me.

Does anybody know any more about getting this to work?

thp:
On Windows, the MPRIS plugin doesn't really make any sense, since it wouldn't work with any media player there. It's something that only makes sense on freedesktop.org-based machines (Linux and the *BSDs) with D-Bus. While one can probably install enough of D-Bus and maybe even get a player to work with D-Bus, this is not something that I'd recommend.

[...]

PR #273, opened 2017-04-16
Disable MPRIS listener extension in Windows and macOS

thp:
Based on the discussion in #272.


vossad01:
I agree that it makes sense to do this.

I have not looked into extensions before, is the following correct?
* `__only_for__`
    Extension only appears if executing ui is in given list
* `__mandatory_in__`
    Extension is enabled and hidden if executing ui is in given list
* `__disable_in__`
    Extension is disabled and hidden if executing ui is in given list

If it is it just struck me that `__only_for__`  and `__disable_in__` provide different ways to do the same thing, such that this could alternatively be `__disable_in__ = 'win32,osx'`.

elelay:
@vossad01 your reasoning is correct. But besides boolean logic, having 2 keywords adds flexibility to express one's intent I guess.

vossad01:
@elelay Does this need to land in `gtk3` too?

elelay:
> Does this need to land in gtk3 too?

Yes, I forgot. Done.

PR #283, opened 2017-04-21
Remove some characters from filename when copy fails

### LINKED ISSUE #282: Send to folder does not use VFAT safe names
When a filename contains characters that are not valid on a destination VFAT target, send to fails with the following exception:

```
1492734407.836906 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/home/user/dev/gpodder/src/gpodder/gtkui/main.py", line 1750, in save_episodes_as_file
    shutil.copyfile(copy_from, copy_to)
  File "/usr/lib/python3.5/shutil.py", line 115, in copyfile
    with open(dst, 'wb') as fdst:
OSError: [Errno 22] Invalid argument: '/media/user/0123-4567/Podcasts/Episode 84: Ur Fave is Problematic.mp3'

  File "/home/user/dev/gpodder/src/gpodder/gtkui/main.py", line 1750, in save_episodes_as_file
    shutil.copyfile(copy_from, copy_to)
  File "/usr/lib/python3.5/shutil.py", line 115, in copyfile
    with open(dst, 'wb') as fdst:
OSError: [Errno 22] Invalid argument: '/media/user/0123-4567/Podcasts/Episode 84: Ur Fave is Problematic.mp3'
```

This was also reported by @dannn-o in unrelated issue #279 .

This is important because a lot of MP3 players will expose themselves as USB Mass Storage with msdos formatting.

**Podcast for reproduction:** http://rss.acast.com/callyourgirlfriend (most every recent episodes includes a colon (**:**) which will trigger the failure.

## Workaround
Send to a supported target volume first then copy to the ultimate destination counting on the OS to handle sanitizing the file name.  (This works on Linux Mint Cinnamon anyway).

vossad01:
This aims to provide better VFAT/Windows support by replacing troublesome characters if there is originally a failure. This would fix #282.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -29,6 +29,7 @@ from gi.repository import GdkPixbuf
 from gi.repository import GObject
 from gi.repository import Pango
 import random
+import re
 import sys
 import shutil
 import subprocess
@@ -1747,7 +1748,11 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     base, extension = os.path.splitext(copy_from)
                     filename = self.build_filename(episode.sync_filename(), extension)
                     copy_to = os.path.join(folder, filename)
-                    shutil.copyfile(copy_from, copy_to)
+                    try:
+                        shutil.copyfile(copy_from, copy_to)
+                    except (OSError, IOError):
+                        copy_to = os.path.join(folder, re.sub(r"[\"*/:<>?\\|]", "_", filename))
+                        shutil.copyfile(copy_from, copy_to)
 
     def copy_episodes_bluetooth(self, episodes):
         episodes_to_copy = [e for e in episodes if e.was_downloaded(and_exists=True)]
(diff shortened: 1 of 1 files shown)


vossad01:
Unrelated observation: when testing this, I noticed that if there is not room on the destination drive, the copy will attempt anyway, resulting in a partial file at the destination.

src/gpodder/gtkui/main.py
```diff
@@ -1747,7 +1748,11 @@ def save_episodes_as_file(self, episodes):
                     base, extension = os.path.splitext(copy_from)
                     filename = self.build_filename(episode.sync_filename(), extension)
                     copy_to = os.path.join(folder, filename)
-                    shutil.copyfile(copy_from, copy_to)
+                    try:
+                        shutil.copyfile(copy_from, copy_to)
+                    except (OSError, IOError):
+                        copy_to = os.path.join(folder, re.sub(r"[\"*/:<>?\\|]", "_", filename))
```
elelay: Check that the new copy_to is different than the original one. Otherwise there is no point in trying again.

vossad01:
I can add that if you have an idea when it would be useful.  If the destination does not support these characters the user could easily have 100s of such messages (which they likely can't do anything about), if it does support it then they will have no messages.

If the reason for the exception is something other than the filename, then it is highly likely an exception will be thrown on the second attempt and propagate from there.

elelay:
And the new exception will include the retry line so we'll know we have changed the name.
OK.

Issue #275, opened 2017-04-18
When refreshing, show "updating <current>" instead of "updated <last>" in status?


mckea:
A couple of my feeds take a long time to update - but I'm not sure which ones since the update status message shows the last successfully updated feed instead of the one currently in progress.

Might this be changed?

elelay:
#284 has landed in gtk3

PR #284, opened 2017-04-21
Display currently updating podcast when updating subscriptions

vossad01:
Display currently updating podcast when updating subscriptions. Prior to this, it would display the last podcast that was updated rather than the currently updating podcast.  This implements #275.


elelay:
Great! I have a podcast often timeouting in my library and now it's shown, not the podcast before.

PR #288, opened 2017-04-23
disable java, javascript, plugins in HTML shownotes

elelay:
Better play safe.
I found only 1 minor issue with this : the audio player embedded in No Agenda shownotes doesn't work.
But what's the point in using it?


vossad01:
Thanks!  I feel we are a safer having this.

Issue #248, opened 2017-03-26
gtk3: localdepends.py downloads an incompatible mygpoclient


adamvoss:
When running from git on the `gtk3` branch, and you run the following do download dependencies:

    python3 tools/localdepends.py

The result when your try to run gPodder is:

> ERROR: Uncaught exception: Traceback (most recent call last):
  File "./bin/gpodder", line 159, in <module>
    main()
  File "./bin/gpodder", line 152, in main
    from gpodder.gtkui import main
  File "/home/vossad01/dev/gpodder/src/gpodder/gtkui/main.py", line 56, in <module>
    from gpodder import my
  File "/home/vossad01/dev/gpodder/src/gpodder/my.py", line 62, in <module>
    from mygpoclient.simple import MissingCredentials
  File "/home/vossad01/dev/gpodder/src/mygpoclient/simple.py", line 23, in <module>
    from mygpoclient import json
  File "/home/vossad01/dev/gpodder/src/mygpoclient/json.py", line 79
    except ValueError, ve:
                     ^
> SyntaxError: invalid syntax

clearly this version is not compatible with Python 3.

### Workaround
Clone [gpodder/mygpoclient](https://github.com/gpodder/mygpoclient) and manually copy the source files.

Issue first mentioned in #187

elelay:
I'm not inclined to modify the script to fetch the archive from github.
A new version of mygpoclient will have to be released to solve the issue.

vossad01:
@thp It looks like [you are the owner on pypi](https://pypi.python.org/pypi/mygpoclient), is there anything more that would need to be done before a new version could be published?

adamvoss:
Reopening because a new release has not been published yet.

adamvoss:
Verified using a new clone, checking out the gtk3 branch, then using the steps from [running from Git](https://gpodder.github.io/docs/run-from-git.html).  gPodder started and successfully checked for new episodes.

PR #290, opened 2017-04-23
fix #159 - gtk3: Empty tree views do not indicate why

### LINKED ISSUE #159: gtk3: Empty tree views do not indicate why
**On gtk2,** 
On the Podcasts tab,
the main GtkTreeView will say (centered text) "Loading..." when the application has launched but before the list of episodes has been populated.
I believe there may be different text when it has completed loading but no episodes were found.

On the Progress tab,
The GtkTreeView will say "No active tasks" when it has an empty queue.

**On gtk3,**
The GtkTreeView on both tabs does not have any text, thus users do not know whether they are still waiting on loading or if it is correctly blank.

elelay:


vossad01:
Thanks for working on this!  This seems to fix the "Progress" tab, but the Podcast tab still showed blank in my testing.

GTK2:
screenshot from 2017-04-23 11-00-54

GTK3:
screenshot from 2017-04-23 11-03-32

elelay:
Strange: it works for me.
From an empty GPODDER_HOME, ArchLinux, gtk3 3.22.10-1, default theme.

show_empty

elelay:
Same with `en_US.utf8` locale.
Any stack trace?

vossad01:
Sorry my bad.  Midway through I switched to a different (slower) box to better confirm this and after working around #248 I accidentally switched back to `gtk3` rather than this PR.

elelay:
No worries :-)

PR #292, opened 2017-04-25
Convert README to Markdown

adamvoss:
I kind of hate to do this because the README is a very well done for a plain text readme, it is practically a work of art.  However, it does not work as well for the web: fixed width fonts are not as attractive to read, does not play well on mobile devices (narrow view ports), no navigation anchors.

There is more that could probably be done with respect to utilizing markdown features to add more formatting if someone wanted, but I think this is a fine initial conversion.


thp:
:+1:

Issue #268, opened 2017-04-12
Removing feed deletes entire folder


Thlayli:
I'm starting to come out of shock and get angry. I just changed the URL of a feed, and since I couldn't change the feed URL in Podcast settings, I decided to delete and re-add the feed. It warned me about deleting files, but I assumed the worst it would do is delete the 15 or so items I'd downloaded from Gpodder. Apparently that warning means "we're about to remove 76GB of data that was already in that folder without even using the recycle bin."

WHAT THE HECK POLICY IS THAT?

Thlayli:
Perhaps implement the following recycle bin code so it doesn't permanently delete anyone else's entire collection of TV shows. :tongue: 

http://www.hardcoded.net/articles/send-files-to-trash-on-all-platforms.htm

elelay:
I don't use a trashcan on my system so I wouldn't want this (no way to empty the trash). But I understand we should leave files unknown to gPodder untouched when deleting a podcast.

vossad01:
I think the language contradicts the behavior, but I agree with the behavior.

Unless you really go digging in the wiki to find the right environment variable to set up, the gPodder download directory exists within **GPODDER_HOME**, together with other gPodder files, and consistent with the fact the files are managed and governed by gPodder.  This is similar to how how _Calibre_ manages **Calibre Library** and _iTunes_ manages **iTunes Music** (or whatever the name is now that iTunes does so much more than music).  It is reasonable to expect gPodder controls these files.  On startup, it emits warnings (to log) when unknown external files are found and will also automatically associate external files it can match to podcast episodes even if they were not downloaded via gPodder.  I have been known to access my podcast episodes from external HDD without gPodder installed and appreciate that it keeps the podcast folder [that it manages] clean.

However, the term "Remove" **is** open to misinterpretation.  A user may think that means "Remove from gPodder" not "Remove from file system."  Just like you "Delete" an episode, it should say "Delete podcast."  This could be done in addition to the text being changed to be more explicit about deletion behavior (ex. "This podcast and all its episodes will be **permanently deleted**.")

Regarding the Trash/Recycling Bin behavior, I agree with @elelay, that I would want direct deletion.  I also recognize that many users do appreciate use of a Recycle Bin, so I think it would be acceptable to either:
- make the deletion dialog have a check-box to indicate permanent delete is desired,
- or have separate menu entries "Move podcast to trash" and "Delete podcast" from which the user can choose (a la Amarok).

[...]

PR #294, opened 2017-04-26
Rename "Remove podcast" to "Delete podcast"

### LINKED ISSUE #270: Rename "Remove podcast" to "Delete podcast"
The term "Remove" is open to misinterpretation. A user may think that means "Remove from gPodder" not "Remove from file system."  This can lead to issues such as  #268.  Just like you "Delete" an episode, it should say "Delete podcast."

While doing this it would helpful to clarify the behavior of removal in the confirmation dialog.  Example: "This podcast and all its episodes will be **permanently deleted**."  (Cancel/Okay)

adamvoss:
Simple text changes to rename "Remove podcast" to "Delete podcast".  Fixes #270.

### DIFF
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -70,7 +70,7 @@
         </item>
         <item>
           <attribute name="action">win.massUnsubscribe</attribute>
-          <attribute name="label" translatable="yes">Remove podcasts</attribute>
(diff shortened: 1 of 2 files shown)


thp:
I'm a bit unsure about the all-caps "PERMANENTLY DELETED", but "remove" -> "delete" is a good change.

adamvoss:
I would have preferred bold, but that did not seem simple to do.  #268 was somewhat critical in tone so I figured it was easiest to take the CYA approach.

Issue #205, opened 2017-01-12
Syncing multiple files to device takes too long


rogg:
Does gPodder copy podcasts to the device one at a time or all at once? I ask because syncing multiple files to my Sansa Clip Zip takes much longer than simply copying them with the file manager, which does them one at a time.

#### Example (two podcasts, ~50 MB each):
* Nautilus: 25 seconds (including safely ejecting)
* gPodder: 110 seconds

thp:
Can you check if setting `sync_disks` to `False` in the advanced configuration editor helps?

https://github.com/gpodder/gpodder/blob/master/src/gpodder/config.py#L170

rogg:
Config editor showed device_sync.after_sync.sync_disks was already False by default. I changed it to True, restarted gPodder and got 140 seconds.

PR #300, opened 2017-05-08
Refactor async tasks: Download/Synchronize/SendTo

elelay:
Download Tasks are rate and thread limited in the UI as before.
Sync, SendTo Tasks are thread limited to 1 (I've yet to see an MP3
player performing well on concurrent write).

task.py is moderately activity-kind agnostic.

SendTo has a similar algorithm as Sync to file-based device:
 - check if FS has enough room;
 - copy file chunk by chunk to track progress/cancel.

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -24,12 +24,9 @@
 #
 #  Based on libwget.py (2005-10-29)
 #
-
-
-
 import logging
-logger = logging.getLogger(__name__)
 
+from gpodder.task import Task, TaskCancelledException, DOWNLOAD_ACTIVITY
 from gpodder import util
 from gpodder import youtube
 from gpodder import vimeo
@@ -37,22 +34,21 @@ from gpodder import escapist_videos
 import gpodder
 
 import socket
-import threading
 import urllib.request, urllib.parse, urllib.error
 import urllib.parse
 import shutil
 import os.path
 import os
 import time
-import collections
 
 import mimetypes
 import email
 
-from email.header import decode_header
 
+logger = logging.getLogger(__name__)
 
 _ = gpodder.gettext
+
 
 def get_header_param(headers, param, header_name):
     """Extract a HTTP header parameter from a dict
@@ -76,6 +72,7 @@ def get_header_param(headers, param, header_name):
         logger.error('Cannot get %s from %s', param, header_name, exc_info=True)
 
     return value
+
 
 class ContentRange(object):
     # Based on:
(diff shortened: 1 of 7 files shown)


elelay:
Sync to device syncs one file at a time (may help #205)

vossad01:
Thanks for this!!!  Definitely an area that can benefit from cleanup!

I read through the diff to the extent of making sure there is nothing obviously insidious or not idiomatic.  Since I haven't done the diligence of testing it, I am not merging it at this time.

elelay:
@vossad01 thanks for the review.

One thing that's not perfect is that it lets one overcommit in SendTo: one can send many files to a disk where there is not enough room. It doesn't warn you on creation. Only when tasks are executed is it checked. Maybe one would prefer to be warned beforehand. But it would require inspecting currently enqueued tasks, finding their type (SendTo), finding the target disk (recursively calling os.stat) and summing up all that. Too much?

vossad01:
Personally, I don't mind that it would fill up the drive then error.  I'd at least end up with as much media as I could on my device that way even if I didn't get everything that tried to transfer.  Presumably I would see failed tasks for what I didn't get.

adamvoss:
I spent a little time testing this.  However, I am not sure if any of what I saw is actually resultant from these changes.

Most notable was the following warring/error which repeated many many times when I sent a bunch of files to a local folder then switched to the progress tab to watch:

```
(gpodder:4851): Gtk-CRITICAL **: /build/gtk+3.0-2Ut_nl/gtk+3.0-3.18.9/./gtk/gtktreeview.c:6623 (validate_visible_area): assertion `has_next' failed.
There is a disparity between the internal view of the GtkTreeView,
and the GtkTreeModel.  This generally means that the model has changed
without letting the view know.  Any display from now on is likely to
be incorrect.


(gpodder:4851): Gtk-CRITICAL **: gtk_tree_model_sort_get_value: assertion 'VALID_ITER (iter, tree_model_sort)' failed

(gpodder:4851): Gtk-CRITICAL **: gtk_tree_model_sort_get_path: assertion 'priv->stamp == iter->stamp' failed

(gpodder:4851): Gtk-CRITICAL **: gtk_tree_path_to_string: assertion 'path != NULL' failed
```

The other things did not reproduce I'm not especially concerned about them:

- At one point it erroneously said the drive was full and died due to a Segmentation fault Fault
- Using a combination of pausing and resuming (my mouse is dieing and accidentally double clicks) I got into a state where only 2 downloads were active with ~4 queued even though my limit was 3.  It corrected once one of the active downloads completed.

elelay:
@adamvoss thanks for the testing!
I've reproduced and fixed the "drive full" crash. It was caused by a combination of typos causing Exceptions, and calling the task creation in background instead of the main thread,  which broke the UI.
I couldn't reproduce the `disparity between the internal view of the GtkTreeView`. If you have time, could you retry with my fix?
Regarding pausing+resuming I'll investigate further...

thp:
Could the tree model change be because we're modifying the model from another thread? If we move the modification to be scheduled (ˋìdle_addˋ) maybe this could fix the issue? Apart from that, I'm okay with this change :)

PR #322, opened 2017-06-25
add generic Extras menu contribution extension point

elelay:
gtk2 implementation.

### DIFF
--- a/share/gpodder/examples/hello_world.py
+++ b/share/gpodder/examples/hello_world.py
@@ -37,3 +37,17 @@ class gPodderExtension:
     def on_unload(self):
         logger.info('Extension is being unloaded.')
 
+
+    def on_ui_object_available(self, name, ui_object):
+        """
+        Called by gPodder when ui is ready
+        """
+        if name == 'gpodder-gtk':
+            self.gpodder = ui_object
+
+    def on_create_menu(self):
+        return [("Say Hello", self.say_hello_cb)]
+
+    def say_hello_cb(self):
+        print("HELLO")
+        self.gpodder.notification("Hello Extension", "Message", widget=self.gpodder.main_window)
--- a/share/gpodder/extensions/minimize_on_start.py
+++ b/share/gpodder/extensions/minimize_on_start.py
@@ -18,5 +18,8 @@ class gPodderExtension:
 
     def on_ui_object_available(self, name, ui_object):
         if name == 'gpodder-gtk':
-            ui_object.main_window.iconify()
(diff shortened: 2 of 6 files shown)


adamvoss:
I noticed that gPodder needed to be restarted after enabling the extension, before the entry would appear.

The enabling the hello_world extension also causes the following additional logging:

adamvoss:
I tested by installing the Hello World extension (`ln -s share/gpodder/examples/hello_world.py share/gpodder/extensions/`) then enabling it in gPodder preferences.

I did notice that I needed to restart gPodder for the menu entry to appear, though the logging from the extension did activate without restart.  Is needing to restart gPodder expected?  Should we add a note in the extension description or something about this limitation?

elelay:
Good point!
This might also be corrected in code, by calling the relevant hooks upon activation.
It's a bit tricky because `extensions.py` is UI agnostic.

elelay:
I'm implemented dynamic menu loading in #323. Extensions menu items are added in a proper section of the *Extras* menu.

elelay:
also on gtk2 now.

adamvoss:
Thanks for implementing this!  It seems to have some unintended consequences though.  Many extensions will throw an exceptions:
```
1501112378.452013 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/home/user/dev/gpodder/src/gpodder/gtkui/desktop/preferences.py", line 413, in on_extensions_cell_toggled
    self.on_extension_enabled(container.module)
  File "/home/user/dev/gpodder/src/gpodder/gtkui/main.py", line 3543, in on_extension_enabled
    extension.on_ui_initialized(self.model,
AttributeError: gPodderExtension instance has no attribute 'on_ui_initialized'
```

because they do not implement `on_ui_initialized`.   The easiest fix would be to check for the existence of the member, but I am definitely open to cleaner solutions.

Additionally, this also causes start-up actions to happen when the extension enables which is perhaps not always desired.  For example enabling "Minimize on start" minimizes the application immediately when enabled.

elelay:
good catch @adamvoss!
Now I check before calling methods.
Regarding minimize on start, I'm tented to add a boolean `isRealStart` to `on_ui_iniatlized` and `on_ui_object_available` but it'll break compatibility. Maybe introduce a new `on_lifecycle` with varying parameters and the `isRealStart` information ? Or a boolean somewhere `realStart` that extensions can check by themselves before doing things meant for startup?

elelay:
Added a new `on_application_started` hook

Issue #351, opened 2017-10-19
gPodder does not start up if the C locale is selected and certain extensions are active


ideasman42:
With this command and Python3, loading gpodder gives an error.

Note that I found this by accident, the following command is just an easy way to redo.

```
export LC_CTYPE=C
gpodder
```

Gives the following error:

```
1508423397.708435 [gpodder.sync] WARNING: Could not find gpod
1508423397.708852 [gpodder.sync] WARNING: Could not find eyed3.mp3
1508423397.729301 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/dsk/src/gpodder/src/gpodder/gtkui/main.py", line 3686, in do_activate
    self.window = gPodder(self, self.bus_name, core.Core(UIConfig, model_class=Model), self.options)
  File "/dsk/src/gpodder/src/gpodder/core.py", line 47, in __init__
    gpodder.user_extensions = extensions.ExtensionManager(self)
  File "/dsk/src/gpodder/src/gpodder/extensions.py", line 331, in __init__
    container = ExtensionContainer(self, name, config, filename)
  File "/dsk/src/gpodder/src/gpodder/extensions.py", line 201, in __init__
    self.metadata = ExtensionMetadata(self, self._load_metadata(filename))
  File "/dsk/src/gpodder/src/gpodder/extensions.py", line 236, in _load_metadata
    extension_py = open(filename).read()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 189: ordinal not in range(128)

Traceback (most recent call last):
  File "/dsk/src/gpodder/src/gpodder/gtkui/main.py", line 3686, in do_activate
    self.window = gPodder(self, self.bus_name, core.Core(UIConfig, model_class=Model), self.options)
  File "/dsk/src/gpodder/src/gpodder/core.py", line 47, in __init__
    gpodder.user_extensions = extensions.ExtensionManager(self)
  File "/dsk/src/gpodder/src/gpodder/extensions.py", line 331, in __init__
    container = ExtensionContainer(self, name, config, filename)
  File "/dsk/src/gpodder/src/gpodder/extensions.py", line 201, in __init__
    self.metadata = ExtensionMetadata(self, self._load_metadata(filename))
[...]

PR #352, opened 2017-10-19
Fix ambiguous encoding #351

ideasman42:


src/gpodder/extensions.py
```diff
@@ -289,10 +289,10 @@ def load_extension(self):
             return
 
         basename, extension = os.path.splitext(os.path.basename(self.filename))
-        fp = open(self.filename, 'r')
+        fp = open(self.filename, 'rb')
```
elelay: Why open `.py` as rb?

src/gpodder/extensions.py
```diff
@@ -233,7 +233,7 @@ def _load_metadata(self, filename):
         if not filename or not os.path.exists(filename):
             return {}
 
-        extension_py = open(filename).read()
+        extension_py = open(filename, encoding='utf-8').read()
```
elelay: Looks good to me.

ideasman42:
Right, and if people want to share extensions it's best to stick to one encoding.
(maybe it could be detected, although I'm not sure it's worth it).

Note that another way to handle this is using Pythons AST to parse the information instead of reading as text. But think this isn't necessary.

elelay:
@ideasman42 can you explain why you change open 'r' to open 'rb'?

ideasman42:
@elelay because for loading as data for a module, there is no need to use string/encoding.
It can load as data instead (It's a little faster and less complicated)

thp:
This is not the way to go. The right filesystem encoding needs to be used. I'll comment further in #351.

PR #357, opened 2017-10-20
Updates the Help destination url

adamvoss:
I think the docs landing page is a little more useful than the user manual, especially since there is currently no backwards navigation from the user manual.   Also, trivial cleanup of a method name.


elelay:
Thanks!
Would you like to do it for master as well?

thp:
Here for master: https://github.com/gpodder/gpodder/pull/365

Issue #377, opened 2018-01-07
opml export error


simbiotiqu:
On version 3.10 I get an error when trying to export an opml file. The tmp gets created but stays empty. Gui error says check writ permissions but console gives this:
```
1515283760.099278 [gpodder.opml] ERROR: Could not open file for writing: /home/simbiotiqu/Downloads/podcast_sub.opml
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/gpodder/opml.py", line 190, in write
    fp.write(data)
TypeError: write() argument must be str, not bytes
```

elelay:
@simbiotiqu any chance that you modify the pull-request?
Thanks,

simbiotiqu:
sorry didn't see the request
changed the pr

elelay:
@simbiotiqu thanks for the update.
I was eager to have the fix merged in master, that's why I've been a bit pushy.

PR #378, opened 2018-01-07
Update opml.py

simbiotiqu:
to fix opml export error #377


elelay:
Merged.
Many thanks :-)

PR #388, opened 2018-01-28
Set 'Ctrl+E' as the shortcut for episode's "Toggle New Status"

eunikolsky:
Verified that it works on OSX 10.11.6 (where the shortcut is actually
'Cmd+E') for one and multiple selected episodes.

(created the patch into the `gtk2` branch because the latest released app package is for `3.9.6`; when I can run the `3.10.0` I'll create a patch into `master` as well)


thp:
Thanks! Note that I'm not sure when/if we will do a 3.9.x release still. But for macOS, a version of 3.10.x is in the works I hear.

PR #390, opened 2018-01-30
Fix W293 blank line contains whitespace

MarkusHackspacher:
Remove whitespace at blank lines.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [pycodestyle]
 count=1
-select = W1, W293, E6, E8, E9
+select = W1, W293, E9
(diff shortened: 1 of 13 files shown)


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [pycodestyle]
 count=1
-select = W1
+select = W1, W293, E6, E8, E9
```
elelay: What's E6, E8 ?

MarkusHackspacher: E6 and E8 is just a reminder, no test case. 
I had test in my local machine the case from E1 to E9 and in this cases I got no Error back.

elelay:
Merged, Thanks :-)

PR #392, opened 2018-02-01
Fix W291 trailing whitespace

MarkusHackspacher:
Remove trailing whitespace.


elelay:
Merged, thanks.
If you intend to continue fixing pycodestyle errors, please discuss them beforehand at #393, so we don't reject a PR after you spent time modifying a large number of files.

Issue #393, opened 2018-02-02
pycodestyle checks


elelay:
@MarkusHackspacher has begun fixing whitespace errors and the like and enabling pycodestyle checks.
Going further, additional checks should be discussed here before submitting pull requests, to prevent loosing time  on fixing rules we don't agree on.

Here is current list of errors with the default config (first column is error count):
Error codes are here : https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
```
> pycodestyle share src/gpodder tools *.py
```
- [x]   1     E114 indentation is not a multiple of four (comment)
- [x]   1     E227 missing whitespace around bitwise or shift operator
- [x]   1     E271 multiple spaces after keyword
- [x]   1     E272 multiple spaces before keyword
- [x]   1     E703 statement ends with a semicolon
- [x]   1     E711 comparison to None should be 'if cond is not None:'
- [x]   1     E721 do not compare types, use 'isinstance()'
- [x]   2     E129 visually indented line with same indent as next logical line
- [x]   2     E266 too many leading '#' for block comment
- [x]   2     E712 comparison to True should be 'if cond is True:' or 'if cond:'
- [x]   3     E125 continuation line with same indent as next logical line
- [x]   3     E211 whitespace before '('
- [x]   3     E222 multiple spaces after operator
- [x]   3     E713 test for membership should be 'not in'
- [x]   8     E115 expected an indented block (comment)
- [x]   8     E124 closing bracket does not match visual indentation
- [x]   8     E262 inline comment should start with '# '
- [x]  11     E228 missing whitespace around modulo operator
- [x]  11     E401 multiple imports on one line
- [x]  13     E305 expected 2 blank lines after class or function definition, found 1
- [x]  13     E306 expected 1 blank line before a nested definition, found 0
- [x]  13     E731 do not assign a lambda expression, use a def
- [x]  18     E221 multiple spaces before operator
- [ ]  21     ~~E127 continuation line over-indented for visual indent~~
- [x]  27     E202 whitespace before ')'
- [ ]  31     E701 multiple statements on one line (colon)
- [x]  33     E303 too many blank lines (2)
- [x]  34     E225 missing whitespace around operator
- [x]  48     E111 indentation is not a multiple of four
- [ ]  50     E722 do not use bare except
- [x]  58     E261 at least two spaces before inline comment
- [x]  64     E251 unexpected spaces around keyword / parameter equals
- [x]  67     W391 blank line at end of file
[...]

PR #400, opened 2018-02-10
fix E305 expected 2 blank lines after class or function definition

MarkusHackspacher:

In the first PR "fix E305" I found some errors and make this fine now.


elelay:
merged. Thanks!

PR #402, opened 2018-02-11
Extension to run a command on download

elelay:
Not sure if it should go in share/gpodder/extensions (it's usable as is) or share/gpodder/examples.


thp:
Ah, didn't see this. IMHO should be good to go in `share/gpodder/extensions`, ready to merge from my side.

elelay:
Great, Thanks!

PR #409, opened 2018-02-17
fix E262 inline comment should start with '# '

MarkusHackspacher:


elelay:
Merged, thanks!
Don't forget to add the error to setup.cfg once it's fixed ;-)

PR #417, opened 2018-03-01
fix E111 indentation is not a multiple of four

MarkusHackspacher:
E111 to E119 are solved, to check this is E11 im setup.cfg.


elelay:
Merged. Thanks!

PR #418, opened 2018-03-04
fix W503 line break before binary operator

MarkusHackspacher:

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -269,9 +269,9 @@ class PodcastEpisode(PodcastModelObject):
                     return title
 
         # "#001: Title" -> "001: Title"
-        if (not self.parent._common_prefix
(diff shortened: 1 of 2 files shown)


src/gpodder/model.py
```diff
@@ -269,9 +269,9 @@ def trimmed_title(self):
                     return title
 
         # "#001: Title" -> "001: Title"
-        if (not self.parent._common_prefix
-                and re.match('^#\d+: ', self.title)
-                and len(self.title)-1 > LEFTOVER_MIN):
+        if (not self.parent._common_prefix and
```
elelay: not found of that one, because it looks like not applies to all lines.
Better
```
        if (
                not self.parent._common_prefix and
                re.match('^#\d+: ', self.title) and
                len(self.title)-1 > LEFTOVER_MIN):
```

Issue #384, opened 2018-01-17
Found a way to add youtube channels, users and playlists without an api key


ghost:
i found this on stackoverflow, all you have to do is parse the channel id, username or playlist number and swap it into the urls below

https://www.youtube.com/feeds/videos.xml?channel_id=CHANNELID
https://www.youtube.com/feeds/videos.xml?user=USERNAME
https://www.youtube.com/feeds/videos.xml?playlist_id=YOUR_YOUTUBE_PLAYLIST_NUMBER

is it possible for this to be implemented into the application soon?

predator8bit:
Too bad it doesn't work any more. It used to play just fine, but now I have to use the 'Open website' plugin and then use 'open with...' in my browser to play my videos with mpv.
![player](https://user-images.githubusercontent.com/27356954/36075635-f766f3c2-0f51-11e8-8c1e-5809d6975123.png)
Arch Linux + Gpodder 3.10.0

XIncognito10:
hmm thats odd. All 3 methods are working for me right now

XIncognito10:
I'm actually watching a youtube channel that i subscribed to about 15 minutes ago

predator8bit:
Opened a separate issue for my problem.

pantar85:
have been doing this for as long as i have been using gpodder and never needed a api key.

go to channel i want to subscribe to and beside the home tab, select the videos tab. copy and paste this link.

will look something like https://www.youtube.com/user/CHANNEL_NAME/videos

has always worked for me since i first started using gpodder back in 2012/2013.

always presumed the youtube api key requirement was for 4k videos or something- never bothered with it.

XIncognito10:
Should i try to work on a PR that implements this and phases out the youtube api?

elelay:
Yes, please! Don't hesitate to ask if you have questions on the way ;-)
The YouTube API is to convert things like https://www.youtube.com/user/sixguntv/videos to https://www.youtube.com/channel/UCaITiY0rLVsRr71dP-kQgEg but I agree it's not needed for the (majority?) of urls with the id.

XIncognito10:
Where would i add the code to change it? i tried podcastparser but that doesnt work

elelay:
Right into https://github.com/gpodder/gpodder/blob/master/src/gpodder/youtube.py I guess...

PR #437, opened 2018-04-08
Implements #384: Youtube Feeds without API

XIncognito10:
Implemented api-less youtube feeds (#384)
users, playlists and channels are supported

### DIFF
--- a/src/gpodder/youtube.py
+++ b/src/gpodder/youtube.py
@@ -239,7 +239,7 @@ def get_cover(url):
                 logger.debug('Youtube cover art for {} is: {}'.format(url, parser.url))
                 return parser.url
 
-        except Exception as e:
+        except Exception:
             logger.warning('Could not retrieve cover art', exc_info=True)
 
 
--- a/src/gpodder/youtube.py
+++ b/src/gpodder/youtube.py
@@ -20,9 +20,7 @@
 #  Justin Forest <justin.forest@gmail.com> 2008-10-13
 #
 
-
 import gpodder
-
 from gpodder import util
 
 import os.path
@@ -73,8 +71,8 @@ formats_dict = dict(formats)
 V3_API_ENDPOINT = 'https://www.googleapis.com/youtube/v3'
 CHANNEL_VIDEOS_XML = 'https://www.youtube.com/feeds/videos.xml'
 
-
-class YouTubeError(Exception): pass
+class YouTubeError(Exception):
+    pass
 
 
 def get_fmt_ids(youtube_config):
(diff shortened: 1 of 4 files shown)


XIncognito10:
There you go, let me know if you need anything else

elelay:
Code is good.
Testing it more I noticed that the channel image is not retrieved when subscribing to e.g. https://www.youtube.com/user/sixguntv/playlists in the new code.
This is because there is no channel id for `get_real_cover()` in `youtube.py`.
It could maybe be fixed by modifying `get_real_cover()` but I noticed that the feed from e.g. https://www.youtube.com/feeds/videos.xml?user=sixguntv does contain the channel id.
Using your method to resolve the feed, then fetching and extracting `yt:channelId` when subscribing would be better, what do you think ?

Excerpt from the feed:
```xml
<feed xmlns:yt="http://www.youtube.com/xml/schemas/2015" xmlns:media="http://search.yahoo.com/mrss/" xmlns="http://www.w3.org/2005/Atom">
 <link rel="self" href="http://www.youtube.com/feeds/videos.xml?user=sixguntv"/>
 <id>yt:channel:UCaITiY0rLVsRr71dP-kQgEg</id>
 <yt:channelId>UCaITiY0rLVsRr71dP-kQgEg</yt:channelId>
 <title>SixgunTV</title>
```

XIncognito10:
Yeah, I'd agree. It seems it would be more reliable since all the feeds, would come with that id

elelay:
Yes, with pleasure, if you're willing to ;-)
You can use `util.urlopen` to load the feed, and maybe `xml.etree.ElementTree.parse` to parse the feed?

XIncognito10:
There you go, i modified get_real_cover() to use the new api. I also took the liberty of replacing the regular expression with a dedicated html parser, but you can let me know if you would like to revert the changes to the latter

src/gpodder/youtube.py
```diff
@@ -209,26 +203,44 @@ def return_user_feed(url, channel):
...
+            channel_url = 'https://www.youtube.com/channel/{}'.format(channel_id)
+            html_data = util.urlopen(channel_url).read().decode('utf-8')
+            parser = YouTubeHTMLCoverParser()
+            parser.feed(html_data)
+            if parser.url:
+                logger.debug('Youtube cover art for {} is: {}'.format(url, parser.url))
+                return parser.url
 
-    return for_each_feed_pattern(return_user_cover, url, None)
+        except Exception as e:
```
elelay: no need for `as e`

XIncognito10:
im not sure i follow

XIncognito10:
Do you mean like converting all the feeds to their respective channels? If so i didnt do it because with the current setup, the user can have playlist subscriptions different from user subscriptions different from channel subscriptions, instead of all of them being channel subscriptions. I think thats more flexible. Is that what you meant?

elelay:
My bad, I thought each playlist, user, channel would get a different /channelId/. Let me investigate some more...

XIncognito10:
sure! no problem

elelay:
Merged, thanks!

PR #441, opened 2018-04-17
fix E225 missing whitespace around operator

MarkusHackspacher:


elelay:
Merged, thanks!

PR #442, opened 2018-04-17
Commands for single episodes downloading

nmaggioni:
[GPodder v4](https://github.com/gpodder/gpodder-core)'s ability to operate on single episodes would be really handy for a project of mine ([DashPodder](https://github.com/nmaggioni/DashPodder)), but as @thp [hinted](https://github.com/gpodder/gpodder-core/issues/22) it would be better if I proposed such capability on the V3 branch - so here it is! :)

This is a (somewhat rudimentary) approach to downloading single episodes basing on their GUID URL which is already known to the internal data storage; **suggestions on how to improve the commands' syntax or GUIDs display are welcome**.

On the podcasts I've tested GUIDs are either the link to the episode's MP3 file or a shorthand to its page on the podcast's website - the only relevant thing about this is the length of the value: it could get large and cause troubles when being displayed.

---

### Commands changes

+ `download [URL]` :arrow_right: `download [URL] [GUID]` - If a _GUID_ is given, only corresponding episode(s) are downloaded.
+ `episodes [URL]` :arrow_right: `episodes [URL] [SHOWGUID]` - If _SHOWGUID_ is specified **(can be any value)**, episodes' GUIDs are shown alongside the episode's number in the listing.

### DIFF
--- a/bin/gpo
+++ b/bin/gpo
@@ -48,8 +48,10 @@
   - Episode management -
 
     download [URL]             Download new episodes (all or only from URL)
+    downloadguid URL GUID      Download single episode
     pending [URL]              List new episodes (all or only from URL)
     episodes [URL]             List episodes (all or only from URL)
+    episodesguid URL           List episodes with GUIDs
 
   - Configuration -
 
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!

PR #443, opened 2018-04-19
Use a command line option instead of "any value is true"

thp:
A small amendment I'd like to make to #442.


elelay:
You're right Thomas, it's better to have a proper option.
Cheers,

Issue #420, opened 2018-03-08
Unable to delete partial download (WindowsError 32) with v3.9.6


Vhati:
I'm using the [PortableApps](https://portableapps.com/apps/internet/gpodder_portable) 3.9.6 release on Win7.

When I download several new episodes at once (from various podcasts), limiting the queue to one transfer at a time, I sometimes get this error.

	2018-03-08 15:00:12,631 [gpodder.download] ERROR: Download failed: [Error 32] The process cannot access the file because it is being used by another process: '[...]\\gPodderPortable\\Data\\config\\Downloads\\Nature Podcast\\media.mp3.partial'
	Traceback (most recent call last):
	  File "[...]\gPodderPortable\App\gPodder\src\gpodder\download.py", line 864, in run
		shutil.move(self.tempname, self.filename)
	  File "[...]\gPodderPortable\App\Python\Lib\shutil.py", line 303, in move
		os.unlink(src)
	WindowsError: [Error 32] The process cannot access the file because it is being used by another process: '[...]\\gPodderPortable\\Data\\config\\Downloads\\Nature Podcast\\media.mp3.partial'

I'm new to gPodder and haven't pinned down the curcumstances yet (whether the queue slot matters, etc.). In the latest occurrence, I think the problem episode was queued in slot 4 out of 6.

.
This is the relevant code ([GitHub link](https://github.com/gpodder/gpodder/blob/35fcc8b81aa434ad0a651619633a5d748ff0fd58/src/gpodder/download.py#L859-L867)).

	# Re-evaluate filename and tempname to take care of podcast renames
	# while downloads are running (which will change both file names)
	self.filename = self.__episode.local_filename(create=False)
	self.tempname = os.path.join(os.path.dirname(self.filename),
		os.path.basename(self.tempname))
	shutil.move(self.tempname, self.filename)

	# Model- and database-related updates after a download has finished
	self.__episode.on_downloaded(self.filename)

A copy IS made with the episode's filename, but "media.mp3.partial" DOES NOT disappear. Thus the move() call fails.

That 'in use' error prevents gPodder from acknowledging the successful download (no musical note icon), and my custom "on_downloaded" extension doesn't run.

The partial file's continued existence, in turn, prevents re-download attempts ("[gpodder.download] ERROR: HTTP 416: [...] Requested Range Not Satisfiable"), even if I tell gPodder to 'delete' the episode.

If I go through the filesystem and delete the partial file myself, I can then re-download the episode without errors.

[...]

PR #449, opened 2018-04-29
Fix file descriptor leaks into child processes

Vhati:
This pull request modifies Popen() calls in spawning panucci, spawning other players, util.gui_open(), and util.bluetooth_send_file().

Without this edit (an extra arg to Popen() calls), file handles leak into child processes. Those handles then remain in use, even after gPodder is done with them.

As discovered in issue #420 , this is problematic when gPodder spawns a player while a download is in progress. The partial file will stay locked after gPodder completes the transfer, causing an "in use" error when renaming the file from ".partial" to ".mp3". That happens even if the player was spawned to play some other file - the handle was inherited, not an arg.

.
Notes:

* On Windows, util.gui_open() relies on os.startfile(). I have not touched that and do not know if file descriptor (handle) inheritance is an issue there.
* On Windows, disabling inheritance **also** disables standard in/out/err redirection. Child processes that need to redirect streams cannot be fixed this way and will continue to have lock issues. However, short-lived requests for info from external commands should be tolerable.
* On Unix, standard streams are exempted and will function normally. So this fix could be applied in redirect-dependent code intended only to run on Unix.
* **Warning:** The following extensions have redirection. This fix was not applied to them. An alternate means of logging errors from long-lived external commands may be needed.
    * enqueue_in_mediaplayer.py
    * normalize_audio.py
    * rockbox_convert2mp4.py
    * ubuntu_unity.py
    * video_converter.py
* I recommend a notice about Popen() be included in documentation/examples for extenson developers.
* I hadn't set up a dev environment to test this commit, but it was a trivial edit (based on tests I did with a v3.9.6 PortableApps build for Windows).

.
[Article](https://docs.python.org/2/library/subprocess.html#subprocess.Popen): Python API: Subprocess.popen()

> If close_fds is true, all file descriptors except 0, 1 and 2 will be closed before the child process is executed. (Unix only). Or, on Windows, if close_fds is true then no handles will be inherited by the child process. Note that on Windows, you cannot set close_fds to true and also redirect the standard handles by setting stdin, stdout or stderr.

### DIFF
--- a/share/gpodder/examples/hello_world.py
+++ b/share/gpodder/examples/hello_world.py
@@ -1,9 +1,9 @@
 
-# Use a logger for debug output - this will be managed by gPodder
+# Use a logger for debug output - this will be managed by gPodder.
 import logging
 logger = logging.getLogger(__name__)
 
-# Provide some metadata that will be displayed in the gPodder GUI
+# Provide some metadata that will be displayed in the gPodder GUI.
 __title__ = 'Hello World Extension'
 __description__ = 'Explain in one sentence what this extension does.'
 __only_for__ = 'gtk, cli'
@@ -11,9 +11,9 @@ __authors__ = 'Thomas Perl <m@thp.io>'
 
 
 class gPodderExtension:
-    # The extension will be instantiated the first time it's used
+    # The extension will be instantiated the first time it's used.
     # You can do some sanity checks here and raise an Exception if
-    # you want to prevent the extension from being loaded..
+    # you want to prevent the extension from being loaded.
     def __init__(self, container):
         self.container = container
 
@@ -40,7 +40,7 @@ class gPodderExtension:
 
     def on_ui_object_available(self, name, ui_object):
         """
-        Called by gPodder when ui is ready
+        Called by gPodder when ui is ready.
         """
         if name == 'gpodder-gtk':
             self.gpodder = ui_object
@@ -50,3 +50,24 @@ class gPodderExtension:
 
     def say_hello_cb(self):
         self.gpodder.notification("Hello Extension", "Message", widget=self.gpodder.main_window)
+
+
+# Concurrency Warning
+#
+# When using subprocess.Popen() to spawn a long-lived external command,
+# such as ffmpeg, be sure to include the "close_fds=True" argument.
+#
+# https://docs.python.org/3/library/subprocess.html#subprocess.Popen
+#
+# This is expecially important for extensions responding to
+# on_episode_downloaded(), which runs whenever a download finishes.
+#
+# Otherwise that process will inherit ALL file descriptors gPodder
+# happens to have open at the moment (like other active downloads).
+# Those files will remain 'in-use' until that process exits, a race
+# condition which prevents gPodder from renaming or deleting them.
+#
+# Caveat: On Windows, you cannot set close_fds to true and also
+# redirect the standard handles (stdin, stdout or stderr). To collect
+# output/errors from long-lived external commands, it may be necessary
+# to create a (temp) log file and read it afterward.
--- a/share/gpodder/extensions/command_on_download.py
+++ b/share/gpodder/extensions/command_on_download.py
@@ -64,7 +64,7 @@ class gPodderExtension:
         env = os.environ.copy()
         env.update(info)
 
-        proc = subprocess.Popen(command, shell=True, env=env)
+        proc = subprocess.Popen(command, shell=True, env=env, close_fds=True)
         proc.wait()
         if proc.returncode == 0:
             logger.info("%s succeeded", command)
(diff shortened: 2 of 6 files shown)


elelay:
Thanks for your PR!
I'll look into it this week.

elelay:
a bit late, but I've looked into your PR. It's great.
Only remaining issue is:

> Warning: The following extensions have redirection. This fix was not applied to them. An alternate means of logging errors from long-lived external commands may be needed.
>
>    enqueue_in_mediaplayer.py

redirection only for freedesktop, not win32, so it's ok

>    normalize_audio.py
>    rockbox_convert2mp4.py

problem with those 2. **Shall we add a special case with no redirect on windows or simply keep redirects nowing that they will have no effect?**

>    ubuntu_unity.py

not win32, so it's ok

>    video_converter.py

same problem as normalize_audio.py

And
> I recommend a notice about Popen() be included in documentation/examples for extenson developers.

Please add a comment about that in `share/gpodder/examples/hello_world.py`.

Vhati:
I added the warning comment. Since it was a corner case, not functionally relevant to what hello_world's simple code demonstrates, I stuck it at the end.

> Shall we add a special case with no redirect on windows or simply keep redirects knowing that they will have no effect?

Hrm. Mysteriously silent commands are frustrating.

It would've been tidy to create log files via tempfile.mkstemp(), tell the command to fill them, read afterward, and delete. Unfortunately, those commands (normalize-* and ffmpeg) don't have an arg for log files, so you'd have to do shell=True for "1> tmp-out.txt 2>tmp-err.txt". Ick.

**Suggestion:** Add "close_fds=True", let Windows mute redirection, and log more verbosely about failures.

gPodder should log the command line that resulted in non-zero exit status. Conditionally on Windows, log a message that stdout and stderr weren't preserved. Troubleshooters can run the command manually to see what happened.

Vhati:
Grr.

    p = subprocess.Popen(["cmd", "/?"], close_fds=True, stdout=subprocess.PIPE)
    
    ValueError: close_fds is not supported on Windows platforms if you redirect stdin/stdout/stderr

Okay, fine. Include "close_fds=True" in all cases, but conditionally call Popen() on Windows **without any redirection** to dodge the ValueError. Then, when the returncode is non-zero on Windows, conditionally log about that silence.

elelay:
Thanks for the precisions. Following your latest comment, creating a wrapper that does that and replacing every occurence of `subprocess.Popen` with `util.popen` seems a good way to solve it. It could return a subclass of `subprocess.Popen` to log about that silence in `wait()` and `poll()`.

Vhati:
> util.popen seems a good way to solve it.

Agreed.

elelay:
Great! Would you like to do it?

Vhati:
[popen_test_v1_py.txt](https://github.com/gpodder/gpodder/files/2019960/popen_test_v1_py.txt)

I'm a novice at subclassing in Python. How's this?

elelay:
That's great! I'll take it.
But I would wrap `communicate(input)` to throw an exception if `__redirects_disabled` and `input` is not `None`. Otherwise it gets stuck. See attached popen_test_v1_1.py

[popen_test_v1_1.py.txt](https://github.com/gpodder/gpodder/files/2020189/popen_test_v1_1.py.txt)

Vhati:
> throw an exception if __redirects_disabled and input is not None. Otherwise it gets stuck.

That particular command (findstr hello) is waiting on stdin. Without redirection, that's the console. You can type to feed it. If communicate()'s arg is None or absent, the command would still wait.

So Python's API was wise in raising the ValueError early, in the constructor, where redirection is first requested.

[popen_test_v1_2.py.txt](https://github.com/gpodder/gpodder/files/2020953/popen_test_v1_2.py.txt)

This revision no longer tries to disable redirection. It catches the ValueError, inserts a preemptive explanation into the log, and re-raises.

Vhati:
Looks reasonable to my eyeballs.

In hello_world.py, you added on [line 68](https://github.com/gpodder/gpodder/commit/b307b41a4080bd927317f991cef8e8d15f817e69#diff-be62eca97f1399c571859fc0e6c3e31dL68) that forgetting close_fds creates a filesystem locking race condition only on Windows.

Was that aspect platform specific? I wasn't aware. I happened to report it on Windows and hadn't checked others.
.
.
It seems those tests are getting kept for posterity. In util.py [line 2008](https://github.com/gpodder/gpodder/commit/b307b41a4080bd927317f991cef8e8d15f817e69#diff-82ffe7799b265c763fb1ad8971cec69fR2008), that descriptive comment had not been revised to encompass the pipe commands. "Commands that will complain on stderr." should be something like "Problematic commands."

Do those test print() calls need to be changed now that the tests run within gPodder?
.
.
I like the way you multiplied the "unavailable" stdout/stderr stand-ins, btw.
.
.
/ Having args printing was nice, but I guess that does makes more sense in boilerplate error handling, as desired.  ; )

elelay:
Thanks for the review!

> Was that aspect platform specific? I wasn't aware. I happened to report it on Windows and hadn't checked others.
Having the file open on Linux doesn't prevent it from being renamed or deleted, AFAIK. This is why I think it's Windows specific.

 I'll adjust the "problematic commands" comment.

PR #450, opened 2018-05-01
fix E124 closing bracket does not match visual indentation

MarkusHackspacher:
add E124 and E220 to E229 as E22 in setup.cfg.


elelay:
Merged, thanks!

PR #451, opened 2018-05-05
fix E401 multiple imports on one line

MarkusHackspacher:


elelay:
merged, thanks !

PR #454, opened 2018-05-06
fix E731 do not assign a lambda expression, use a def

MarkusHackspacher:


elelay:
Merged, thanks

PR #456, opened 2018-05-07
Command for single episode deletion

nmaggioni:
Pretty much a followup to #442, this brings in a new command for deleting single episodes given their feed's URL and their GUID's.

In order to avoid possible GUID duplicates _(two feeds could theoretically have equal values somewhere between their episodes list)_ the feed URL is mandatory.

---

### New commands

+ `delete [URL] [GUID]` - Delete from feed at URL an episode with given GUID.


elelay:
Merged, thanks!

PR #457, opened 2018-05-08
 Fix E502 the backslash is redundant between brackets 

MarkusHackspacher:

E502 is not in setup.cfg, because in #454  I make a change, too.


elelay:
Merged, thanks.
I've added E502 to the config

PR #458, opened 2018-05-10
Fix unable to drag and drop from episode list

jawbkr:
Attempting to drag and drop an episode to the playlist of an external media player is no longer working in 3.10.x and causes:

```
  File "/home/jawbkr/github/gpodder/src/gpodder/gtkui/main.py", line 1023, in drag_data_get
    selection_data.set(selection_data.target, 8, '\r\n'.join(uris))
AttributeError: 'SelectionData' object has no attribute 'target'
```


elelay:
Good catch, thanks!

PR #461, opened 2018-05-17
Codestyle e26x

MarkusHackspacher:
fix E261 at least two spaces before inline comment 
fix E265 block comment should start with '# ' 

see #393


elelay:
Merged, thanks

elelay:
I notice an added 

> TODO: try .set_sensitive(false) to gray-out

in main.py. `set_enabled` does that. I've removed the TODO, but thanks :-)

PR #463, opened 2018-05-21
fix E201 whitespace after '('

MarkusHackspacher:
see #393 

part 1


elelay:
Merged, thanks.
Looking forward for part2 ;-)

PR #487, opened 2018-06-25
Avoid python-3.7 reserved keyword "async"

tmzullinger:
The "async" keyword was added in python-3.7¹.  Using it throws
SyntaxError exception.

¹ https://docs.python.org/3.7/whatsnew/3.7.html
  https://bugs.python.org/issue30406

### DIFF
--- a/src/gpodder/gtkui/services.py
+++ b/src/gpodder/gtkui/services.py
@@ -101,7 +101,7 @@ class CoverDownloader(ObservableService):
         """
         self.request_cover(channel, custom_url)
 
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!
Does gPodder otherwise run under python 3.7?

tmzullinger:
Sorry for the delayed reply.  I don't know if gPodder has any other issues with python-3.7, unfortunately.  But once the current python-3.7 rebuild in fedora's rawhide branch is completed, it should be easier to test.  Thanks.

PR #488, opened 2018-07-01
Avoid async keyword for Python 3.6

thp:
Patch adapted from Debian's packaging. Original patch author: @tmancill

https://sources.debian.org/patches/gpodder/3.10.3-2/async-is-a-keyword.patch/

Related Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902794

### DIFF
--- a/src/gpodder/gtkui/services.py
+++ b/src/gpodder/gtkui/services.py
@@ -101,7 +101,7 @@ class CoverDownloader(ObservableService):
         """
         self.request_cover(channel, custom_url)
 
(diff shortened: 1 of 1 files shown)


tmancill:
@thp Thanks for putting together a PR for this so quickly - you beat me to it.

elelay:
Isn't it #487?
The diff seems identical

thp:
Yes, it's the same. I didn't update my local master branch, that's why I didn't see it.

PR #505, opened 2018-07-31
isort imports

MarkusHackspacher:
see #393


elelay:
It is not reported by isort.
Also mutagen is a library. gpodder is this project. So it should be before.
Maybe you should install mutagen and rerun isort.

MarkusHackspacher:
Yes, your right, I install mutagen and it changed back. But on travis is not mutagen install and show an error. Line 710 and 711: https://travis-ci.org/gpodder/gpodder/builds/409960267

PR #510, opened 2018-08-04
Change desciption from Run Command on Download

TZocker:
Like 
https://gpodder.github.io/docs/extensions/commandondownload.html

I think is't better than to explain it with itself.

greetz

PR #512, opened 2018-08-04
Update de.po

TZocker:
wrong msgid from #510

sry my fault wrong order...


elelay:
Merged, thanks!
Would you be interested in an email before releases so that you have the opportunity to update the translation (no commitment, only an opportunity)? I'm trying to figure out a way to keep translations up to date in the future...

TZocker:
https://github.com/gpodder/gpodder/compare/master...TZocker:patch-1

PR #517, opened 2018-08-11
Updated Dutch translation

Vistaus:
Fully updated the translation; it's 100% translated now :)


elelay:
Merged, Thanks!
Would you be interested in an email before releases so that you have the opportunity to update the translation (no commitment, only an opportunity)? I'm trying to figure out a way to keep translations up to date in the future...

Vistaus:
You're welcome :)

And yes, I'm interested in receiving an email so that I can update the translation before releases :)

PR #522, opened 2018-08-14
fix E127 continuation line over-indented for visual indent

MarkusHackspacher:
see #393

### DIFF
--- a/bin/gpo
+++ b/bin/gpo
@@ -405,8 +405,10 @@ class gPodderCli(object):
 
             return '   '
 
-        episodes = ('%3d.%s %s %s' % (i + 1, ((' %s' % e.guid) if
-                                              show_guid else ''),
+        def guid_str(episode):
+            return ((' %s' % episode.guid) if show_guid else '')
+
+        episodes = ('%3d.%s %s %s' % (i + 1, guid_str(e),
                                       status_str(e), e.title)
                     for i, e in enumerate(podcast.get_all_episodes()))
         return episodes
@@ -815,8 +817,8 @@ class gPodderCli(object):
         defaults = defaults or ()
         minarg, maxarg = len(args) - len(defaults), len(args)
 
-        if len(command_line) < minarg or (len(command_line) > maxarg and
-                                          varargs is None):
+        if (len(command_line) < minarg or
+                (len(command_line) > maxarg and varargs is None)):
             self._error('Wrong argument count for %s.' % func.__name__)
             return False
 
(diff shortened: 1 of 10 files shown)


bin/gpo
```diff
@@ -170,15 +169,17 @@ class gPodderCli(object):
         self._model = self.core.model
 
         self._current_action = ''
-        self._commands = dict((name.rstrip('_'), func)
+        self._commands = dict(
+            (name.rstrip('_'), func)
```
elelay: why not keep it on same line?

bin/gpo
```diff
@@ -404,8 +405,10 @@ class gPodderCli(object):
 
             return '   '
 
-        episodes = ('%3d.%s %s %s' % (i + 1, ((' %s' % e.guid) if show_guid else ''), status_str(e), e.title)
-                for i, e in enumerate(podcast.get_all_episodes()))
+        episodes = ('%3d.%s %s %s' % (i + 1, ((' %s' % e.guid) if
+                                              show_guid else ''),
+                                      status_str(e), e.title)
+                    for i, e in enumerate(podcast.get_all_episodes()))
```
elelay: this is not readable. I'd put the string formatting of one episode in a local function, with the guid part in a local variable in it.

bin/gpo
```diff
@@ -815,7 +816,7 @@ class gPodderCli(object):
         minarg, maxarg = len(args) - len(defaults), len(args)
 
         if len(command_line) < minarg or (len(command_line) > maxarg and
-                varargs is None):
+                                          varargs is None):
```
elelay: can you split at the first or instead?

bin/gpo
MarkusHackspacher: Arguments on first line forbidden when not using vertical alignment.
https://www.python.org/dev/peps/pep-0008/#indentation

bin/gpo
MarkusHackspacher: I am not sure what you mean, 
had move this line so that is aligned with opening delimiter.

elelay: Sorry, I mean split it like this at the 'or' instead of the 'and':
```
 if len(command_line) < minarg or
         (len(command_line) > maxarg and varargs is None):
```

bin/gpo
elelay: My bad, I didn't see that the line was continued. It's ok as it is.

elelay:
Merged, thanks!

elelay:
Still 22 errors when I add E127 to setup.cfg?
Are you on your way to fixing them?
Thanks,

PR #527, opened 2018-08-20
Update de.po

TZocker:
Hi,

They are default settings...

    wrap at 79 is checked
    preserve formatting of existing files is checked

I think something in Poedit isn't working...


elelay:
Thanks,
that change is good, anyway.

Issue #556, opened 2018-10-07
error when trying to add a cover to an mp3 file


bohegeek:
When we choose to Tag Download file using Mutagen in the preference, putting the covert art doesn't work for mp3 file.

elelay:
Merged in 4d8954e.

PR #557, opened 2018-10-07
correction for issue #556

bohegeek:
the update is not working because the jpg file is not opened as a binary file.
The fix is tested on both OS X and Linux


bohegeek:
Correction for bug #556

elelay:
Merged, thanks!

PR #573, opened 2018-12-10
fix #563 add confirmation if file already exist

### LINKED ISSUE #563: When saving to local folder will overwrite files
When saving a episode to Local folder, gpodder doesn't check if a file already exists, and overwrite that file, this happens in the last version 3.10.5

bth:


bth:
I added the filename in the message with this commit: 635e2bb
Continuous integration failed on this file: share/gpodder/extensions/minimize_on_start.py but I didn't change it

elelay:
Nevermind about the check fail. I've fixed it in master.
Thanks,

elelay:
Merged, thanks!

PR #597, opened 2019-01-23
Extend `tagging` extension to modify the id3 version

geoffoxholm:
This PR enables users to modify the `id3` tag version of their podcasts. 

I added support for the `tagging` extension to force the `id3` version (and optionally not change any of the tags). I did this because my car does not support `id3` version `2.4`.


elelay:
Thanks for the submission.
What happens when the filen isn't MP3?

geoffoxholm:
Thanks for the suggestion. The `kwarg` only works if the file is an `mp3`. I've updated the PR to check if the file is of the correct type.

elelay:
Merged, thanks :-)

PR #609, opened 2019-02-18
Remove .ico format icon from hicolor/16x16

bugthing:
The presence of this file was causing the flatpak build to fail.
  https://github.com/flathub/org.gpodder.gpodder/pull/11
I have addressed this in the packaging process but ideally the file would not be present in the upstream repo (here)
I am not sure if the file is there are specific reason, so please forgive me if I have stepped on someone elses feature... please feel free to close this and tell me to "do one"  if this is PR is complete BS


elelay:
It seems indeed not used anymore.
Thanks,

PR #624, opened 2019-05-18
Add extension storage to channels and episodes

elelay:
Original use case: I'm writing a custom feed handler for OPDS (special kind of atom to distribute books). The way custom feed handlers work is each one is called until one returns not None. So my OPDS custom feed handler is called on every podcast. I can't find out if the feed is opds or not until I parse the feed. Too bad to have essentially done everything (download, parse) to throw it away because it's not OPDS. So I want to remember that a feed is or isn't opds.

This PR:
 - adds an *ext_data* column to podcast and episode, converted to a JsonConfig object when needed.
 - adds `get_ext_data(ext_name)` to podcast and episode, returning the relevant subtree of the JsonConfig
 - hooks into `save()` to serialize back the JsonConfig in the database when needed.

Alternatives: 
 - let extensions store this info in their configuration.  
  **pros**: don't have to extend the schema   
  **cons**: data is not really attached to the podcast. It needs to be cleaned when the user unsubscribes
 - go one step further and allow arbitrary tables per extensions, where my OPDS extension would declare a foreign key to podcasts and a bool column for *ignored*  
  **pros**: more efficient than a common json column for all extensions (compact, less parsing)  
  **cons**: harder to implement.

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -39,6 +39,7 @@ import gpodder
 import podcastparser
 from gpodder import (coverart, escapist_videos, feedcore, schema, util, vimeo,
                      youtube)
+from gpodder.jsonconfig import JsonConfig
 
 logger = logging.getLogger(__name__)
 
@@ -104,7 +105,10 @@ class PodcastModelObject(object):
     A generic base class for our podcast model providing common helper
     and utility functions.
     """
-    __slots__ = ('id', 'parent', 'children')
+    __slots__ = ('id', 'parent', 'children', '_parsed_ext_data', '_ext_data_dirty')
+
+    def __init__(self):
+        self._ext_data_dirty = False
 
     @classmethod
     def create_from_dict(cls, d, *args):
@@ -119,6 +123,24 @@ class PodcastModelObject(object):
             setattr(o, k, v)
 
         return o
+
+    def get_ext_data(self, extension):
+        """
+        Get extension-specific storage for this object.
+        It's a JsonConfigSubtree so setting attribute values will be saved.
+        Make sure to use a versionned key.
+        """
+        if not self._parsed_ext_data:
+            self._parsed_ext_data = JsonConfig(self.ext_data, on_key_changed=self._mark_ext_data_dirty)
+        return getattr(self._parsed_ext_data, extension)
+
+    def _mark_ext_data_dirty(self, name, old_value, new_value):
+        self._ext_data_dirty = True
+
+    def save_ext_data(self):
+        if self._ext_data_dirty:
+            self.ext_data = repr(self._parsed_ext_data)
+            self._ext_data_dirty = False
 
 
 class PodcastEpisode(PodcastModelObject):
@@ -199,6 +221,7 @@ class PodcastEpisode(PodcastModelObject):
         return None
 
     def __init__(self, channel):
+        super().__init__()
         self.parent = channel
         self.podcast_id = self.parent.id
         self.children = (None, None)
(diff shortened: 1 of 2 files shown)


thp:
So the OPDS handler is a subclass of the normal Atom feed handler? What if you make it so that the OPDS handler is always used, and when it's downloaded and parsed, you just call the superclass function (the normal Atom feed handler) to parse it as "normal" feed if you detect that it's not OPDS?

Alternatively, you could just store e.g. a JSON file or minidb file in `gpodder.home` with a "list of URLs that are not OPDS", then it is independent of the main database and doesn't affect the schema at all. This is how the `soundcloud` plugin does it: `self.cache_file = os.path.join(gpodder.home, 'Soundcloud')` (`class SoundcloudUser` in `src/gpodder/plugins/soundcloud.py`).

elelay:
I hear what you say about subclassing the normal feed handler to also handle the normal case, but I would like to let other custom handlers intercept the feed if it's not OPDS (I wrote the OPDS handler as a brand new class, where I tried urllib3 for download and direct sax instead of podcastparser rules, just to experiment, but that's off topic).

Thanks for pointing out that soundcloud already has a cache file outside the db. It seems very clear and easy to use. I see how I could copy the `__init__()` and `commit_cache()` code in the opds extension.

Would the soundcloud plugin benefit from this PR?
It would be even more convenient to store this cache for each subscription in the ext_data['soundcloud'] column of the database, since loading and parsing would be taken care of.
Storing the cache per podcast in the database would have the good property of automatically purging the cache on unsubscribe. It would allow easy purging of removed track metadata (remove those not seen), which is missing in current implementation, but easy enough to add (just group track metadata by username).
I don't think the episode's ext_data['soundcloud'] would be useful because existing episodes would have to be searched by url for each track.

I would like to invite more people to write small extensions to scratch their own itch, as a first step to contributing. So I'd like an easy way for an extension to:
 - save global parameters -- already available as the configuration;
 - attach data to podcast or episodes -- this PR's ext_data.

Yes, it will create an incompatibility with older gPodder releases (like the one in debian) because of the schema bump.
Do you think it is overkill (very little extensions will need this feature) or ill-suited (different extensions will want to store data differently)?
Do you think it would slow down the main function of gPodder?

elelay:
Initial OPDS support is here: https://github.com/gpodder/gpodder/blob/opds/share/gpodder/extensions/opds.py

thp:
Well, the schema breakage isn't that much of an issue I guess (as long as it's migrated from old version to new version seamlessly).

But still then, it could be done via separate tables (e.g. each plugin can create its own table of additional data) and those tables can then somehow be linked with the podcast / episodes table so that some generic code can do the clean-ups without modifying the main table structures.

Or the plugin could even have a function that is called as part of the cleanup (e.g. it would then do `DELETE FROM ext_soundcloud_podcast_data WHERE podcast_id = ?` and `DELETE FROM ext_soundcloud_episode_data WHERE episode_id = ?` or something like that).

But looking at it again from this perspective, the `ext_data` column that stores JSON data should be fine, as it can store free-form data, and there's not even a need to specially code the expiration of things. And looking further, `get_ext_data(ext_name)` is actually nice, so that each extension can just have a per-episode and per-podcast JSON object to store additional data in.

It might just be the cleanest solution to the whole thing, if you do go with that approach, maybe also convert the current Soundcloud code to use this :) :+1:

elelay:
OK. Trying to use it in soundcloud shows that JsonConfig is not useable as a cache:
 - it works with attributes, not dict-like keys. It splits attribute names on '.' => bad to store an url
 - there is no way to know if a key is present. It always returns a JsonConfigSubtree.
So I withdraw my proposal and will use a cache file in the opds extension.
Thanks for the feedback!

Issue #518, opened 2018-08-11
[notification] Call for translation update


elelay:
This issue will stay open.
I'll add a comment approximately one week before each release, to give translators a heads-up.

Review/update translations if you have time, write a comment if you want, be part of the team ;-)

elelay:
![subscribe](https://user-images.githubusercontent.com/222469/43993206-2ace5658-9d8a-11e8-90a0-2b5aa8495d2e.png)
@Vistaus let's try this instead of email.

Vistaus:
:+1:

TZocker:
👍

elelay:
I'd like to release 3.10.4 next week-end. Any pending translation?
@TZocker I've re-run make messages to update de.po with max line width=80 characters (the default and same as other translation files). Can you check in your poedit advanced preferences that
1. wrap at 79 is checked
2. preserve formatting of existing files is checked

Playing with these settings, I could reproduce your formatting by unchecking both.

elelay:
Dear translators,
I'd like to release 3.10.6 shortly (not before Dec. 28). If you would like to check translations one more time in 2018, now is the time :-)
Thanks,

elelay:
No translation changes for 3.10.7, so relax and enjoy the show ;-)
Cheers,

elelay:
Hi,
in 3.10.9 there seems to be only 2 changed messages to translate, with the inclusion of podcast titles in error messages...
Please update translations accordingly :-)
Thanks,

elelay:
Hi,
I'd like to release 3.10.10 this week-end or next week.
Please can you check the 2 `episode_website_context_menu.py` translations?
The action is now available also on the podcast, not only episode, so that changes the extension description slightly...
Thanks,

elelay:
Actually, there is also the description for the new youtube-dl extension...

TZocker:
https://github.com/gpodder/gpodder/pull/663

K4rBie:
#664

mhussaincov:
hi there,
this is not really gpodder related but it kinda is,
here goes:
currently videos are not downloading via youtube-dl, someone has
submitted an issue about the problem,
so your youtube-dl extention/ addon will not function till the issue
is fixed or a work around is put inplace.
just thought I would let you all be aware of the fact before people
would create issues thinking that the youtube-dl extention was at
fault.
I hope this helps someone,
Majid Hussain

[...]

PR #631, opened 2019-05-29
Update polish translation

K4rloss:
Full polish translation. Checked and working with the newest gPodder version, hopefully I didn't miss any mistakes.

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -1,8 +1,8 @@
-# SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PACKAGE package.
 #
 # Translators:
+# Karol Bieńkowski <karolbienkowski@protonmail.com>, 2019.
 #  <fklebczyk@gmail.com>, 2013.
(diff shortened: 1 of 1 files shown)


po/pl.po
```diff
@@ -101,23 +99,23 @@ msgid "(unknown)"
 msgstr "(nieznane)"
 
 #: src/gpodder/util.py:1372 src/gpodder/util.py:1394
-#, python-format
+#, fuzzy, python-format
```
elelay: fuzzy translations will be ignored at runtime: you want to remove them!

K4rloss: Oh, I see what happened: Poedit complained about *\r* signs at the end of every translation with different plural versions. So it tagged them as fuzzy. I don't really know if it something to worry about so I just set them as checked, because they are translated well.

elelay: I think they're fine with "\r" even if msgmerge and poedit complain.

K4rloss:
Fuzzy translations were fixed, my bad.

elelay:
Merged, thanks :-)

elelay:
If you would like to keep the Polish translation up to date, please subscribe to issue #518 to be notified when I plan a new release...

PR #659, opened 2019-08-24
Fix Win32 Powershell Notifications

Hiltronix:
On Windows 7, the Powershell notifications that gPodder  3.x is using are never deleted from the desktop.  The notification icons permanently stay visible in the taskbar system tray, untill your manually roll over them with the mouse cursor.

Changing the subprocess  from "run" to "Popen" so the fuction is non-blocking.

Non Win 10 Windows script will create it's own time delay, after which it removes the taskbar system tray icon, and then deletes it's own script file.

Removing the Python delete file, and have the Powershell script delete itself when it's done, because we want the script to be non-blocking and be removed from the temp folder after use.

I have not tested on Win 10 as I use Win 7, so I don't know if Win 10 has the same issues with Powershell permant  taskbar system tray icons.  So this fix is just for <Win10 script notifications.


elelay:
Thanks!
It looks good. Let me test it on Win10.

elelay:
Merged, thanks :-)

PR #667, opened 2019-09-11
Update Norwegian Nynorsk translation

huftis:
Here’s a complete Norwegian Nynorsk translation of gPodder. (I don’t why the strings are in a different order; the .po file was automatically updated when I did a ‘make install’.)


elelay:
Maybe it's because of the locale (LC_COLLATE) when running make messages.
Can you try running `LANG=en_GB.UTF-8 make messages` on gPodder master to see if strings are reordered?
When I run make messages myself (with this LANG) on your branch it reorders strings back...

PR #668, opened 2019-09-11
Embed previous msgids of changed/fuzzied messages

huftis:
When a string is changed (marking the translation for the old string
as ‘fuzzy’), the original string is now kept for comparison. This
makes life much easier for the translators, as they can easily see what
has changed in the string, and adjust their translation accordingly.
(Modern PO editors show the changes as a coloured (word) diff.)


elelay:
Great! I didn't know about it.
Thanks,

PR #675, opened 2019-09-15
Use locale-aware formatting for file sizes

### LINKED ISSUE #671: File sizes are not formatted according to user’s locale
The file sizes shown in episode list (and other places) are not formatted according to the user’s locale, i.e., they use a dot as the decimal separator instead of the locale-specific decimal separator.

For example, they are formatted like this:

`17.3 MiB`

For most European locales (e.g. ‘nn_NO.utf8’), the decimal separator should be a comma:

`17,3 MiB`

huftis:
For example, for locales that use decimal commas, the file sizes are
now formatted like ‘12,3 MiB’ instead of ‘12.3 MiB’.

Fixes #671.

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -581,7 +581,7 @@ def format_filesize(bytesize, use_si_units=False, digits=2):
             used_value = bytesize / float(value)
             used_unit = unit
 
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!

PR #676, opened 2019-09-15
Format number of episodes using (locale-aware) thousands separator

huftis:
The number of episodes shown in the ‘pill boxes’ in the podcast list
are now formatted like ‘12,345’ or ‘12 345’ (depending on the locale),
instead of ‘12345’.


elelay:
Merged, thanks!

PR #713, opened 2019-12-25
Deleted illegal character in gettext string

ciampix:


elelay:
Thanks for your submission!
There are 3 mixed changes (in the future, please open different merge requests for different subjects).

1. Good to see the Italian translation updated :-) If you want to keep it up to date, you can subscribe to #518 to be notified before each release.

2. .gitignore changes are fine

3. What's wrong with '\r' in gettext strings?

ciampix:
First of all: thanks and sorry I did it wrong ... I have to figure out how to do that commits right since I did all in different phases...
1) done
2) ok
3) </BR>a) are the only ones in all source 
  b) gettext complains that its use is illegal so the gettext utils emit errors for _all_ translations 
  c) I think that gettext utils manage the \r \n conversion for all platforms (Mac, Windows & Linux/Unix) if you use just the \n for all. Anyway it is forbidden according to the gettext manual and I imagine that the python libs comply with the standard gettext library
PS: there is something wrong with the make install... it changes the ownership of some files in place before copying into the right positions so, after you do a make install you have to chown all sources again to be able to continue working on them...

ciampix:
PPS: there are some other files to be gitignored, do you mind if I create another pull request for them?

elelay:
OK. Thanks for the detailed explanation.
Looking forward to PRs for additional .gitignore entries or to fix the install ;-)

PR #724, opened 2020-01-08
Save episode column sorting and position

auouymous:
This saves and restores changes the user makes to eposide column sorting and positions.


auouymous:
Thanks, I didn't know I could access them directly.

elelay:
Yes, the config object is a bit magic...

elelay:
Merged, thanks!

PR #722, opened 2020-01-04
Youtube fixes

elelay:
multiple fixes for Youtube.
Fixes #711 
Fixes #714 
Fixes #719

PR #726, opened 2020-01-11
fix #722 youtube error_message referenced before assignment

auouymous:
Initialize error_message to avoid exceptions on some python installations.


elelay:
Merged, thanks!

PR #728, opened 2020-01-11
Fix icon for deleted episodes

auouymous:
Deleted episodes show a broken icon and should be changed to 'edit-delete' like is used on the menus.

https://lazka.github.io/pgi-docs/Gtk-3.0/constants.html#Gtk.STOCK_DELETE


elelay:
Merged, thanks

PR #731, opened 2020-01-13
Remove executable bit for files that shouldn't be executable.

auouymous:
There are also three files in tools/ that are executable and three that are not. Should all or none of them be executable?


elelay:
all should be executable, I guess. And all should use `#!/usr/bin/env python3` instead of `/usr/bin/python3`

auouymous:
Done, except for getversion.py which didn't have a shebang line.

elelay:
Merged, thanks :-)

Issue #701, opened 2019-11-09
an add-on for automatically filter before download


alexis26:
hi,
some channel have a big content and mix their topics...
...it would be nice to have an add-on (an option) for automatically filter before downloading
like "only donwload if there is [example] in the title"

What do you think about ?
ty

elelay:
Yes, you can create such an add-on.
See https://gpodder.github.io/docs/extensions.html for explanations on how to write it.

You'll want to use the `on_podcast_updated` hook:
````
    def on_podcast_updated(self, podcast):
        ''' Called when a podcast is refreshed, but before showing the download dialog '''
        keyword = None
        # define your keyword per podcast
        if podcast.url == 'http://radiofrance-podcast.net/podcast09/rss_11370.xml':
            keyword = 'scandinave'
        # then if there is a filter, mark all non-matching episodes old
        if keyword:
            for e in podcast.get_episodes(gpodder.STATE_NORMAL):
                if e.is_new:
                    if keyword not in e.title:
                        e.mark_old()

````

**Comment by auouymous** [comment]
@elelay, Do you know of any extensions that modify the podcast settings dialog, or do you know how I could add a tab to that dialog? A filtering extension would be more useful if the user could edit filters in each podcast's settings dialog instead of hardcoding them in the extension's python file.

**Comment by auouymous** [comment]
The code says on_ui_object_available shouldn't be used but with the following patch, I was able to add a tab to the podcast's settings dialog. What I can't figure out though, is how to save changes to self.container.config.&lt;extension&gt;.* made by the extension. The changes show up in the edit config dialog but are lost when gpodder restarts.

```
diff --git a/src/gpodder/gtkui/desktop/channel.py b/src/gpodder/gtkui/desktop/channel.py
index bcdd8717..406184bd 100644
--- a/src/gpodder/gtkui/desktop/channel.py
+++ b/src/gpodder/gtkui/desktop/channel.py
@@ -94,6 +94,8 @@ class gPodderChannel(BuilderWidget):
         self.imgCoverEventBox.connect('button-press-event',
                 self.on_cover_popup_menu)
 
+        gpodder.user_extensions.on_ui_object_available('channel-gtk', self)
+
     def on_button_add_section_clicked(self, widget):
         text = self.show_text_edit_dialog(_('Add section'), _('New section:'),
             affirmative_text=Gtk.STOCK_ADD)
```

elelay:
[...]

PR #739, opened 2020-01-31
allow extensions to add tabs to preferences and channel settings dialogs

auouymous:
Needed by #701 for extensions to add additional channel settings. Adds two callbacks for extensions to add notebook tabs in preferences and channel settings dialogs. Also adds two on_ui_object_available signals ('preferences-gtk' and 'channel-gtk') for extensions to get raw access to the preferences and channel settings dialog windows.


elelay:
Just what I was wishing for!
Thanks

PR #741, opened 2020-02-01
show dialog when error updating feed

### LINKED ISSUE #594: Double exception when updating in 3.10.6
While checking for new episodes, it pauses for a minute on one podcast, outputs the following error and continues. There is code to display which podcast timed out but it appears to not be working. The podcast it paused on is probably a temporary issue but the url is https://feeds.pacific-content.com/commandlineheroes.

```
1547695146.205541 [gpodder.gtkui.main] ERROR: Error: <urlopen error timed out>
  File "/usr/lib/python3.6/urllib/request.py", line 1318, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "/usr/lib/python3.6/http/client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1234, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1026, in _send_output
    self.send(msg)
  File "/usr/lib/python3.6/http/client.py", line 964, in send
    self.connect()
  File "/usr/lib/python3.6/http/client.py", line 1392, in connect
    super().connect()
  File "/usr/lib/python3.6/http/client.py", line 936, in connect
    (self.host,self.port), self.timeout, self.source_address)
  File "/usr/lib/python3.6/socket.py", line 724, in create_connection
    raise err
  File "/usr/lib/python3.6/socket.py", line 713, in create_connection
    sock.connect(sa)
socket.timeout: timed out

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.6/site-packages/gpodder/gtkui/main.py", line 2624, in update_feed_cache_proc
    channel.update(max_episodes=self.config.max_episodes_per_feed)
  File "/usr/lib/python3.6/site-packages/gpodder/model.py", line 1021, in update
    result = self.feed_fetcher.fetch_channel(self)
  File "/usr/lib/python3.6/site-packages/gpodder/model.py", line 67, in fetch_channel
    return self.fetch(url, channel.http_etag, channel.http_last_modified)
  File "/usr/lib/python3.6/site-packages/gpodder/feedcore.py", line 234, in fetch
    return self._parse_feed(url, etag, modified)
  File "/usr/lib/python3.6/site-packages/gpodder/feedcore.py", line 190, in _parse_feed
    stream = util.urlopen(url, headers)
  File "/usr/lib/python3.6/site-packages/gpodder/util.py", line 1201, in urlopen
    return opener.open(request)
  File "/usr/lib/python3.6/urllib/request.py", line 526, in open
    response = self._open(req, data)
  File "/usr/lib/python3.6/urllib/request.py", line 544, in _open
    '_open', req)
  File "/usr/lib/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.6/urllib/request.py", line 1361, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/lib/python3.6/urllib/request.py", line 1320, in do_open
    raise URLError(err)
urllib.error.URLError: <urlopen error timed out>

```

### LINKED ISSUE #736: Showing Offline-Channels/Pages
With a certain amount of channels/pages in your subscriptions-list after a long period of time you might want to clean your list of those not existing anymore or have changed their URLs. When you check all channels for updates gpodder gives out a task-icon info about the channels that failed to update. This info does not show the channel-name and (afaik) does not go into some sort of log, it also disappears when you "mouse-over" it.
It would be handy to have the offline pages somehow marked in your list (on the sidebar inside gpodder gui) or a log with channel-name info about the offline/online status or at least to see the name of the ones that are offline somewhere. 
Right now keeping your subscriptions tidy means checking one after another if they're still working.
Maybe in some way this can be optimised. Something in this manner would be a good and user friendly feature in my opinion :) thanks!

auouymous:
This fixes #594 and partially fixes #736.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -1327,7 +1327,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
                 box.add(Gtk.HSeparator())
 
-                if len(channel.description) < 500:
+                if channel._update_error is not None:
+                    description = _('ERROR: %s') % channel._update_error
+                elif len(channel.description) < 500:
                     description = channel.description
                 else:
                     pos = channel.description.find('\n\n')
@@ -2610,6 +2612,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     self.pbFeedUpdate.set_text(progression)
 
                 try:
+                    channel._update_error = None
                     util.idle_add(indicate_updating_podcast, channel)
                     channel.update(max_episodes=self.config.max_episodes_per_feed)
                     self._update_cover(channel)
@@ -2617,8 +2620,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     d = {'title': html.escape(channel.title), 'url': html.escape(channel.url), 'message': html.escape(str(e))}
                     if d['message']:
                         message = _('Error while updating %(title)s at %(url)s: %(message)s')
+                        channel._update_error = str(e)
                     else:
                         message = _('The %(title)s feed at %(url)s could not be updated.')
+                        channel._update_error = '?'
                     self.notification(message % d, _('Error while updating feed'), widget=self.treeChannels)
                     logger.error('Error: %s', str(e), exc_info=True)
 
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2625,7 +2625,14 @@ class gPodder(BuilderWidget, dbus.service.Object):
                         message = _('The %(title)s feed at %(url)s could not be updated.')
                         channel._update_error = '?'
                     self.notification(message % d, _('Error while updating feed'), widget=self.treeChannels)
-                    logger.error('Error: %s', str(e), exc_info=True)
+                    logger.error('Error: %s', str(e), exc_info=(e.__class__ not in [
+                        gpodder.feedcore.BadRequest,
+                        gpodder.feedcore.AuthenticationRequired,
+                        gpodder.feedcore.Unsubscribe,
+                        gpodder.feedcore.NotFound,
+                        gpodder.feedcore.InternalServerError,
+                        gpodder.feedcore.UnknownStatusCode,
+                    ]))
 
                 updated_channels.append(channel)
 
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -704,20 +704,21 @@ class PodcastListModel(Gtk.ListStore):
         if self._cover_downloader is None:
             return pixbuf_overlay
 
-        if pixbuf_overlay is None:
+        if pixbuf_overlay is None:  # optimization: we can pass existing pixbuf
             pixbuf_overlay = self._get_cached_thumb(channel)
 
         if pixbuf_overlay is None:
+            # load cover if it's not in cache
             pixbuf = self._cover_downloader.get_cover(channel, avoid_downloading=True)
             pixbuf_overlay = self._resize_pixbuf(channel.url, pixbuf)
             self._save_cached_thumb(channel, pixbuf_overlay)
 
         if add_overlay:
-            if channel.pause_subscription:
+            if getattr(channel, '_update_error', None) is not None:
+                pixbuf_overlay = self._overlay_pixbuf(pixbuf_overlay, self.ICON_ERROR)
+            elif channel.pause_subscription:
                 pixbuf_overlay = self._overlay_pixbuf(pixbuf_overlay, self.ICON_DISABLED)
                 pixbuf_overlay.saturate_and_pixelate(pixbuf_overlay, 0.0, False)
-            elif getattr(channel, '_update_error', None) is not None:
-                pixbuf_overlay = self._overlay_pixbuf(pixbuf_overlay, self.ICON_ERROR)
 
         return pixbuf_overlay
 
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -497,6 +497,8 @@ class PodcastChannelProxy(object):
         self.cover_thumb = None
         self.auto_archive_episodes = False
 
+        self._update_error = None
+
     def get_statistics(self):
         # Get the total statistics for all channels from the database
         return self._db.get_podcast_statistics()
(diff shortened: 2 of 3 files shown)


elelay:
Hum. Not too sure: it happens that a flurry of podcast fail for some reason (network link down, etc.)
Wouldn't it spawn a lot of dialogs to click through?

auouymous:
Good point. What if it set a flag with reason in the channel on error and cleared it at start of update. The channel description could be replaced with "ERROR: <reason>" and bolded when flag is set.

auouymous:
Would be better if it could change the text to red but that would cause issues with different themes. Or if it could display an error icon, but I don't know how to do that.

auouymous:
I'll be adding a couple changes later today or tomorrow. A channel that failed to update could be hidden because it lacks episodes. Counting the errors and displaying the dialog will inform the user of how many channels failed and that unhiding deleting episodes might be required to see them. It should also not display a stack trace for common network exceptions.

auouymous:
I'm not sure what the dialog should say, but you can also search for "error:" in the channel pane without viewing all episodes.

src/gpodder/gtkui/main.py
```diff
@@ -2638,6 +2652,10 @@ def update_progress(channel):
 
                 util.idle_add(update_progress, channel)
 
+            if nr_update_errors > 0:
+                self.notification(_('%d channel(s) failed to update, view all episodes to see them') % nr_update_errors,
```
elelay: Maybe we can always show podasts in error? (I've not looked at the code).

auouymous: I can look into it, but there would be no way to hide a channel with temporary errors short of restarting gpodder or successfully updating the channel.

auouymous: Added a patch that always shows the channel and removed the message to view all.

auouymous:
Fixed the double html escaping of error messages.

elelay:
Nice!
Don't you have conflicts between the Error dialog and the "Select New episodes to download" dialog?
For me, both were showed, but the "Select New.." had to be closed before I could close the Error dialog.

elelay:
I've added an icon on the cover. What do you think?

This is also the first time I push commits on top of one of your PR. Is it OK for you or do you prefer that I open a different PR with my changes in the future?

src/gpodder/gtkui/model.py
```diff
@@ -539,18 +541,23 @@ def __init__(self, cover_downloader):
...
-            if model.get_value(iter, self.C_CHANNEL) == SectionMarker:
+            if channel == SectionMarker:
                 return True
             key = self._search_term.lower()
             columns = (model.get_value(iter, c) for c in self.SEARCH_COLUMNS)
             return any((key in c.lower() for c in columns if c is not None))
 
         if model.get_value(iter, self.C_SEPARATOR):
             return True
+        elif channel.__class__ == GPodcast and channel._update_error is not None:
```
elelay: Are you checking for the class here to discard the 'All Episodes from all podcasts' item?

auouymous: The proxy channel class (all episodes) has an update_errors field to prevent it from failing for everything that uses description. The class check there is to exclude the category labels when forcing channels with errors to always show.

auouymous:
I just noticed the pause icons on channels have been missing for quite some time. Displaying exception in _overlay_pixbuf() shows `gtk-icon-theme-error-quark: Icon 'gtk-media-pause' not present in theme Adwaita (0)`. Other code in gpodder uses 'media-playback-pause', should the channel overlay also use it?

Wouldn't 'dialog-error' be better than 'dialog-warning'?

Your patch duplicates the pixbuf_overlay branch when it could prepend to the existing branch.
```
+        if pixbuf_overlay is None:
+            pixbuf_overlay = self._get_cached_thumb(channel)
 
         if pixbuf_overlay is None:
```

It should also check for error before checking if paused, because the user might pause feeds and manually update them (I do).

Is the second patch related to this PR?

auouymous:
If the channel is paused and then manually updated and has errors, it overlays the error icon on the image that already contains the pause image. Not a problem when the pause image is smaller, but it can peek around the edges of the error icon if larger.

auouymous:
And yes, the selection window must be closed before the error dialog window can, not sure how to fix that.

src/gpodder/gtkui/model.py
elelay: ok, thanks for the clarification.

auouymous: I forgot to mention that it can be changed to use the same getattr as your icon patch, and avoid the class comparison.

```
+       elif getattr(channel, '_update_error', None) is not None:
```

elelay:
Thanks for the review!

> Other code in gpodder uses 'media-playback-pause', should the channel overlay also use it?

Yes

> Wouldn't 'dialog-error' be better than 'dialog-warning'?

I like the warning sign icon better.

> Your patch duplicates the pixbuf_overlay branch when it could prepend to the existing branch.
> 

I think `_get_cached_thumb(channel)` could return None. First check is for the optimization when we just created the pixbuf so have it at hand so there is no need to load it from cache. Second check is when the channel icon is not in cache.

> 
> It should also check for error before checking if paused, because the user might pause feeds and manually update them (I do).

Yes

> Is the second patch related to this PR?

No, I put it in master and forced push to remove it from the PR.

auouymous:
> > Wouldn't 'dialog-error' be better than 'dialog-warning'?
> 
> I like the warning sign icon better.

You prefer the icon your theme uses, or you would rather indicate the failure to update a channel as a warning instead of an error?

elelay:
> You prefer the icon your theme uses, or you would rather indicate the failure to update a channel as a warning instead of an error?

warning fits better: those errors are often transient. Plus there is a confusion with access denied with the error sign (in my theme at least). So a bit of both :-)

auouymous:
Should we get rid of the notification to prevent confusion with the dialog not being closable until the update window is closed?

elelay:
> Should we get rid of the notification to prevent confusion with the dialog not being closable until the update window is closed?

I agree. Maybe we could later modify this area of the UI to show a status with #errors, #new, etc.?

elelay:
So: this PR is read to merge, right?

elelay:
Merged, thanks! It's been long running but it's a nice UX improvement.

PR #743, opened 2020-02-04
extension to filter expisodes from being automatically downloaded

auouymous:
For #701.

### DIFF
--- a/share/gpodder/extensions/filter.py
+++ b/share/gpodder/extensions/filter.py
@@ -107,8 +107,8 @@ class gPodderExtension:
         #    self.channel               = PodcastChannel
         #    self.url                   = current filter url
         #    self.f                     = current filter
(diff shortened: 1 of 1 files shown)


auouymous:
@elelay I added your changes but removed an extra blank line to satisfy the linter.

I swapped the frames and renamed Allow to Except, and it looks much nicer. I left the variables as 'allow' because 'except' is a keyword, and I didn't rename allow_* settings because it would break filters for anyone using the current extension. I also sorted the filters after a change is made because it removes the filter and appends to end, which makes it harder to see what changed if Settings.json is tracked with version control.

elelay:
Great, thanks!

elelay:
Can you provide a manual page (new file in https://github.com/gpodder/gpodder.github.io/tree/master/docs/extensions)?
Then, we'll add a link to it in the extension/

PR #754, opened 2020-02-16
701 auouymous filter extension

elelay:
#743 refactoring


elelay:
@auouymous I've refactored the code to have a common class handling the 2 widgets (removes ~60 lines).
The emit_change method may not be super gtk/pythonic but it works.
What do you think?

elelay:
Closed in favor of #743

PR #763, opened 2020-03-13
show url when get_channel_desc() fails

auouymous:
Showing a (usually) cryptic youtube feed url isn't much but it is better than showing nothing.


elelay:
Merged, thanks

PR #765, opened 2020-03-22
Youtube dl improvements

elelay:
Following Ben's questions on the mailing list:
 - update Youtube-dl to latest version,
 - fix output filename generation,
 - catch blocked video errors video by video to keep non-blocked videos in the channel

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -290,7 +290,9 @@ class gPodderYoutubeDL(download.CustomDownloader):
 
     def fetch_video(self, url, tempname, reporthook):
         opts = {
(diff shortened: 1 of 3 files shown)


elelay:
@auouymous does it work for you?

auouymous:
I checked both boxes and it updated and downloaded without issues. I can't test if it handles errors since I don't often get any. The text shownotes didn't break from the change, but they weren't broken before so not sure how to test that.

elelay:
OK, thanks! The text shownotes have to be enabled from code if you have webkit2 gobject bindings installed (comment-out https://github.com/gpodder/gpodder/blob/master/src/gpodder/gtkui/shownotes.py#L42). Before the changed they showed the css as text, followed by the description.

PR #770, opened 2020-03-30
Update he.po

ShaiGerbi:
I translated some strings to Hebrew


elelay:
Merged, thanks!
If you would like to continue updating the Hebrew translation before releases, please subscribe to #518
I post there a message about a week before release if new strings appeared or were modified

PR #773, opened 2020-04-05
Fix #685 more discoverable search

### LINKED ISSUE #685: Discoverable way to search feeds
I'd like to be able to search the feeds I have subscribed to.

elelay:
To address #685.
Still UX in progress: feedback welcome :-)

- [x] add menu items to trigger podcast/episode search
- [x] link / doc search syntax

### DIFF
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -207,6 +207,7 @@
                             <property name="orientation">horizontal</property>
                             <child>
                               <object class="GtkEntry" id="entry_search_podcasts">
+                                <property name="hexpand">True</property>
                                 <property name="visible">True</property>
                                 <property name="secondary-icon-stock">gtk-close</property>
                                 <signal name="icon-press" handler="hide_podcast_search"/>
@@ -321,6 +322,7 @@
                             </child>
                             <child>
                               <object class="GtkEntry" id="entry_search_episodes">
+                                <property name="hexpand">True</property>
                                 <property name="visible">True</property>
                                 <property name="secondary-icon-stock">gtk-close</property>
                                 <signal name="icon-press" handler="hide_episode_search"/>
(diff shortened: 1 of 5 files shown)


auouymous:
Old and new search methods both work fine. I've only ever accidentally opened the search widgets and never intentionally tried to use them. But expanding them to full width is good.

The episode menu has episode actions at top and find at bottom. It makes sense for the podcast menu to have the same layout, with actions at top and find at bottom.

elelay:
> The episode menu has episode actions at top and find at bottom. It makes sense for the podcast menu to have the same layout, with actions at top and find at bottom.

Good point: I've moved it.

auouymous:
Works for me.

Is https://github.com/gpodder/gpodder/pull/773/commits/e315815d01381fd9c950ae6eda2ee7eb5c9422d8 related to this change?

https://github.com/gpodder/gpodder/pull/773/commits/298f81c3ea376407c8a803309c1c23ae3cc663df contains an extra space before the on_find_episode_activate function.

elelay:
> Is [e315815](https://github.com/gpodder/gpodder/commit/e315815d01381fd9c950ae6eda2ee7eb5c9422d8) related to this change?

Sort-of: I discovered the regression when documenting EQL search syntax.

 
> [298f81c](https://github.com/gpodder/gpodder/commit/298f81c3ea376407c8a803309c1c23ae3cc663df) contains an extra space before the on_find_episode_activate function.

Thanks! Fixed the linter warning...

auouymous:
I didn't know about this EQL until now, maybe there should be a clickable info/help icon on the search widget to make everyone aware of it.

But when I type "(" it writes an `unexpected EOF while parsing (<eql-string>, line 1)` error to console. And another if I type "()" followed by a backspace. It also throws that error for every character I type after "(" until I type the closing ")", if I type slow. And also if the query inside parens is not fully valid. That error should be commented out and only enabled when debugging.

Typing "(downloaded and minutes)" and then backspacing over the "s" in "minutes" throws a `'NoneType' object is not subscriptable` error for every episode that was matched before.

It also writes the names of all episodes to console, everytime the query changes.

auouymous:
The writing of episode names is due to a print in https://github.com/gpodder/gpodder/commit/e315815d01381fd9c950ae6eda2ee7eb5c9422d8.

auouymous:
The two `print(e)`'s in query.py are the cause of both error messages, first is the NoneType and second is the invalid syntax.

Issue #785, opened 2020-04-22
MKV Videos Aren't Downloaded


ChrisMTheMan:
GPodder never downloads MKV videos.

elelay:
Example feed please

ChrisMTheMan:
https://www.youtube.com/user/FreeDocumentary
https://www.youtube.com/channel/UCctuj_C9POmIHS4CkojVD1g
https://www.youtube.com/channel/UCBo9TLJiZ5HI5CXFsCxOhmA
https://www.youtube.com/channel/UCaJ_xrKTkBWMNbc8G-2SIwA

In all 4 of these cases, higher quality MKV files are bypassed in favor of lower quality MP4 videos.  I have a few more examples if you need them.  Thank you for taking a look.

auouymous:
Gpodder can't combine audio and video formats (adaptive formats) and 22 (720p) is the highest quality format youtube offers. They used to offer more and may still offer them on some old videos.

https://github.com/gpodder/gpodder/blob/master/src/gpodder/youtube.py#L42

That lists a few adaptive formats (the adaptive format for 22 is '136+140') that can be used by the youtube-dl extension when the primary formats aren't available. Running `youtube-dl -F <url>` from the command line will display a list of all formats, and the adaptive formats say "audio only" or "video only". Not all videos will have all formats, they only offer up to the quality provided by the creator. The formats for one of the videos you gave has the following formats.

```
249          webm       audio only tiny   57k , opus @ 50k (48000Hz), 8.80MiB
250          webm       audio only tiny   74k , opus @ 70k (48000Hz), 11.59MiB
140          m4a        audio only tiny  131k , m4a_dash container, mp4a.40.2@128k (44100Hz), 22.44MiB
251          webm       audio only tiny  145k , opus @160k (48000Hz), 22.79MiB
278          webm       256x144    144p   98k , webm container, vp9, 25fps, video only, 13.06MiB
160          mp4        256x144    144p  118k , avc1.4d400c, 25fps, video only, 7.68MiB
242          webm       426x240    240p  205k , vp9, 25fps, video only, 18.48MiB
133          mp4        426x240    240p  252k , avc1.4d4015, 25fps, video only, 16.01MiB
243          webm       640x360    360p  415k , vp9, 25fps, video only, 38.16MiB
134          mp4        640x360    360p  632k , avc1.4d401e, 25fps, video only, 31.02MiB
[...]

PR #787, opened 2020-04-24
refactor youtube formats and add settings for more resolutions

auouymous:
This refactors youtube formats making it easier to add additional formats for higher resolutions. It also adds all VP9 and H.264 formats, most requiring the youtube-dl extension (now indicated in the video setting). No obsolete non-DASH formats were removed, as old videos might have them, but newer videos won't.

Closes https://github.com/gpodder/gpodder/issues/785


elelay:
Merged, thanks!

PR #790, opened 2020-04-27
Gtk UI: Force-resize icons in player list

thp:
Ran into this issue on Ubuntu with the package `milkytracker` installed. The `.desktop` file contains `MimeType=audio/x-xm` (this is why it shows up in the list of players -- starts with `audio/`), and the only icon it installs is a 128x128 icon. This messes up the layout (even when the item is not selected).

Ideally we'd somehow blacklist items there or restrict items a little more (e.g. I also see Mp3splt-gtk, TiMidity++ and HandBrake in there, apps which probably nobody is going to use as "player" for podcast downloads), but not sure how to do that without blacklisting those items manually or potentially excluding valid choices.

Screenshot from 2020-04-27 18-23-46

### DIFF
--- a/src/gpodder/gtkui/desktopfile.py
+++ b/src/gpodder/gtkui/desktopfile.py
@@ -101,7 +101,7 @@ class UserApplication(object):
             (icon_name, extension) = os.path.splitext(os.path.basename(self.icon))
             theme = Gtk.IconTheme()
             if theme.has_icon(icon_name):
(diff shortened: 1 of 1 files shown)


elelay:
Good point about forcing icon size!

Regarding blacklisting, is it worth the hassle? Users will just pick what they want from the list.
The list can grow a bit, yes: 13 in my case, but is still manageable.

PR #791, opened 2020-04-27
Gtk UI: Add support for GtkHeaderBar / GNOME3 CSDs

thp:
It seems like on recent GNOME 3 versions, there's no more app menu (at least on the Ubuntu 19.10 I have here), it has been replaced with GtkHeaderBar + client-side decorations.

This creates a header bar and adds a refresh and menu button, similar to what other GNOME3 apps do.

### DIFF
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -218,6 +218,7 @@
                           <object class="GtkGrid" id="vbox42">
                             <property name="visible">True</property>
                             <property name="orientation">vertical</property>
+                            <property name="hexpand">True</property>
                             <child>
                               <object class="GtkButton" id="btnUpdateFeeds">
                                 <property name="label" translatable="yes">Check for new episodes</property>
--- a/src/gpodder/gtkui/app.py
+++ b/src/gpodder/gtkui/app.py
@@ -134,12 +134,17 @@ class gPodderApplication(Gtk.Application):
             # when the last one is closed the application shuts down
             self.window = gPodder(self, self.bus_name, core.Core(UIConfig, model_class=Model), self.options)
 
+            # If $XDG_CURRENT_DESKTOP is set then it contains a colon-separated list of strings.
+            # https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
+            # See https://askubuntu.com/a/227669 for a list of values in different environments
+            xdg_current_desktops = os.environ.get('XDG_CURRENT_DESKTOP', '').split(':')
+
             if gpodder.ui.osx:
                 from . import macosx
 
                 # Handle "subscribe to podcast" events from firefox
                 macosx.register_handlers(self.window)
-            else:
+            elif 'GNOME' in xdg_current_desktops:
                 # Use GtkHeaderBar for client-side decorations on recent GNOME 3 versions
                 self.header_bar_menu_button = Gtk.Button.new_from_icon_name('open-menu-symbolic', Gtk.IconSize.SMALL_TOOLBAR)
                 self.header_bar_menu_button.set_action_name('app.menu')
@@ -159,6 +164,7 @@ class gPodderApplication(Gtk.Application):
 
                 # Tweaks to the UI since we moved the refresh button into the header bar
                 self.window.btnUpdateFeeds.hide()
+                self.window.default_btn_update_feeds_visible = False
                 self.window.vboxChannelNavigator.set_row_spacing(0)
 
                 self.window.main_window.set_titlebar(self.header_bar)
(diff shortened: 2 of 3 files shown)


thp:
Here's the documentation reference from GNOME:
https://gitlab.gnome.org/GNOME/Initiatives/-/wikis/App-Menu-Retirement

auouymous:
Ugh! CSD is the worst! Could this use gnome detection?

thp:
> Ugh! CSD is the worst! Could this use gnome detection?

Does CSD not work (i.e. break/crash) on your machine? Or do you just happen to not like it very much? (in which case, I don't see the need for adding an option honestly)

If you plan to commit to maintaining, testing and supporting the non-GNOME code path, please feel free to make a pull request.

I am, however, open to making CSDs disabled on Windows if there is an issue with them there (but don't have a setup to test this at the moment). Please note that it's not used on macOS at the moment (since it has the global menu and probably works fine).

This is mostly about fixing the app menu being broken in recent Ubuntu/GNOME versions, not about adding more choices.

auouymous:
I didn't ask for an option, I asked for detection because Gnome/Ubuntu isn't the only DE/distro. It is easier to detect gnome/ubuntu and enable CSD than to detect every other platform/DE and disable it. There is no reason to have CSD when the user has a window manager that actually manages windows. So yes, it breaks usability and accessibility.

thp:
Which DE are you using where it breaks? Does the in-window main menu contain the "app menu" that has the "preferences", "about gpodder" and "quit" menu items?

auouymous:
I don't use a DE, I only use a window manager. Gpodder without this patch has a gPodder menu item on the normal menu bar. CSD removes all the features my window manager provides. Tiling window managers don't use a titlebar and have optional borders, CSD forces that on those users. Non-tiling window managers hide their own titlebar/borders and get CSD's fat titlebar and borders, which look nothing like the window manager's.

thp:
Updated PR to take `$XDG_CURRENT_DESKTOP` into account. It's a colon-separated list, e.g. on my Ubuntu it's `ubuntu:GNOME`. As long as `GNOME` doesn't appear in your list, it will default to the old UI.

auouymous:
Thank you, it works here.

elelay:
@thp I can't reproduce the issue on a full up to date ubuntu 19.10 : the menu is fully functional.

ubuntu_1910

thp:
> @thp I can't reproduce the issue on a full up to date ubuntu 19.10 : the menu is fully functional.
> 
> ubuntu_1910

Weird, must just be something with my local installation then (I did dist-upgrade from an older release and played around with selecting "pure" GNOME session, so maybe that affected things).

elelay:
No, you're just leaving on the edge: I'm running 3.34.2 but 3.36 is out.
This was not detrimental to the PR anyway, just an observation...

thp:
I also have 3.34.2, and just re-checked for master branch (without this PR):

- With the "GNOME" session, I do get a button in the titlebar that brings up the menu
- With the "Ubuntu" session, I don't get a button in the titlebar, but in your case there's clearly a "gPodder" menu item, so I assume something is wrong with my installation :)

In case of this branch, in the "Ubuntu" session I get the proper setting, but now in the GNOME session, I do get double menus :) Need to look into this...

elelay:
I'm with @auouymous regarding removing the menubar, but we shall go with what the DE recommends by default so your PR including detection is good.

Still a nagging doubt: what about older gnome versions?

I think we should provide an option for users wanting to keep a menu, like the existing 'Display > Show toolbar' or a setting that users can change from *edit settings*.

thp:
I created a new account on my machine and (with the same versions of GNOME) the "gPodder" menu was shown in the menu bar with the master branch. After some digging, I tracked this down to this `gsettings` property:

Broken:

    org.gnome.settings-daemon.plugins.xsettings overrides {'Gtk/ShellShowsAppMenu': <1>}

Fixed:

    org.gnome.settings-daemon.plugins.xsettings overrides @a{sv} {}

I "repaired" this in my user profile with:

    gsettings reset org.gnome.settings-daemon.plugins.xsettings overrides

So as far as I'm concerned, It's fine to leave it as-is (was just misconfiguration on my machine), but I've updated the PR as I've tested it yesterday with KDE Plasma (and the "vanilla" GNOME account) and cleaned it up a bit.

thp:
And for completeness sake, on KDE Plasma, client-side decorations DO look out-of-place, so in the end I'm convinced that CSDs should only be done for the environments in which they look "native" (that is, GNOME).

elelay:
It sure looks great:

csd

Do we need a settings for gnome users disliking it?

auouymous:
Disable it if the GTK_CSD environment variable is zero. `export GTK_CSD=0`

https://www.google.com/search?q=GTK_CSD

auouymous:
The URL in c2f94e8 says the default value for GTK_CSD is 1, @thp can you confirm this on your gnome? If that is the case, the gnome check could be removed, and only enable CSD when GTK_CSD=1.

thp:
`GTK_CSD` is not set in my environment.

elelay:
Merged, thanks!

PR #795, opened 2020-05-02
enable filter extension in cli (gpo)

auouymous:
There is no way to modify or show filters in gpo (yet), but this at least allows them to be used when updating feeds with gpo.

### DIFF
--- a/share/gpodder/extensions/filter.py
+++ b/share/gpodder/extensions/filter.py
@@ -4,15 +4,17 @@
 
 import re
 
-from gi.repository import Gtk
-
 import gpodder
+
+import gi  # isort:skip
+gi.require_version('Gtk', '3.0')  # isort:skip
+from gi.repository import Gtk  # isort:skip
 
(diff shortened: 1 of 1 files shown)


elelay:
Filters could be added by hand in the settings, given detailed documentation.
I've added a bit to it: https://gpodder.github.io/docs/extensions/filter.html
Feel free to enhance it or give more relevant examples :-)

I thought it was a good example, because there are full episodes and excerpts from them in the feed, which is most inconvenient because you don't want to download excerpts when you listen to full-length episodes. But the title doesn't give a clue: the duration in the description is a good indicator, but better is the file size (>40MB). Maybe extending the filter extension to use [EQL](https://github.com/gpodder/gpodder/blob/master/src/gpodder/query.py) would be cool?

elelay:
Merged, thanks!

auouymous:
Excellent idea! #804 and an EQL checkbox in filter extension would allow both EQL expressions with title and description matching.

PR #798, opened 2020-05-05
don't clear custom youtube formats

auouymous:
When a youtube format is selected from video preferences, it temporarily saves custom IDs. The custom IDs are lost if gpodder is restarted and might also be lost when preferences is closed. This patch uses a permanent ID value of zero to represent custom IDs and prevents losing them when testing. It might also be useful if someone wants to select a format that isn't in their custom IDs to download a single video, without worrying about losing the custom IDs.


elelay:
Merged, thanks!
I was hit by it too

PR #803, opened 2020-05-08
better indicate optional imports (gpod and eyed3.mp3)

### LINKED ISSUE #756: Use better warning messages and hide them when not needed
At every startup, my.py spams 'Flush requested, but sync disabled.' (it might also do it while running) and sync,py spams warnings about not finding gpod and eyed3. The import warnings need better messages to indicate they are optional and only impact feature X. Users with non-related issues often see those warnings and think fixing them will fix their real issue.

I know the warnings are needed for anyone who actually uses those features but it would be nice if loading my.py and sync.py could be disabled to make the warning spam go away when syncing is not used. I use a local patch that removes all of those warnings, but a proper fix would benefit all who don't use syncing.

### LINKED ISSUE #799: gpo error Could not find gpod
The Linux command line tool "gpo" produces the error "Could not find gpod", like this:
```
Storing data in /data/LT/gpo (GPODDER_HOME is set)
Storing downloads in /data/LT/Groups/Videos_non-English (GPODDER_DOWNLOAD_DIR is set)
Could not find gpod
```
I'm running Lubuntu 18.04, and the gpodder version that installed gpo is 3.10.15, which I got with "Install from git".

I have searched for code that might be producing this error, but haven't been able to find it.

Can you help, please.

auouymous:
Move optional import errors to time of use to avoid spamming all users with sync-only warnings. And better indicate why the modules are needed.

This fixes #799 and partially fixes #756 because it doesn't fix the 'Flush requested, but sync disabled.' warning.

I have this patch installed in my gpodder, but **I do not own any mp3/ipod devices to test this.**

### DIFF
--- a/src/gpodder/sync.py
+++ b/src/gpodder/sync.py
@@ -47,7 +47,6 @@ try:
     import gpod
 except:
     gpod_available = False
-    logger.warning('Could not find gpod')
 
 # pymtp_available = True
 # try:
@@ -56,10 +55,13 @@ except:
 #     pymtp_available = False
 #     logger.warning('Could not load gpopymtp (libmtp not installed?).')
 
+mplayer_available = True if util.find_command('mplayer') is not None else False
+
+eyed3mp3_available = True
 try:
     import eyed3.mp3
 except:
-    logger.warning('Could not find eyed3.mp3')
+    eyed3mp3_available = False
 
 
 if pymtp_available:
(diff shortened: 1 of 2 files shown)


src/gpodder/sync.py
```diff
@@ -130,14 +130,16 @@ def get_track_length(filename):
...
-        mp3file = eyed3.mp3.Mp3AudioFile(filename)
-        return int(mp3file.info.time_secs * 1000)
-    except Exception as e:
-        logger.warn('Could not determine length: %s', filename, exc_info=True)
+    if eyd3mp3_available:
+        try:
+            mp3file = eyed3.mp3.Mp3AudioFile(filename)
+            return int(mp3file.info.time_secs * 1000)
+        except:
+            pass
```
elelay, line 139: better `except Exception`.
See https://moserei.de/2017/09/20/python-bare-except.html

Also I'd rather see the error if there is one, at the expanse of duplicating the log.warn

src/gpodder/sync.py
```diff
@@ -130,14 +130,16 @@ def get_track_length(filename):
...
+
+    logger.warn('Could not determine length: %s', filename)
```
elelay, line 141: those 2 lines would be in an `else:` case

elelay:
Also, can't mutagen do the job of finding the audio file length from ID3, as a fallback for eyed3 not available (open question)?

src/gpodder/sync.py
auouymous: Which two lines?

src/gpodder/sync.py
auouymous: Fixed. I also fixed the mplayer case, or should its except not use Exception?

I also moved the two logger lines inside an else on the eyed3 branch, is that what you meant? One issue I see is mplayer failing, showing an error that it couldn't detect length and then eyed3 silently getting length. Should the eyed3 try use a finally with a warning that it detected the length, so the user knows?

```
finally:
    if ...mplayer...:
        logger.warn('eyed3.mp3 successfully determined length: %s', filename)
```

Of course these warnings would only be seen by gpo users and anyone looking in logs...

elelay: Perfect!

> Should the eyed3 try use a finally with a warning that it detected the length, so the user knows?

It would be nice, but maybe overkill. As you wish. Let me know if you want to leave it as is. I'll merge then.

auouymous: Added in second commit.

auouymous:
I don't know anything about mutagen, but it would require refactoring to support a third detector.

And another unrelated question, should the mplayer detection be moved to an mplayer_available variable to avoid checking for each sync'd file?

auouymous:
Removed the unneeded 'pass' from both exceptions and refactored so additional methods, such as mutagen, can easily be supported.

elelay:
perfect, thanks!

PR #804, opened 2020-05-08
add EQL functions to match text in title and/or description

auouymous:
> Normal EQL queries cannot be mixed with RegEx or string matching yet

This adds four functions S(), s(), R() and r() inside EQL. The comment in patch describes how they work. This could be useful for normal EQL usage, but will be needed for the filter extension to support EQL along with string and regexp matches.


elelay:
Great!

PR #805, opened 2020-05-08
shownotes: don't use blue hyperlinks on dark themes

elelay:
As seen when setting to the dark variant of themes, blue hyperlinks on black don't provide enough contrast.
This fixes it by checking the GSettings property and setting the color accordingly. It doesn't detect a theme with a dark background but not listed as a dark variant, but it was also the case before.

To check, add this line to your `~/.config/gtk-3.0/settings.ini`:
`gtk-application-prefer-dark-theme = true`

### DIFF
--- a/src/gpodder/gtkui/draw.py
+++ b/src/gpodder/gtkui/draw.py
@@ -459,7 +459,7 @@ def investigate_widget_colors(type_classes_and_widgets):
             # Create an empty style context
             style_ctx = Gtk.StyleContext()
             # Create an empty widget path
-            widget_path =  Gtk.WidgetPath()
+            widget_path = Gtk.WidgetPath()
             # Specify the widget class type you want to get colors from
             for t, c, r in type_and_class:
                 widget_path.append_type(t)
--- a/src/gpodder/gtkui/draw.py
+++ b/src/gpodder/gtkui/draw.py
@@ -405,3 +405,77 @@ def get_foreground_color(state=Gtk.StateFlags.NORMAL, widget=Gtk.TreeView()):
         color = style_context.get_color(state)
         p = p.get_parent()
     return color
+
+
+def investigate_widget_colors(type_classes_and_widgets):
+    """
+    investigate using Gtk.StyleContext to get widget style properties
+    I tried to compare gettings values from static and live widgets.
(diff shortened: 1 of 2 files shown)


auouymous:
I don't have webkit and can't test the html shownotes, but the about dialog uses gtk3's link_color and hover_link_color. Could shownotes use that as well?

elelay:
>  the about dialog uses gtk3's link_color and hover_link_color

sorry, I can't find any reference to link_color or hover_link_color. This would definitely be best.

auouymous:
Can't you do the shownotes link the same way the about dialog displays its two links?

elelay:
the about dialog uses pango markup for links (good color).
It's not good for displaying arbitrary text because it can't reflow paragraphs (insert line breaks).

Still you had a good point: trying to get the color from a Gtk.Label widget. It doesn't give me the hyperlink color, so now hyperlinks are underlined with the same color as the text. But it works in clear and dark themes.

I also reduced the font size for title and subtitle: they were much too big in hindsight.

elelay:
dark
light

auouymous:
https://stackoverflow.com/questions/38871450/how-can-i-get-the-default-colors-in-gtk

The second answer has code to print link color.

auouymous:
0e04659 will partially fix the progress icon. I have been using a patch that sets hard-coded bgc, fgc and txc in draw_cake() because it grabbed the wrong colors from theme. The following was most likely hard-coded due to those functions not returning the correct colors, and can now be changed to the unused fgc.

```
-        ctx.set_source_rgb(0.289, 0.5625, 0.84765625)                                                                                                                                                
+        ctx.set_source_rgb(fgc.red, fgc.green, fgc.blue)                                                                                                                                             
```

But the problem with using selected state is that the progress is harder to see when the episode is selected and its background is the same color as the progress icon. I don't know about others, but my theme uses a different color for Gtk.ProgressBar widgets that makes it easy to see when in a selected container widget. I tried to pass `widget=Gtk.ProgressBar` to get_xx_color() but its fg/bg colors are for text, not the actual bar.

auouymous:
Apparently text shownotes can display links and these patches have properly themed them, thanks!

elelay:
It's a deal, then :-)

Issue #818, opened 2020-06-07
new feature: select or search multiple Podcasts and combine their episodes in a list


rubina-ulula:
new feature: select or search multiple Podcasts and combine their episodes in a list

or allow to make podcast groups

elelay:
Podcasts can be combined in sections (create / assign sections in each podcast settings).
But it won't let you see just podcasts from this section.
I wonder why the podcast list is single selection only...

auouymous:
Maybe allow section name to be selected and display episodes for all podcasts in section.

As for searching, #820 makes it possible. The SsRr functions could also search in podcast and sections instead of just title and description, but that might be unwanted.

PR #820, opened 2020-06-09
add podcast and section EQL variables for use with SsRr functions

auouymous:
(s('...', podcast))
(s('...', section))

Allows searching based on podcast title or section name. Partially implements #818.


elelay:
merged, thanks!

PR #829, opened 2020-06-25
add geo_bypass and geo_bypass_country options for youtube-dl

### LINKED ISSUE #828: Add youtube-dl geo-bypass option in GPodder
This is a feature request for GPodder's youtube-dl. II'm running GPO 3.10.16 under windows 10 with youtube-dl extension enabled, letting it manage channels and downloads. I can see in the log files that lots of youtube videos can't be downloaded because of geo-restrictions. However youtube-dl itsself does provide a way to bypass these with the parameter "--geo-bypass". I'd really like to see this become available in GPodder as well. Sadly i'm not gifted in coding, but for what it's worth, thanks to all who support my wish and work to make it become true.

auouymous:
Closes #828

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -36,6 +36,11 @@ DefaultConfig = {
     # If for some reason youtube-dl download doesn't work for you, you can fallback to gpodder code.
     # Set to False to fall back to default gpodder code (less available formats).
     'manage_downloads': True,
+    # Bypass geographic restriction via faking X-Forwarded-For HTTP header
+    'geo_bypass': False,
+    # Two-letter ISO 3166-2 country code that will be used for explicit geographic
+    # restriction bypassing via faking X-Forwarded-For HTTP header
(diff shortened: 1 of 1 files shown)


share/gpodder/extensions/youtube-dl.py
```diff
@@ -270,6 +275,10 @@ def __init__(self, gpodder_config=None):
         if not sys.stdout:
             logger.debug('no stdout, setting YoutubeDL logger')
             self._ydl_opts['logger'] = logger
+        if ytdl_config.geo_bypass:
+            self._ydl_opts['geo_bypass'] = True
+        if ytdl_config.geo_bypass_country != '':
```
elelay, line 280:
```suggestion
        if ytdl_config.geo_bypass_country:
```

auouymous: I haven't fixed this yet because I'm not sure if this patch is the right way to handle it, or if it even works without the third setting. A per-podcast setting would be more useful than a global setting that might cause issues for other podcasts.

PR #839, opened 2020-07-06
Fix #796 MKV Files Not Properly Renamed

### LINKED ISSUE #796: MKV Files Not Properly Renamed
Using the following GPodder options, MKV files that are downloaded are not renamed properly.

1. Youtube-dl extension is in use with the "bestvideo+bestaudio" option.
2. Rename Episodes extension is in use.

If an MP4 or WEBM file is downloaded, the renaming works properly.  If an MKV file is downloaded it is not renamed and a zero-byte MP4 file is created that *is* properly named.

I don't have example video links available right now but will have them in a day or two.

elelay:
Try known youtube-dl extensions in case it changed it.
eg. when downloading mp4 video + opus audio it need to combine it in an mkv container.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -95,9 +95,6 @@ class YoutubeCustomDownload(download.CustomDownload):
         # youtube-dl doesn't return a content-type but an extension
         if 'ext' in res:
             dot_ext = '.{}'.format(res['ext'])
-            ext_filetype = mimetype_from_extension(dot_ext)
-            if ext_filetype:
-                headers['content-type'] = ext_filetype
             # See #673 when merging multiple formats, the extension is appended to the tempname
             # by YoutubeDL resulting in empty .partial file + .partial.mp4 exists
             # and #796 .mkv is chosen by ytdl sometimes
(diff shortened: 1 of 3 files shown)


elelay:
@auouymous it fixes the issue. Do you agree?

auouymous:
That fixes the issue with MKV files, but they get a WEBM or MP4 extension, depending on video format. Another issue is that the first download has an unknown ext and uses .bin.partial which gets renamed in download.py. If the file is deleted, format changed and episode is re-downloaded, the newly downloaded file keeps the previous extension.

Could retrieve_resume() return a new tempname and the filename in episode be updated with it? The youtube-dl extension could strip the extension from tempname, leaving .partial. If that tempname exists, it inserts dot_ext in returned_tempname and renames the file. Otherwise it uses the code in this patch to find the file and inserts try_ext in returned_tempname and renames the file. That way the files on disk have the correct extension.

Should MAX_FILENAME_WITH_EXT_LENGTH in model.py account for the extra extension youtube-dl can append after .partial? ".webm" would be the longest.

elelay:
Thanks for your comments!

> That fixes the issue with MKV files, but they get a WEBM or MP4 extension, depending on video format.
Another issue is that the first download has an unknown ext and uses .bin.partial which gets renamed in download.py. If the file is deleted, format changed and episode is re-downloaded, the newly downloaded file keeps the previous extension.

Yes: it's annoying.

> Could retrieve_resume() return a new tempname and the filename in episode be updated with it?

We'll have to make it work with autodetected resume at startup...

> Should MAX_FILENAME_WITH_EXT_LENGTH in model.py account for the extra extension youtube-dl can append after .partial? ".webm" would be the longest.

ah, yes, good point.

elelay:
bc5327d converts the episode extension to .mkv if necessary by returning the mimetype associated to the extension chosen by youtubeDL and ensuring it's used in `download.py`.
I think using the new mimetype in download.py was intented, but inadvertently broken at some point by refactoring `episode.extension()` and `episode.local_filename()`.

auouymous:
That works.

PR #848, opened 2020-07-18
fix youtube-dl not responding to config change

elelay:
now it checks if manage_downloads/manage_channels each time fetch_channel or custom_downloader is called.


auouymous:
LGTM, I'll let you know if it breaks anything.

PR #850, opened 2020-07-21
Discard episode when its GUID collides with a newer episode.

### LINKED ISSUE #847: Some podcast episodes aren't saved to database
On Windows 10, using gPodder 3.10.16, portable version.

After adding http://www.emptysea.net/tote/podcast1.xml to gPodder (first run, new database), all episodes (currently ~101) are correctly shown in the GUI. However, after restarting gPodder only 22 (in no particular order) are visible in the episodes list. Other feeds can be added and are displayed properly without issues.

The log lists the same error for each missing episode (only the object address changes):

`[gpodder.dbsqlite] ERROR: Cannot save <gpodder.gtkui.model.GEpisode object at 0x04736918>: UNIQUE constraint failed: episode.podcast_id, episode.guid`
`Traceback (most recent call last):`
`  File "C:\apps\gpodder-3.10.16-portable\data\lib\python3.8\site-packages\gpodder\dbsqlite.py", line 214, in _save_object`
`    cur.execute(sql, values)`
`sqlite3.IntegrityError: UNIQUE constraint failed: episode.podcast_id, episode.guid`

auouymous:
This prevents broken feeds from causing broken behavior in gpodder. Fixes #847.

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -149,10 +149,21 @@ class PodcastParserFeed(Feed):
         if self.max_episodes > 0 and len(entries) > self.max_episodes:
             entries = entries[:self.max_episodes]
 
+        num_duplicate_guids = 0
+
         # Search all entries for new episodes
         for entry in entries:
             episode = channel.EpisodeClass.from_podcastparser_entry(entry, channel)
             if episode is None:
+                continue
+
+            # Discard episode when its GUID collides with a newer episode
+            if episode.guid in seen_guids:
+                num_duplicate_guids += 1
+                channel._update_error = ('Discarded {} episode(s) with non-unique GUID, contact the podcast publisher to fix this issue.'
+                        .format(num_duplicate_guids))
+                logger.warn('Discarded episode with non-unique GUID, contact the podcast publisher to fix this issue. [%s] [%s]',
+                        channel.title, episode.title)
                 continue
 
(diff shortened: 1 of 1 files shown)


auouymous:
**Don't merge this yet.** I need to setup a test feed and verify it doesn't break anything if a new episode has the same GUID as an older downloaded episode.

auouymous:
The database exception only happens when multiple new episodes share a GUID, and do not share a GUID with an existing episode. All new episodes with the same GUID as an existing episode become updates for it, but the old episode details eventually overwrite the updates from newer episodes due to sorting newest to oldest. It basically discards all newer episodes and there was no way to get newer enclosures.

With this patch, the old episode keeps the updates from newer episodes, because the old episode is discarded. A downloaded file on the old episode is now "transfered" to the "new" episode. This has no impact if enclosure was used as GUID (both share the same enclosure). But if the new episode has a different enclosure, the episode must be deleted and re-downloaded to get the new enclosure. Of course no users will know this is possible, so this appears to discard old episodes but keeps their downloaded files.

The new patch behavior should be okay since it fixes the database error and other issues related to it, while informing the user about the discarded episodes.

elelay:
Seems correct to me. Ready to merge?

auouymous:
Yes.

elelay:
Merged, thanks!

PR #855, opened 2020-08-10
DORMANT is a power-saving mode of being online. 

youve:
Without this change, gPodder thinks I'm offline when I am really online, connected to wifi at home with iwd on a laptop running ArchLinux.

You can read about DORMANT mode here: https://askubuntu.com/questions/1194048/what-does-mean-dormant-mode-in-ip-link-show-wlan0


elelay:
Great, thanks!

PR #857, opened 2020-08-18
only send a single channel update error notification to avoid spamming the user

### LINKED ISSUE #856: Lots of "[Errno 11001] getaddrinfo failed" notifications
gPodder running on Windows. When updating feeds, it sometimes shows a lot of notifications:
> #### Error while updating feed
> Error while updating Some Podcast at https://example.com/podcast.rss: \&lt;urlopen error [Errno 11001] getaddrinfo failed\&gt;
> gPodder

Looks very similar to #825, although the Errno is slightly different.
The suggestion that it is a transient network failure seems reasonable. Sometimes it's just one or two podcasts, sometimes it's every single one. When it's every single one, you get, for example, 20 very similar notifications appearing sequentially (you have to dismiss each one before the next one is shown) which is rather inconvenient.
There might not be a lot gPodder can do to fix the transient failures themselves. However, you could consider two improvements:
1. Retry a few times (with a short delay between tries), before giving up and reporting the failure. Most of these failures seem to be very short-lived, so trying again will probably fix most of them.
2. When updating multiple podcasts, collect the errors and show a single notification when all updates have finished if any have failed.

There also appear to be a minor issue with the encoding of the text: The `<` and `>` have been html escaped, but the notification isn't rendering the escape codes.

auouymous:
Fixes #856

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2583,6 +2583,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         @util.run_in_background
         def update_feed_cache_proc():
             updated_channels = []
+            nr_update_errors = 0
             for updated, channel in enumerate(channels):
                 if self.feed_cache_update_cancelled:
                     break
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!

PR #866, opened 2020-09-11
Show update error in the podcast dialog (for copy-paste and seeing it fully)

elelay:


elelay:
@auouymous this adds to the error visible in the popup but not copiable.

auouymous:
Works for me.

elelay:
ok, cool!

PR #867, opened 2020-09-12
Youtube live streaming support

auouymous:
Allows live streams to be streamed (but not downloaded) while live. The video player must support HLS m3u8 URLs (ffplay does). It adds a new `Video` setting to select HLS formats, and supports custom formats in `Edit config`.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -295,7 +295,7 @@ class gPodderYoutubeDL(download.CustomDownloader):
         #
         # See https://github.com/ytdl-org/youtube-dl#format-selection for details
         # about youtube-dl format specification.
-        fmt_ids = youtube.get_fmt_ids(gpodder_config.youtube)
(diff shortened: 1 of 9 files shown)


elelay:
Looks good.
I've added a fix to allow streaming videos if a custom video player was selected but the default audio player.
Do you agree?

auouymous:
I don't understand what you said.

elelay:
gPodder enables the **Stream** menu item only if the user has selected a player (vlc, ...).
episodes are streamed by executing the selected player with the download url.
audio episodes execute the audio player, video episodes the video player.

Before 8503fef the **Stream** menu item was disabled for video episodes (youtube live video) even if I selected VLC as video player because I kept the default audio player. Now gPodder looks at the video player for video episodes and the audio player for other kinds.

auouymous:
Ah, I had a non-default audio player and the stream option does dim when changed to default. LGTM.

elelay:
Let's merge, then!
Thanks,

PR #868, opened 2020-09-12
Add menu item to undownload downloaded episodes.

### LINKED ISSUE #864: Feature Request: add redownload dropdown menu option

auouymous:
Closes #864.

### DIFF
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -125,6 +125,10 @@
           <attribute name="label" translatable="yes">Cancel</attribute>
         </item>
         <item>
+          <attribute name="action">win.undownload</attribute>
+          <attribute name="label" translatable="yes">Undownload</attribute>
+        </item>
+        <item>
           <attribute name="action">win.delete</attribute>
           <attribute name="label" translatable="yes">Delete</attribute>
         </item>
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2995,8 +2995,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     channel_urls.add(episode.channel.url)
                     episodes_status_update.append(episode)
                     if undownload:
-                        # Mark episode as new
-                        episode.mark(is_played=episode.is_new)
+                        # Undelete and mark episode as new
+                        episode.state = gpodder.STATE_NORMAL
+                        episode.is_new = True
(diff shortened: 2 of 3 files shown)


elelay:
My concern is that a user will be puzzled by the choice between `Delete` and `Undownload`.
Maybe a better label? **Delete but/and keep New**?
Another option would be a popup to ask if marking new after deletion, but it would be annoying in the majority of cases (deleted because it's played or the user doesn't want to play it)...

elelay:
Actually there is already a confirmation popup. Add a checkbox to **mark new to redownload*?

auouymous:
Undownload is like an undo for download, the first four letters are even "Undo" and it uses the edit-undo icon. `Delete but keep new` doesn't make sense if an old/new checkbox appears on confirmation or if is_new won't be changed (see below). So it would need to be something like `Delete but keep episode`, which doesn't quite make sense as to what is happening. `Delete file, keep episode` might make more sense?

I changed the title and message for show_confirmation() when undownloading versus deleting, but I don't see how to add a checkbox to that dialog. Maybe it would be better to never mark as new, just change state to normal and leave it in its current is_new state. I originally marked it as new because I was using the mark(is_played) method to undelete it. The second commit makes it easy to remove the is_new line and have it still function properly. When redownloading a corrupt download, the episode will be marked new after it gets redowloaded, it wouldn't matter if it was new or old. When waiting to redownload for a better format, the user can manually mark it new for automatic download at next update or leave as old and manually redownload after a few hours.

auouymous:
Rebased. The episode must be marked as new otherwise the feature is worthless for anyone who uses `View..Downloaded episodes` or `View..Unplayed episodes`, although the later requires an extra step if the episode is played because it gets removed and the view must be changed to undownload it. Anyone wanting the episode to be marked as played can perform the extra step after undownloading. But I imagine most would either immediately download it, or want it unplayed so the next update automatically downloads it.

> Undownload is like an undo for download, the first four letters are even "Undo" and it uses the edit-undo icon. Delete but keep new doesn't make sense if an old/new checkbox appears on confirmation or if is_new won't be changed (see below). So it would need to be something like Delete but keep episode, which doesn't quite make sense as to what is happening. Delete file, keep episode might make more sense?

Just waiting for a response to that.

elelay:
I still think **undownload** is not going to help users and would would very much like to add a checkbox to the confirm dialog. This is what would feel natural to me.

auouymous:
Replaced by #1284.

PR #869, opened 2020-09-12
fix NEW_LOCATION with same url, resulting in Already subscribed message

elelay:
when adding a new podcast.
Seen on https://www.youtube.com/watch?v=f9qOPA05fzE

### DIFF
--- a/src/gpodder/feedcore.py
+++ b/src/gpodder/feedcore.py
@@ -210,17 +210,17 @@ class Fetcher(object):
             ad = FeedAutodiscovery(url)
 
             ad.feed(data.getvalue())
-            if ad._resolved_url:
+            if ad._resolved_url and ad._resolved_url != url:
                 try:
                     self._parse_feed(ad._resolved_url, None, None, False)
                     return Result(NEW_LOCATION, ad._resolved_url)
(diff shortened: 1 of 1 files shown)


auouymous:
Does this work for you? It replaced the already subscribed error with another.

```
gpodder.feedcore.InvalidFeed: Could not parse feed: <unknown>:1:0: syntax error
```

elelay:
> Does this work for you? It replaced the already subscribed error with another.
> 
> ```
> gpodder.feedcore.InvalidFeed: Could not parse feed: <unknown>:1:0: syntax error
> ```

now it's *the right* error ;-)

Subscribing to a single youtube video won't work: no autodiscovery and not a feed. The error already happened, but was hidden by the nonsensical "already subscribed".

auouymous:
In that case, it LGTM. :)

The `<link itemprop="url" href="http://www.youtube.com/channel/UCiAInBL9kUzz1XRxk66v-gw">` tag was found in that video page and could be detected in feedcore.py to allow subscribing via a video page. I can do it when I get time, unless you do it sooner.

elelay:
Please do it when you have time

PR #879, opened 2020-09-26
Avoid stale Today and Yesterday dates by refreshing episodes daily.

auouymous:
The dates are only updated when the view is changed, gpodder is restarted, and perhaps after an update. This can cause episodes to be marked with Today or Yesterday when they are several days old. This patch automatically refreshes the dates every day at midnight.


elelay:
Cool idea! Thanks

PR #893, opened 2020-11-06
Update de.po

TZocker:
https://github.com/gpodder/gpodder/issues/518#issuecomment-723172034


elelay:
Merged, thanks.
I'll see what I can do for the plural...

PR #897, opened 2020-11-09
Do not show new episodes on view->Downloaded

### LINKED ISSUE #468: Add option to hide New Episodes
Hi,
I"m using gpodder 3.9 on Win10 64-bit. The link below shows that I see a list of all podcasts even when I try to filter to downloaded episodes only.
image
this is a critically important feature. Any fixes or suggestions?
Thanks.

tpikonen:
Add 'Always show New episodes' toggle to View menu

Fixes #468.

Adds a new boolean config item ui.gtk.episode_list.always_show_new, which can be toggled from menu with View->Always show New episodes.

If set, the Downloaded episodes view also shows new episodes.

### DIFF
--- a/src/gpodder/config.py
+++ b/src/gpodder/config.py
@@ -162,6 +162,7 @@ defaults = {
                 'descriptions': True,
                 'view_mode': 1,
                 'columns': int('110', 2),  # bitfield of visible columns
+                'always_show_new': True,
             },
 
             'download_list': {
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -390,7 +390,7 @@ class EpisodeListModel(Gtk.ListStore):
             elif episode.state == gpodder.STATE_NORMAL and \
                     episode.is_new:
                 tooltip.append(_('New episode'))
-                view_show_downloaded = True
+                view_show_downloaded = self._config.ui.gtk.episode_list.always_show_new
                 view_show_unplayed = True
             elif episode.state == gpodder.STATE_DOWNLOADED:
                 tooltip = []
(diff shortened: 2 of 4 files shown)


elelay:
Thanks for the patch. See https://github.com/gpodder/gpodder/issues/468#issuecomment-510179540 for proposal to add a checkbox "always show new episodes" in combination with this.

tpikonen:
Ok, I implemented the checkbox. It works for me but please test.

src/gpodder/gtkui/model.py
```diff
@@ -390,7 +390,8 @@ def get_update_fields(self, episode, include_description):
             elif episode.state == gpodder.STATE_NORMAL and \
                     episode.is_new:
                 tooltip.append(_('New episode'))
-                view_show_downloaded = True
+                if self._config.ui.gtk.episode_list.always_show_new:
+                    view_show_downloaded = True
```
elelay, line 394: Replace by `view_show_downloaded = self._config.ui.gtk.episode_list.always_show_new`

elelay (changes requested):
I haven't tested it yet but it looks fine!
Just change the one remark.

I want to push a new release this week and already polled translators for updates so I won't merge your PR just now.
It will be merged next week, though.

Thanks,

src/gpodder/gtkui/model.py
tpikonen: Good catch, I've pushed a fixed version.

src/gpodder/config.py
```diff
@@ -162,6 +162,7 @@
                 'descriptions': True,
                 'view_mode': 1,
                 'columns': int('110', 2),  # bitfield of visible columns
+                'always_show_new': False,
```
elelay, line 165: can you make it True by default?

elelay (approved):
perfect, thanks.
Just make it checked by default for backward compatibility.

src/gpodder/config.py
tpikonen: Done, and pushed.

elelay:
Great, thanks!
Will be merge as is

elelay:
Merged, thanks

PR #901, opened 2020-11-12
Fix translation of Youtube playlist URL to a feed

tpikonen:
There's a bug when converting a youtube playlist URL to a feed, which causes the feed URL to be malformed and thus not found. This patch fixes the URL conversion for youtube playlists.

### DIFF
--- a/src/gpodder/youtube.py
+++ b/src/gpodder/youtube.py
@@ -485,7 +485,7 @@ def parse_youtube_url(url):
 
         if 'list=' in query:
             playlist_query = [query_value for query_value in query.split("&") if 'list=' in query_value][0]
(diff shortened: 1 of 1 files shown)


elelay:
Hi,
thanks for the patch. Is it with all playlists ? Do you have an example ?

tpikonen:
The original code is obviously wrong, the strip() method removes all characters in the set 'list=' from the beginning or end of playlist_id. YT playlist_ids seem to start with a 'P', but any id ending with one of these letters will be corrupted.

elelay:
You're right: it couldn't work!

elelay:
Merged, thanks!

PR #902, opened 2020-11-14
Prevent an empty entries bug in YouTube-DL from removing all episodes.

### LINKED ISSUE #888: Gpodder Downloading All youtube Episodes Every Time
About a week ago Gpodder suddenly began trying to download every episode of every subscription every time a download cycle starts.  I am using Gpodder 3.10.16.  I have not recently updated Gpodder nor done any Windows updates.

auouymous:
Fixes #888.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -151,7 +151,7 @@ class YoutubeFeed(model.Feed):
         filtered_entries = []
         seen_guids = set()
         for i, e in enumerate(entries):  # consumes the generator!
-            if e.get('_type', 'video') == 'url' and e.get('ie_key') == 'Youtube':
+            if e.get('_type', 'video') in ('url', 'url_transparent') and e.get('ie_key') == 'Youtube':
                 guid = video_guid(e['id'])
                 e['guid'] = guid
                 if guid in seen_guids:
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -41,6 +41,7 @@ DefaultConfig = {
 
 # youtube feed still preprocessed by youtube.py (compat)
 CHANNEL_RE = re.compile(r'''https://www.youtube.com/feeds/videos.xml\?channel_id=(.+)''')
+USER_RE = re.compile(r'''https://www.youtube.com/feeds/videos.xml\?user=(.+)''')
 PLAYLIST_RE = re.compile(r'''https://www.youtube.com/feeds/videos.xml\?playlist_id=(.+)''')
 
 
@@ -395,9 +396,13 @@ class gPodderYoutubeDL(download.CustomDownloader):
         if m:
             url = 'https://www.youtube.com/channel/{}/videos'.format(m.group(1))
         else:
-            m = PLAYLIST_RE.match(channel.url)
+            m = USER_RE.match(channel.url)
             if m:
-                url = 'https://www.youtube.com/playlist?list={}'.format(m.group(1))
+                url = 'https://www.youtube.com/user/{}/videos'.format(m.group(1))
+            else:
+                m = PLAYLIST_RE.match(channel.url)
+                if m:
+                    url = 'https://www.youtube.com/playlist?list={}'.format(m.group(1))
         if url:
             logger.info('Youtube-dl Handling %s => %s', channel.url, url)
             return self.refresh(url, channel.url, max_episodes)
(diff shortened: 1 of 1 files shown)


elelay:
thanks for looking into it!
Rather than returning None, it should return `feedcore.Result(feedcore.NOT_MODIFIED, the_url)`. This is already handled by model.py.
Is it for all channels? Also with youtube-dl 2020.11.1.1?

auouymous:
I forgot to mention the `entries` length isn't known in refresh() due to it being a generator, and therefore it couldn't return `feedcore.NOT_MODIFIED`.

I should also add logging and set channel update error because a first time subscriber won't see any episodes for a channel this happens on. The issue could be with channels that haven't had a new episode in some time, or it might just be random. This could of course produce a lot of channels with update errors, but if random, a user would like to know why gpodder isn't see an episode they know exists.

I had found a reddit post saying youtube updated the page layout a couple months ago and youtube-dl hadn't fully handled it yet. And yes, this is with 2020.11.1.

One issue with this fix is that if a channel deletes all episodes, gpodder won't also remove them, and they will fail when a download is attempted.

elelay:
OK it's a generator indeed...
Waiting for your improvements :-)

auouymous:
I thought 2020.11.12 had fixed the issue because my test script was working fine with `process=True` (returns a list instead of a generator). 2020.11.1 would return an empty list or empty generator depending on `process`, however, 2020.11.12 returns the correct list but the generator only contains playlist urls. Appending '/videos' to the url returns a generator with all upload videos, as expected.

But the generator is always empty in gpodder, with or without '/videos', even though the same options are used. Are you able to see anything gpodder does differently than this script?

```
#!/bin/env python3
import youtube_dl

url = 'https://www.youtube.com/user/Lamborghini'
#url = 'https://www.youtube.com/user/Lamborghini/videos'

process = True # list
#process = False # generator

opts = { 'youtube_include_dash_manifest': False, 'no_cachedir': True, 'no_color': True }
result = youtube_dl.YoutubeDL(opts).extract_info(url, download=False, process=process)
print(result)
if not process:
    entries = [e for i, e in enumerate(result['entries'])]
    print('')
    for e in entries:
        print(e)
```

Also, `USER_RE = re.compile(r'''https://www.youtube.com/feeds/videos.xml\?user=(.+)''')` allows manage_channels to work with user feeds, and I can submit a patch once this is fixed but doing so right now would only amplify the current issue.

elelay:
> Are you able to see anything gpodder does differently than this script?

gPodder youtube_dl extension does this:
```
opts = {
            'youtube_include_dash_manifest': False,  # only interested in video title and id
            'cachedir': cachedir,
            'no_color': True,  # prevent escape codes in desktop notifications on errors
            'quiet': True,
        }
with youtube_dl.YoutubeDL(opts) as ydl:
            ie_result = ydl.extract_info(url, download=False, process=False)
```

elelay:
:tada: There is a new version 2020.11.17 :tada: 

regarding your test script, it seems using /user/Lamborghini with process=False returns a playlist selection, as seen on the front page, /user/Lamborghini/videos returns all videos, as seen on the videos tab. So we should add /videos to the url if it's /user/xxx ?

auouymous:
I have a patch in gpodder that replaces cachedir with no_cachedir, so the script is the same as my gpodder.

We will probably need to append '/videos' but first we need to fix the issue, which I just found. _process_entries() checks for _type == 'url' but everything is 'url_transparent'.

```
- if e.get('_type', 'video') == 'url' and e.get('ie_key') == 'Youtube':
+ if e.get('_type', 'video') in ('url', 'url_transparent') and e.get('ie_key') == 'Youtube':
```

I don't know if it is correct, but it seems to work. Should I replace the patches with it?

elelay:
> I don't know if it is correct, but it seems to work. Should I replace the patches with it?

yes, by all means :-). I think I was being conservative and only handled url.

auouymous:
It was painful waiting forever for my feeds to update and then clean out the extra "old" episodes per channel, but it worked. The last patch might cause some to think the update process got a lot slower if they have a lot of user feeds. But in reality, it is now managing *all* of their youtube feeds instead of just *some*, you might want to mention that in the changelog.

Unrelated, is there a reason https://github.com/gpodder/gpodder/blob/master/share/gpodder/extensions/youtube-dl.py#L196 limits the list to max_episodes when the generator was already limited at https://github.com/gpodder/gpodder/blob/master/share/gpodder/extensions/youtube-dl.py#L164 ?

elelay:
Great! So it's ready to merge, then.

> Unrelated, is there a reason https://github.com/gpodder/gpodder/blob/master/share/gpodder/extensions/youtube-dl.py#L196 limits the list to max_episodes when the generator was already limited at https://github.com/gpodder/gpodder/blob/master/share/gpodder/extensions/youtube-dl.py#L164 ?

No reason: maybe I wrote them at different times. line 164 is enough.

Thanks,

PR #913, opened 2020-11-27
Fix certifi certs path in case .local/lib/python38 exists in user home directory in win32 installer build script

### LINKED ISSUE #912: Windows Installer Build Script: installer creation fails when bootstrap.sh has been run before
https://github.com/gpodder/gpodder/blob/72dcda9e56f4573e4887cb4510958efa84d42651/tools/win_installer/bootstrap.sh#L27-L29

This creates `C:\Users\<username>\.local\lib\python3.8\site-packages\...`. The path then appears in `sys.path` in python right before `C:/msys64/mingw32/lib/python3.8/site-packages`.

Then in `_base.sh`:

https://github.com/gpodder/gpodder/blob/72dcda9e56f4573e4887cb4510958efa84d42651/tools/win_installer/_base.sh#L122-L123

`${site_packages}` becomes `C:/Users/<username>/.local/lib/python3.8/site-packages/...` (which should be `C:/msys64/mingw32/lib/python3.8/site-packages`), so `./build.sh` ends up with error:

```
cp: cannot stat 'C:/Users/<username>/.local/lib/python3.8/site-packages/certifi/cacert.pem': No such file or directory
```

shunf4:
This fixes #912.

### DIFF
--- a/tools/win_installer/_base.sh
+++ b/tools/win_installer/_base.sh
@@ -119,7 +119,7 @@ function install_deps {
     # replace ca-certificates with certifi's
     build_pacman --noconfirm -Rdds mingw-w64-"${ARCH}"-ca-certificates || true
     mkdir -p ${MINGW_ROOT}/ssl
-    site_packages=$(build_python -c  'import sys;print(next(c for c in sys.path if "site-packages" in c))')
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!

PR #915, opened 2020-12-01
fix: win installer build: more accurate site_packages path 2

shunf4:
Another place in https://github.com/gpodder/gpodder/issues/912

https://github.com/gpodder/gpodder/blob/72dcda9e56f4573e4887cb4510958efa84d42651/tools/win_installer/_base.sh#L159


elelay:
Merged, thanks :-)

PR #919, opened 2020-12-09
Log youtube-dl version and warn if known to have issues.

auouymous:
The logging might help with debugging youtube-dl issues, and the warning can be used to flag known broken versions.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -29,6 +29,9 @@ __only_for__ = 'gtk, cli'
 __authors__ = 'Eric Le Lay <elelay.fr:contact>'
 __doc__ = 'https://gpodder.github.io/docs/extensions/youtubedl.html'
 
+want_ytdl_version = '2020.11.12'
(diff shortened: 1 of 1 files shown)


computerfreak105:
Thanks, good idea.
But how can one update the youtube-dl which is used by GPO under windows? It’s something I really would like to know.
When I use youtube-dl stand alone I do it with „youtube-dl -U“. But I haven’t figured out where this program is actually located in GPO and how to make sure GPO is using the latest version.

I appreciate your work on GPO.
Thanks
Ben

auouymous:
The warning would only trigger for issues known at the time of a gpodder release. When 3.10.17 was released, it was known that youtube-dl versions before 2020.11.12 had issues. Even though 3.10.17 fixed part of the issue, an updated youtube-dl would be required to properly work.

I assume the windows version includes the version of youtube-dl available at time of release, and this warning would never be shown for someone using a bundled version. If this is true, you can update the bundled youtube-dl by grabbing the newest git version of gpodder at https://ci.appveyor.com/project/elelay/gpodder/build/artifacts.

elelay:
Merged, thanks

PR #927, opened 2020-12-20
Don't send kernel version in user-agent.

### LINKED ISSUE #926: gPodder user-agent leaks a lot of information about my system
The podcast feeds gets fetched with an user-agent that leaks the distro and kernel version of the linux machine I'm using. It seems like at most platform.system() should be sent with it. not platform.release()

For example: "Linux/5.9.11-3-MANJARO" and "Linux/5.9.14-arch1-1"

auouymous:
Closes #926.

### DIFF
--- a/src/gpodder/__init__.py
+++ b/src/gpodder/__init__.py
@@ -89,7 +89,7 @@ quiet = False
 
 
 # The User-Agent string for downloads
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks

Issue #929, opened 2020-12-22
Add podcast dialog expands text entry when dialog is resized


Johennes:
Discovered in the course of #878: On mobile Linux distros such as postmarketOS or Mobian `GtkDialog`s are automatically stretched to cover the screen. The add-podcast dialog currently expands the hbox with the text entry which looks a bit odd (and is also a bit superfluous because it still only shows a single line of input).

![before](https://user-images.githubusercontent.com/1137962/102929793-89f3ad80-449b-11eb-894e-2b9989979c49.png)

Using `<property name="expand">False</property>` in `gpodderaddpodcast.ui` makes it look like this.

![after](https://user-images.githubusercontent.com/1137962/102930039-02f30500-449c-11eb-812f-104ca2360cca.png)

The large amount of whitespace might still look a bit odd but the text entry and paste button seem more intuitive and usable to me this way.

I'm happy to submit a tiny pull request to fix this, if that would be appreciated.

auouymous:
A PR or patch is the fastest way to get an issue resolved.

Johennes:
:+1: 

Just meant to test the waters first because I have never contributed to gpodder before. Some projects don't accept PRs without a preceding issue and discussion.

PR opened: #930

elelay:
medium-to-big changes should be first discussed via an issue or when it's changing the UX,
just to avoid the rejected PR after the submitter spent time to implement their idea.
But minor adjustments are welcome as a direct PR. The screenshot is a very nice touch, though, since it makes us aware of the problem.

Johennes:
Makes sense, will keep that in mind for the future. :+1:

PR #930, opened 2020-12-23
Prevent text entry from expanding

Johennes:
On mobile screens GtkDialog windows are stretched to cover all of the
screen. This prevents the hbox with the text entry from expanding when
the window grows (which looks rather unintuitive).

Fixes: #929


auouymous:
LGTM, and it doesn't cause issues on the desktop.

elelay:
Merged, thanks

PR #931, opened 2020-12-23
Refactor shownotes

tpikonen:
Currently the header part of the HTML shownotes is attached to the top of the pane and not scrolled with the rest of the description. This wastes space and is annoying on small screens and short panes.

This patch makes the full HTML shownote pane scrollable by converting the header content which was in a separate GtkTextView widget to HTML and using a single WebView. It also unifies TextView and HTML shownote versions to use a GtkOverlay for showing link destinations. The title section also shows more episode details and uses non-breakable spaces between values and units etc.

### DIFF
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -18,6 +18,7 @@
 #
 import html
 import logging
+import re
 from urllib.parse import urlparse
 
 import gpodder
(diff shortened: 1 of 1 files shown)


auouymous:
I can't test with HTML shownotes but it works fine with text shownotes. The header always scrolled in text shownotes and HTML shownotes should do the same, so I agree with this PR.

The new information should probably be in a separate PR since it has nothing to do with scrolling the header. I don't like the period after the channel name and size. Why was `Released` capitalized when `from`, `duration` and `size` aren't? The PR's primary purpose is to scroll the header to save space but then a new line is added that reduces space by adding information that is probably already visible in the episode view.

tpikonen:
I removed the extra episode details from shownotes. I'll make a separate PR for them if or when this one gets merged, as suggested.

auouymous:
Thanks, text shownotes still work fine with the patch.

elelay:
Thanks: looks good while playing with the UI. I'll review the code later :-)

elelay:
Merged, thanks!
Minor tweaks: removed space between title and subtitle and restored translations...

PR #933, opened 2020-12-25
Get correct window size when saving UI state

tpikonen:
The main gPodder window is growing after restart. This patch should fix it. Only tested on Gnome, so please check if this breaks on other environments.

### DIFF
--- a/src/gpodder/gtkui/config.py
+++ b/src/gpodder/gtkui/config.py
@@ -158,10 +158,7 @@ class UIConfig(config.Config):
 
         if cfg.width != -1 and cfg.height != -1:
             window.resize(cfg.width, cfg.height)
-        if cfg.x == -1 or cfg.y == -1:
-            window.set_position(Gtk.WindowPosition.CENTER_ON_PARENT)
-        else:
-            window.move(cfg.x, cfg.y)
+        window.set_position(Gtk.WindowPosition.CENTER_ON_PARENT)
 
         # Ignore events while we're connecting to the window
         self.__ignore_window_events = True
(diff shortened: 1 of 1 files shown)


auouymous:
`widget.get_size()` returns the same values as the event for me, so it seems to work.

I have been using a patch to prevent window walking because the event position includes window decorations. But `widget.get_postion()` returns the correct values, can you confirm this?

```
-            x_pos, y_pos = event.x, event.y
+            x_pos, y_pos = widget.get_postion()
```

But https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-get-position says the application shouldn't save and restore window position, so maybe gpodder's position code should be removed. My window manager remembers the size and location and automatically restores the window, but then gpodder does its own thing and walks away. @elelay Do you know anything about this?

elelay:
Using the i3 tiling manager I don't see the restored size.
OK to drop this behaviour if requested.

tpikonen:
At least on gnome-shell under wayland event.{xy} and widget.get_position() return constant values after moving the window. IIRC wayland clients do not know their position on screen (but maybe there's a mechanism to ask for it from the compositor). This is maybe true for compositing window managers on X11 as well. Given this, maybe the position saving should be removed from gPodder and left to the windowing system to worry about.

Restoring the window size on other hand is useful, at least on gnome. The default window size is too small to be useful.

auouymous:
I agree, gpodder should only save and restore the size with widget.get_size(), and the position code should be removed.

tpikonen:
I added a patch to this PR which removes saving and restoring window position.

Platypus123:
Oh, just got this in latest version 3.10.18 (2021-04-09) for Windows 10, and it's disappointing that gPodder now doesn't open into the place where I closed it. Any chance of making this behaviour a user option?
I find that gPodder takes about 10 seconds to open, so it's annoying having it open over my main work instead of to the side where I usually keep it. I have to wait 10 seconds for it to finish opening before it can be moved.
Thanks!

auouymous:
@Platypus123 I am working on restoring this feature and will have a new release out in the next day or two.

Platypus123:
That's great! I've actually gone back to 3.10.17 for the time being.
I myself don't understand whether this sort of behaviour option should be managed my the application or by Windows... I read about closing by shift- or ctrl- clicking the X close icon, to make Windows remember the position, but it doesn't seem to work for me..

auouymous:
It has been fixed in 3.10.19.

PR #934, opened 2020-12-27
Add more episode details to shownotes

tpikonen:
Add a line after subheading in shownotes with episode release date, length and size.

I usually have just the 'Released' column visible in the episode list TreeView, so having these visible in the shownotes is useful.

The formatting and the items in the details can be discussed in this PR and changed accordingly.

### DIFF
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -57,6 +57,7 @@ def get_shownotes(enable_html, pane):
 class gPodderShownotes:
     def __init__(self, shownotes_pane):
         self.shownotes_pane = shownotes_pane
+        self.details_fmt = _('%s | %s | %s')
 
         self.scrolled_window = Gtk.ScrolledWindow()
         self.scrolled_window.set_shadow_type(Gtk.ShadowType.IN)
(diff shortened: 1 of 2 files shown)


elelay:
Good idea! It seems maybe a bit crowded.

elelay:
Here is a screenshot on my desktop:
capture

tpikonen:
The SailfishOS version of gPodder 4 has the episode details on the shownotes formatted like '2021-04-28 | 100.0 MiB | 35:00'.

I updated (and rebased) this PR to use this format and print the details with a smaller font. I'm going to use this on the adaptive branch, but it would be nice to have this everywhere.

elelay:
Looks good :+1:

condensed

auouymous:
LGTM, but 0.8 is too small to read. 0.9 is readable.

tpikonen:
I changed the scale to 0.9 and rebased.

auouymous:
`Today | 0.00 B | 00:00`

The columns use a hyphen for duration and nothing for size when they are unknown. Should this do something similar?

elelay:
> The columns use a hyphen for duration and nothing for size when they are unknown. Should this do something similar?

yes, it would be better

tpikonen:
The formatting is now the same as in columns. I'm somewhat itching to replace the "" for unknown size with a "-" though.

auouymous:
I agree with the "-", it looks better with or without a duration.

tpikonen:
Also added html.escape() to the HTML version.

elelay:
Merged, thanks :+1:

auouymous:
What about the unresolved if-else code review?

PR #943, opened 2021-01-01
Set episode selector title before creating window.

auouymous:
Window managers can't restore position when the initial title isn't the same as the final title that gets stored.


elelay:
Merged, thanks :-)

PR #945, opened 2021-01-02
Add a subtitle with statistics to header bar

tpikonen:
I thought it would be good to see how many subscribed channels and available episodes there are in the gpodder DB. The header bar subtitle is currently empty, so here's a patch which prints this info there.

I'm not sure if I caught all the places in the code which result in changes in channel and episode count, but the subtitle is rewritten at least when subscribing, unsubscribing and updating.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -99,6 +99,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             self.header_bar.pack_end(self.application.header_bar_menu_button)
             self.header_bar.pack_start(self.application.header_bar_refresh_button)
             self.header_bar.set_show_close_button(True)
+            self.update_header_bar_subtitle()
             self.header_bar.show_all()
 
             # Tweaks to the UI since we moved the refresh button into the header bar
@@ -327,6 +328,15 @@ class gPodder(BuilderWidget, dbus.service.Object):
             'showToolbar', None, GLib.Variant.new_boolean(self.config.show_toolbar))
         action.connect('activate', self.on_itemShowToolbar_activate)
         g.add_action(action)
+
+    def update_header_bar_subtitle(self, text=None):
+        if self.application.want_headerbar:
+            if text is None:
+                chs = self.db.get('SELECT COUNT(*) FROM %s' % self.db.TABLE_PODCAST)
+                eps = self.db.get('SELECT COUNT(*) FROM %s' % self.db.TABLE_EPISODE)
+                self.header_bar.set_subtitle("%s channels, %s episodes" % (chs, eps))
+            else:
+                self.header_bar.set_subtitle(text)
 
     def inject_extensions_menu(self):
         """
@@ -2535,6 +2545,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                 assert channel is not None
                 worked.append(channel.url)
 
+            self.update_header_bar_subtitle()
             util.idle_add(on_after_update)
 
     def find_episode(self, podcast_url, episode_url):
@@ -2741,6 +2752,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                             self.pbFeedUpdate.set_text(message)
 
                     self.show_update_feeds_buttons()
+                    self.update_header_bar_subtitle()
 
             util.idle_add(update_feed_cache_finish_callback)
 
@@ -2835,6 +2847,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
             self.update_episode_list_icons(episode_urls)
             self.update_podcast_list_model(channel_urls)
+            self.update_header_bar_subtitle()
             self.play_or_download()
 
         @util.run_in_background
@@ -3263,6 +3276,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
             # Re-load the channels and select the desired new channel
             self.update_podcast_list_model(select_url=select_url)
+            self.update_header_bar_subtitle()
             progress.on_finished()
 
         @util.run_in_background


auouymous:
This feature is only available for those using CSD. Does it make use of currently empty space in the window, or does it take space away from the channel and episode lists? I can understand total downloaded episodes, but why would you want to always see the total channels and episodes? Maybe the total counts could be displayed as labels in one of the menus?

I've been trying to find space to display the total size and duration of all downloaded episodes, and total downloaded episodes might also be nice. The only empty space is the titlebar (no header bar) when not downloading. If subtitle is similar, `update_header_bar_subtitle` could be renamed and update titlebar or subtitle.

tpikonen:
I'm mostly interested in the channel count, as for episodes you are right, downloaded episode count would be more useful than total episodes.

The subtitle is the second row in the headerbar, with a smaller non-bold font. Currently it's not used for anything. The title update function could of course behave differently depending on want_headerbar.

elelay:
It doesn't add height to the header bar but moves the title up to make space for a line of text
No subtitle : no_subtitle

With subtitle subtitle

elelay:
Thanks @tpikonen: I generally like the idea of this PR.

To move it forward, one needs:
 - to change the information displayed to *CHANNEL_COUNT channels, DOWNLOADED_EPISODE_COUNT episodes, TOTAL_TIME play duration, TOTAL_SIZE disk usage* with any variation in the wording
 - to display this same information after "gPodder" in the titlebar (when not using a header bard) when there is no download progress

auouymous:
Are labels really needed? They require translations and waste space for anyone with tiny titlebars (narrow windows with half width titlebars). "gPodder - CHANNELS channels - EPISODES episodes - HH:MM:SS - 123 MiB". I don't know how to get rid of the channel/episode labels, but it would be nice if there was a general preferences checkbox to toggle channel/episodes counts off, because my titlebar will barely hold the duration and size. Or maybe a checkbox to toggle labels for all stats, it defaults on, and anyone who needs more space can turn them and still know what the values mean.

Channel/episodes counts or not option:
```
"gPodder - HH:MM:SS - 123 MiB"
"gPodder - CHANNELS channels - EPISODES episodes - HH:MM:SS - 123 MiB"
```

Labels or no labels option:
```
"gPodder - CHANNELS - EPISODES - HH:MM:SS - 123 MiB"
"gPodder - CHANNELS channels - EPISODES episodes - HH:MM:SS play duration - 123 MiB disk usage"
```

auouymous:
This is a *little* modification of your patch with duration and size, and ability to toggle labels off (in the code). The time and size are for all downloaded episodes, but it would probably be better to show only downloaded episodes in the selected channel or section instead (not sure how, yet).

The channels only shows total channels, which is a **very** nice stat to have, but some might want to see only channel count for download episodes. Maybe show total if a single channel is selected and show downloaded channel count when all or section is selected?

The episode count is questionable, it shows all episodes, and I'm not sure why anyone would want to see that value. Might be better to show downloaded episodes for all, section or channel. Or maybe "downloaded / total" for all, section or channel?

```patch
diff --git a/src/gpodder/gtkui/main.py b/src/gpodder/gtkui/main.py
index 4c199294..a79684ac 100644
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -99,6 +99,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             self.header_bar.pack_end(self.application.header_bar_menu_button)
             self.header_bar.pack_start(self.application.header_bar_refresh_button)
             self.header_bar.set_show_close_button(True)
+            self.update_title_statistics()
             self.header_bar.show_all()
 
             # Tweaks to the UI since we moved the refresh button into the header bar
@@ -167,6 +168,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         self.default_title = None
         self.set_title(_('gPodder'))
+        self.update_title_statistics()
 
         self.cover_downloader = CoverDownloader()
 
@@ -328,6 +330,34 @@ class gPodder(BuilderWidget, dbus.service.Object):
         action.connect('activate', self.on_itemShowToolbar_activate)
         g.add_action(action)
 
+    def update_title_statistics(self, text=None):
+        if text is None:
+            labels = True
+
+            title = [] if self.application.want_headerbar else [self.default_title]
+            chs = self.db.get('SELECT COUNT(*) FROM %s' % self.db.TABLE_PODCAST) or 0
+            if chs > 0:
+                title.append((_("%s channels") if labels else "%s") % chs)
+            eps = self.db.get('SELECT COUNT(*) FROM %s' % self.db.TABLE_EPISODE) or 0
+            if eps > 0:
+                title.append((_("%s episodes") if labels else "%s") % eps)
+            time = self.db.get('SELECT SUM(total_time) FROM %s WHERE state = 1' % self.db.TABLE_EPISODE) or 0
+            if time > 0:
+                title.append((_("%s duration") if labels else "%s") % util.format_time(time))
+            size = self.db.get('SELECT SUM(file_size) FROM %s WHERE state = 1' % self.db.TABLE_EPISODE) or 0
+            if size > 0:
+                title.append((_("%s disk usage") if labels else "%s") % util.format_filesize(size, digits=1))
+
+            if self.application.want_headerbar:
+                self.header_bar.set_subtitle(' - '.join(title))
+            else:
+                self.gPodder.set_title(' - '.join(title))
+        else:
+            if self.application.want_headerbar:
+                self.header_bar.set_subtitle(text)
+            else:
+                self.gPodder.set_title(' - '.join([self.default_title, text]))
+
     def inject_extensions_menu(self):
         """
         Update Extras/Extensions menu.
@@ -1237,7 +1267,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
                 # Stop updating the download list here
                 self.download_list_update_enabled = False
 
-            self.gPodder.set_title(' - '.join(title))
+            if len(title) > 1:
+                self.gPodder.set_title(' - '.join(title))
+            else:
+                self.update_title_statistics()
 
             self.update_episode_list_icons(episode_urls)
             self.play_or_download()
@@ -2554,6 +2587,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                 assert channel is not None
                 worked.append(channel.url)
 
+            self.update_title_statistics()
             util.idle_add(on_after_update)
 
     def find_episode(self, podcast_url, episode_url):
@@ -2760,6 +2794,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                             self.pbFeedUpdate.set_text(message)
 
                     self.show_update_feeds_buttons()
+                    self.update_title_statistics()
 
             util.idle_add(update_feed_cache_finish_callback)
 
@@ -2854,6 +2889,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
             self.update_episode_list_icons(episode_urls)
             self.update_podcast_list_model(channel_urls)
+            self.update_title_statistics()
             self.play_or_download()
 
         @util.run_in_background
@@ -3282,6 +3318,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
             # Re-load the channels and select the desired new channel
             self.update_podcast_list_model(select_url=select_url)
+            self.update_title_statistics()
             progress.on_finished()
 
         @util.run_in_background
```

tpikonen:
It seems to me that this kind of configurability would require a format string for the extra title / subtitle, which would be documented in the preferences dialog. I'm not sure if it's too much for this rather simple feature, but I could implement it if it's ok for everyone.

auouymous:
I would love a format string with several other options, but ya, it might be overkill. It wouldn't be able to hide zero-value options like the patch does, but I'd still use it. @elelay What do you think?

elelay:
Sorry for the late reply!

Yes, it seems overkill to make it that configurable.
I'm not a fan of changing it depending on the selection: it would be ok for a status bar, but isn't the subtitle for global info?
Regarding short/long form: would it make sense to adapt to the window's width?
Regarding contents, I'd rather see downloaded episode count than total episode count.

auouymous:
Changing the `chs` and `eps` queries in my patch to the following shows number of channels with downloaded episodes, and number of downloaded episodes. It makes more sense than showing total channels and total episodes because it matches the duration and size which is only for downloaded episodes.

```
chs = self.db.get('SELECT COUNT(DISTINCT podcast_id) FROM %s WHERE state = 1' % self.db.TABLE_EPISODE) or 0

eps = self.db.get('SELECT COUNT(id) FROM %s WHERE state = 1' % self.db.TABLE_EPISODE) or 0
```

Detecting window width to disable labels wouldn't work. Somone could have a very tiny font, or very large font. And in my case, my titlebar is literally half the width of the window.

elelay:
So much for detecting window width, then.

I'd go with the long form, because it's just random numbers otherwise. Would you want to add an advanced config for small screen, that would enable this and more tweaks? Because like with the Progress-as-dialog PR there seems to me that there is a conflict between screen real estate and usability,  don't you think?

Channel count not paused seems more informative, regardless of downloaded episodes. Maybe this format (downloads could be taken as currently downloading so not great?)?
````
gPodder - CHANNELS active subscriptions - EPISODES downloads  - HH:MM:SS playback - 123 MiB disk usage
````

auouymous:
> Would you want to add an advanced config for small screen, that would enable this and more tweaks?

The default would show labels, but anyone who knows the labels and has a small window could turn them off and know what the values mean. I have been using the patch, and can process the info much faster without labels than with labels.

> Because like with the Progress-as-dialog PR there seems to me that there is a conflict between screen real estate and usability, don't you think?

I assume you mean the PR to remove the GtkNotebook widget? What is the conflict you are refering to? That PR would free up space to display more episodes. This PR is adding a new feature and doesn't use any additional space.

> Channel count not paused seems more informative, regardless of downloaded episodes, because it's what's going to be refreshed.

What if active channel count was displayed in the Podcasts tab as `%(active)d Podcasts` or `%(active)d/%(total)d Podcasts`? The titlebar/subtitle could display downloaded channels to match the other three stats, and the total count would be shown in the tab. If the GtkWidget was ever removed, the active/total count could be displayed in a label at the top of the channel column.

> Maybe this format (_downloads_ could be taken as currently downloading so not great?)?

Maybe 'downloaded' instead of 'downloads'? But there is a downloading progress meter in titlebar while downloading. Without subtitle, no one will see this PR while downloads are active. And with subtitle, both are visible so maybe blank the subtitle while downloading to avoid confusion about the playback time being remaining download time.

elelay:
> > Would you want to add an advanced config for small screen, that would enable this and more tweaks?
> 
> The default would show labels, but anyone who knows the labels and has a small window could turn them off and know what the values mean. I have been using the patch, and can process the info much faster without labels than with labels.

I'm not sure if you agree with the "tiny-screen" advanced config or want a specific option for this. Please clarify.

> > Because like with the Progress-as-dialog PR there seems to me that there is a conflict between screen real estate and usability, don't you think?
> 
> I assume you mean the PR to remove the GtkNotebook widget? What is the conflict you are refering to? That PR would free up space to display more episodes. This PR is adding a new feature and doesn't use any additional space.

Yes I mean #938 where I think there was no satisfying (to me) placement of the buttons to show/hide the progress (toolbar or channel list). Showing just numbers feels the same to me: yes, screen usage is reduced vs numbers with labels; but a newcomer has to guess what they mean.
 
> > Channel count not paused seems more informative, regardless of downloaded episodes, because it's what's going to be refreshed.
> 
> What if active channel count was displayed in the Podcasts tab as `%(active)d Podcasts` or `%(active)d/%(total)d Podcasts`? The titlebar/subtitle could display downloaded channels to match the other three stats, and the total count would be shown in the tab. If the GtkWidget was ever removed, the active/total count could be displayed in a label at the top of the channel column.

Yes, it seems a good idea to display active channels / total in podcasts tab.

> > Maybe this format (_downloads_ could be taken as currently downloading so not great?)?
> 
> Maybe 'downloaded' instead of 'downloads'? But there is a downloading progress meter in titlebar while downloading. Without subtitle, no one will see this PR while downloads are active. And with subtitle, both are visible so maybe blank the subtitle while downloading to avoid confusion about the playback time being remaining download time.

Yes, blank subtitle during downloads.

PR #946, opened 2021-01-02
Narrow down About dialog, pack items vertically

tpikonen:
This patch makes the about dialog narrower, so that it fits better to a narrow screen.


auouymous:
LGTM, especially since it adds a margin to keep the text from touching the dialog border. But the redundant "License:" text is now stacked directly above the "License" word in the license, @elelay could it be removed?

PR #949, opened 2021-01-10
Allow episodes with youtube-dl compatible URLs

tpikonen:
This PR allows subscribing to feeds with links to youtube-dl compatible content and a config var to enable streaming of unrecognized episode URLs with the configured video player.

Specifically:
 * Allow creation of custom downloaders  in the youtube-dl extension for all URLs supported by youtube-dl
 * Accept episodes in RSS feeds where the 'link' element has an URL for which a custom downloader can be created
 * Remove legacy settings 'player' and 'videoplayer'
 * Add setting 'player.videoplayer_youtube_dl_support' which enables streaming of episodes with URLs with otherwise unrecognized file type

Downloads with youtube-dl work if a suitable format string is set to youtube.preferred_fmt_ids, but this setting is a bit non-obvious for my taste.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -438,8 +438,6 @@ class gPodderYoutubeDL(download.CustomDownloader):
             return None
         if self.is_supported_url(episode.url):
             return YoutubeCustomDownload(self, episode.url, episode)
-        elif self.is_supported_url(episode.link):
-            return YoutubeCustomDownload(self, episode.link, episode)
 
         return None
 
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -285,6 +285,10 @@ class gPodderYoutubeDL(download.CustomDownloader):
             self._ydl_opts['verbose'] = True
         else:
             self._ydl_opts['quiet'] = True
+        # Don't create downloaders for URLs supported by these youtube-dl extractors
+        self.ie_blacklist = ["Generic"]
+        # Cache URL regexes from youtube-dl matches here, seed with youtube regex
+        self.regex_cache = [re.compile(r'https://www.youtube.com/watch\?v=.+')]
         # #686 on windows without a console, sys.stdout is None, causing exceptions
         # when adding podcasts.
         # See https://docs.python.org/3/library/sys.html#sys.__stderr__ Note
@@ -411,16 +415,32 @@ class gPodderYoutubeDL(download.CustomDownloader):
             return self.refresh(url, channel.url, max_episodes)
         return None
 
+    def is_supported_url(self, url):
+        if self.regex_cache[0].match(url) is not None:
+            return True
+        for r in self.regex_cache[1:]:
+            if r.match(url) is not None:
+                self.regex_cache.remove(r)
+                self.regex_cache.insert(0, r)
+                return True
+        with youtube_dl.YoutubeDL(self._ydl_opts) as ydl:
+            for ie in ydl._ies:
+                if ie.suitable(url) and ie.ie_key() not in self.ie_blacklist:
+                    self.regex_cache.insert(0, ie._VALID_URL_RE)
+                    return True
+        return False
+
     def custom_downloader(self, unused_config, episode):
         """
         called from registry.custom_downloader.resolve
         """
         if not self.force and not self.my_config.manage_downloads:
             return None
-        if re.match(r'''https://www.youtube.com/watch\?v=.+''', episode.url):
+        if self.is_supported_url(episode.url):
             return YoutubeCustomDownload(self, episode.url, episode)
-        elif re.match(r'''https://www.youtube.com/watch\?v=.+''', episode.link):
+        elif self.is_supported_url(episode.link):
             return YoutubeCustomDownload(self, episode.link, episode)
+
         return None
 
 
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -327,6 +327,10 @@ class PodcastEpisode(PodcastModelObject):
         if file_type is not None:
             return episode
 
+        # Check if any extensions (e.g. youtube-dl) support the link
+        if registry.custom_downloader.resolve(None, None, episode) is not None:
+            return episode
+
         return None
 
     def __init__(self, channel):
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -290,18 +290,26 @@ class PodcastEpisode(PodcastModelObject):
 
         audio_available = any(enclosure['mime_type'].startswith('audio/') for enclosure in entry['enclosures'])
         video_available = any(enclosure['mime_type'].startswith('video/') for enclosure in entry['enclosures'])
+        link_has_media = False
+        if not (audio_available or video_available):
+            _url = episode.url
+            episode.url = util.normalize_feed_url(entry['link'])
+            # Check if any extensions (e.g. youtube-dl) support the link
+            link_has_media = registry.custom_downloader.resolve(None, None, episode) is not None
+            episode.url = _url
+        media_available = audio_available or video_available or link_has_media
 
         for enclosure in entry['enclosures']:
             episode.mime_type = enclosure['mime_type']
 
             # Skip images in feeds if audio or video is available (bug 979)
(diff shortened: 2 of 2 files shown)


elelay:
interesting :+1:

auouymous:
Commit 386afb1 is not correct. I'm not sure what you are trying to do with it but both regex are identical, the first is checked against episode.url and second against episode.link. Compiling the regex is good idea though.

tpikonen:
I must have been asleep when I made the previous push, it should work now.

But, is checking episode.link really necessary? PodcastEpisode.from_podcastparser_entry() in model.py does not create an episode object without a verified episode.url. Testing the 'link' property when creating the downloader would be then useless, unless episodes are created some other way.

auouymous:
I don't know why the `episode.link` check happens but your cache patch appears to return `episode.url` for the `episode.link` branch.

As for the force default download, does that mean youtube-dl won't be used to download a DRM video that uses a known mime type?

elelay:
@tpikonen What's the point of 0406835?

tpikonen:
@elelay youtube-dl supports URLs to downloadable files, so it will be always used as a downloader unless some kind of filtering is used. The commit was a quick fix to use the default downloader for the usual media file mime types.

On second thought it's better to make this filtering in the extension itself. Suggestions on what mime types, extensions etc. should and should not be downloaded by youtube-dl would be welcome.

tpikonen:
The extension now rejects downloads which would use the Generic youtube-dl extractor. This is a bit slow way to check for suitable URLs (0.3 s on my i5 laptop), but works as intended, so that normal media files are downloaded with the default downloader. Prefiltering by mime type or file extension from URL could be still implemented on top of this.

I tried to look if there are places in gpodder code where episodes without a valid 'url' attribute are be created, but could not find any, so I removed support for URLs in episode.link. If code creating episodes without an 'url' is there, I think it's better to treat it as buggy.

Also rebased to current master.

tpikonen:
This PR currently contains basically two changes:

1. Make the youtube-dl extension support all the URLs supported by youtube-dl and accept episodes from feeds where the episode.link element is supported by youtube-dl or other downloader.
2. Stream episode.url links directly with the video player if a custom downloader exists and a config var is set.

The second change is arguably a bit hacky, but would the first change be acceptable? I can split these two into separate PRs if so.

auouymous:
What feeds does this PR fix?

tpikonen:
I'm using this to follow radio and tv shows from the Finnish broadcaster Yle. The feeds are at https://feeds.yle.fi/ (example: https://feeds.yle.fi/areena/v1/series/1-50697851.rss). There are some feeds there with normal media enclosures, but most are only accessible from the streaming service at areena.yle.fi via the episode link.

share/gpodder/extensions/youtube-dl.py
```diff
@@ -285,6 +285,10 @@ def __init__(self, gpodder_config, my_config, force=False):
             self._ydl_opts['verbose'] = True
         else:
             self._ydl_opts['quiet'] = True
+        # Don't create downloaders for URLs supported by these youtube-dl extractors
+        self.ie_blacklist = ["Generic"]
+        # Cache URL regexes from youtube-dl matches here, seed with youtube regex
+        self.regex_cache = [re.compile(r'https://www.youtube.com/watch\?v=.+')]
```
elelay, line 291: instead of maintaining the regex_cache, can you load all regexps and combine them at startup?
(untested)
```
all_patterns = [(r'https://www.youtube.com/watch\?v=.+')] + [ie._VALID_URL_RE.pattern for ie in ydl._ies if ie.ie_key() not in self.ie_blacklist]
self.regex_cache = re.compile('((?:' + ')|(?:'.join(all_patterns) + '))')
```
Please try it out and let us know if it's more efficient than managing the cache regex list.

share/gpodder/extensions/youtube-dl.py
```diff
@@ -411,16 +415,30 @@ def fetch_channel(self, channel, max_episodes=0):
...
+        return False
+
     def custom_downloader(self, unused_config, episode):
         """
         called from registry.custom_downloader.resolve
         """
         if not self.force and not self.my_config.manage_downloads:
             return None
-        if re.match(r'''https://www.youtube.com/watch\?v=.+''', episode.url):
+        if self.is_supported_url(episode.url):
```
elelay, line 439: I'm happy with not checking `episode.link` as long as you tested with and without youtube_dl managing subscriptions.

src/gpodder/model.py
```diff
@@ -514,8 +518,12 @@ def get_playback_url(self, config=None, allow_partial=False):
             return url + '.partial'
 
         if url is None or not os.path.exists(url):
-            # FIXME: may custom downloaders provide the real url ?
-            url = registry.download_url.resolve(config, self.url, self, allow_partial)
+            if config.player.videoplayer_youtube_dl_support and registry.custom_downloader.resolve(None, None, self) is not None:
```
elelay, line 521: Please can you clarify what the issue is with `registry.download_url.resolve`?

elelay (changes requested):
Please also split out the player handling (and update to player.audio) in a separate PR.

share/gpodder/extensions/youtube-dl.py
tpikonen: Ok. It's still a separate commit in the current patch stack, but can be squashed if necessary.

src/gpodder/model.py
tpikonen: Note that the FIXME comes from the code in current master. 

As far as the question 'may custom downloaders provide the real url ?' goes, I think the answer in the case of youtube and similar is 'No'. Simple methods of obtaining a download URL from e.g. youtube (like in gPodders youtube.py) do not always seem to work, while youtube-dl does.

elelay: Thanks for taking the time to look into it! I don't have this magic function readily available, unfortunately.

share/gpodder/extensions/youtube-dl.py
elelay: Thanks for taking the time to try it out! I haven't this function readily available, unfortunately

tpikonen:
Rebased to master. @elelay are there further issues with the code?

elelay:
LGTM, but I can't test the improvement it brings

auouymous:
I agree, the code looks good but https://feeds.yle.fi/areena/v1/series/1-50697851.rss still doesn't contain any episodes.

tpikonen:
The YLE feeds are a bad example, since they need a [patched](https://github.com/tpikonen/youtube-dl/tree/yle-dl) youtube-dl, but that's what I use this PR for.

There are 3rd party [feeds](https://bbc-rss.herokuapp.com/) for the BBC iPlayer which work, modulo geoblocking (and the server is really slow).

Then there is [RSS-bridge](https://github.com/RSS-Bridge/rss-bridge) which has feeds for lots of things, like Bandcamp ([example](https://rss.nixnet.services/?action=display&bridge=Bandcamp&context=By+band&band=draama-helmi&type=tracks&limit=20&format=Mrss)).

These feeds have images in enclosures, so I changed the feed parsing to reject them in favor of episode.link, when there is something to download there.

Also, the youtube-dl extension needs improving in format handling, but here's quick hack if anyone wants to test the feeds above:

```
diff --git a/share/gpodder/extensions/youtube-dl.py b/share/gpodder/extensions/youtube-dl.py
index 2704ca8a..3368f7b2 100644
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -311,6 +311,7 @@ class gPodderYoutubeDL(download.CustomDownloader):
         opts['format'] = '/'.join(str(fmt) for fmt in fmt_ids)
         if fallback:
             opts['format'] += '/' + fallback
+        opts['format'] += '/bestvideo+bestaudio[abr<=256]/bestaudio[abr<=256]/best'
         logger.debug('format=%s', opts['format'])
 
     def fetch_video(self, url, tempname, reporthook):
```

elelay:
Merged, thanks

tpikonen:
Thanks for merging. I have some ideas to improve #963 so that streaming can be made to work  for ytdl-links as well.

elelay:
Thanks, I've seen your comment above about improving format handling in the youtube-dl extension.
Looking forward to it :+1:

PR #951, opened 2021-01-10
Skip episode media with 'application/*' mime type

tpikonen:
Some peertube feeds (example: https://tilvids.com/feeds/videos.xml?videoChannelId=42) have 'application/x-bittorrent' URLs in the 'enclosure' tags. This patch makes gpodder skip all URLs with 'application/' mime type, if there are enclosures with audio or video available.


elelay:
Merged,
thanks :-)

PR #954, opened 2021-01-15
Unittest fixes

tpikonen:
This PR fixes some problems I had with 'make unittest'.

 * Use either 'pytest' or 'pytest-3' for the pytest executable. The latter is the default name for pytest at least in Debian.
 * Reduce output clutter by fixing regexp string DeprecationWarnings. Regexp patterns are changed to raw strings in files which had warnings. There are some changes in the actual string contents (like '\\\\w' -> '\w'), so it would be good if someone else checks these before committing.
 * Add the coverage database to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ __pycache__
 *~
 *.in.h
 *.ui.h
+.coverage
 src/podcastparser.py
 src/mygpoclient
 src/dbus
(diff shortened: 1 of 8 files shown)


elelay:
looks good

tpikonen:
Hi @elelay, is there anything preventing this PR from being merged?

elelay:
hi,
nothing, really. It simply fell through the cracks.
Thanks,

PR #957, opened 2021-01-28
Strip HTML tags from channel descriptions

tpikonen:
Run util.remove_html_tags() on channel description strings shown in the UI.


auouymous:
LGTM, except for the naming of `channeldesc`, something more descriptive like `channel_text_description` or even just `channel_description` might be better.

tpikonen:
Changed the variable named as suggested.

elelay:
Merged, thanks :-)

PR #960, opened 2021-02-01
Update ru.po

adasiko:
Update russian localization


elelay:
Merged, thanks!
If you're willing to continue updating gPodder translations, please subscribe to #518. I post a comment there to notify translators a week before release.

PR #962, opened 2021-02-06
Remove deprecated gtk elements

paper42:


auouymous:
Prefs...General:
- horizontal gaps around combo boxes are wider, which now matches the margin (nice)
- but there is no vertical gap between the two combo boxes

Prefs...gpodder.net:
- looks better

Prefs...Updating:
- no changes, but should probably get wider horizontal gaps like the other tabs

Prefs...Clean-up:
- same as the Updating tab

Prefs...Devices:
- the first two rows lack the wider gap which the other two labels have
- mountpoint is the only indented/aligned label

paper42:
> Prefs...Devices:
> 
> * mountpoint is the only indented/aligned label

I am not sure what you mean by that. Is it fixed now?

Other than that, I think I managed to address all your points. Thank you for them.

auouymous:
The update interval slider in Prefs...Updating is still missing a wider gap.

In Prefs...Devices, device type and mountpoint are in a grid with the labels aligned to the right side of the column. Add a bunch of characters to the device type or mountpoint labels to better see the issue.

paper42:
Thanks, fixed.

auouymous:
It now looks like device type and mountpoint are centered in the column, is it not possible to left align those two?

auouymous:
Adding `<property name="xalign">0</property>` before the device type and mountpoint labels fixes the problem.

```
<property name="xalign">0</property>
<property name="label" translatable="yes">Device type:</property>
...
<property name="xalign">0</property>
<property name="label" translatable="yes">Mountpoint:</property>
```

paper42:
I am so sorry for the delay, I fixed it. Thanks for your patience.

auouymous:
LGTM.

@elelay Are you okay with the new padding and alignment changes?

elelay:
Great, thanks

PR #965, opened 2021-02-13
Disable open website menu item if channel or episode has no website.

### LINKED ISSUE #958: "Open website" menu link is always enabled for each podcast episode, even if there's no website
I'm using gPodder 3.10.17 (2020-11-23) on Windows 10, 64-bit. The installed version.

This specifically refers to the extension under **Interface** called **"Open website" episode and podcast context menu**. This file looks relevant: https://github.com/gpodder/gpodder/blob/master/share/gpodder/extensions/episode_website_context_menu.py

Each podcast episode has a right-click menu, which has an "Open website" item when that extension is active. This item is always enabled, even if there's no website associated with the episode.

_If there is an associated website_, it performs as expected: it opens the website. Example podcasts with episode links:
- https://feeds.megaphone.fm/darknetdiaries
- https://feeds.simplecast.com/JT6pbPkg

_If there is not an associated website_, Windows Explorer opens and shows C:\Program Files (x86)\gPodder. For me that's the folder with "contents.txt", "uninstall.exe" and a few folders. Example podcasts with no episode links:
- http://feeds.gimletmedia.com/heavyweightpodcast
- https://feeds.megaphone.fm/TPG9562451353

My suggestion would be to disable the "Open website" menu item, if there's no website associated with the episode.

auouymous:
Extensions can return [] to hide a menu item, but making it insensitive is less confusing than hiding it. The first patch allows extensions to disable menu items by returning None for the callback.
 
Closes #958.

### DIFF
--- a/share/gpodder/extensions/episode_website_context_menu.py
+++ b/share/gpodder/extensions/episode_website_context_menu.py
@@ -23,15 +23,21 @@ class gPodderExtension:
     def __init__(self, container):
         self.container = container
 
+    def has_website(self, episodes):
+        for episode in episodes:
+            if episode.link:
+                return True
+
     def open_website(self, episodes):
         for episode in episodes:
-            util.open_website(episode.link)
+            if episode.link:
+                util.open_website(episode.link)
 
     def open_channel_website(self, channel):
         util.open_website(channel.link)
 
     def on_episodes_context_menu(self, episodes):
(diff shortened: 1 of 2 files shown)


elelay:
That's nice!

elelay:
Merged, thanks.
Returning a None calback to mean *disable* is perfect for retro-compatibility and elegant.

PR #966, opened 2021-02-13
Add combined episode duration and size columns.

auouymous:
This reduces horizontal space required to display both size and duration.


elelay:
size can be None :
none_size
eg https://services.premierenetworks.com/coast/podcast.xml

auouymous:
Fixed. Funny how I've used this for a year or so and didn't notice that, even though it is present in all of my feeds.

elelay:
merged, thanks

Issue #967, opened 2021-02-28
Cannot subscibe Logbuch Netzpolitik


mase76:
Hallo!
I cannot add the feed of Logbuch Netzpolitik (https://logbuch-netzpolitik.de/feed/m4a).
This is the error message I get:
Could not parse feed: <unknown> 671:3: not well-formed (invalid token)

Regards
Thomas

auouymous:
You need to contact the feed owner and ask them to fix it. The first page of the feed validates with some minor issues. But it contains `<atom:link rel="next" href="https://logbuch-netzpolitik.de/feed/m4a?paged=2"/>` which does not validate https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Flogbuch-netzpolitik.de%2Ffeed%2Fm4a%3Fpaged%3D2.

PR #968, opened 2021-03-03
Show URL that caused invalid feed error

auouymous:
The error may originate from another page in the feed, and only showing the original URL makes it harder to track down invalid feed issues (see #967).


elelay:
Merged, thanks!

PR #970, opened 2021-03-05
Fix warning on non-200 status codes in coverart

paper42:
before:
```
[gpodder.coverart] WARNING: Cannot save cover art
  File "/home/michal/build/gpodder/src/gpodder/coverart.py", line 103, in get_cover
    raise ValueError(msg)
ValueError: Unknown file type: <URL> (b'')
```

after:
```
[gpodder.coverart] WARNING: Cover art download failed: <URL> returned status code 404
```


elelay:
Nice, thanks!

Issue #974, opened 2021-03-09
"No network connection" while connected to the internet


savannidgerinel:
This has been going on for about a year.

Starting with a fresh gPodder directory, I can import my podcasts from an OPML file and then download all of them. However, even while the download is still in progress, "Check for new episodes" always gives me this error popup:

"No network connection. Please connect to a network and try again."

I am only able to refresh podcasts by deleting the gPodder directory and going through the process again.

The log file is empty and I have no messages on the command line.

gpodder 3.10.17
Packaged for nipkgs 20.09: https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/audio/gpodder/default.nix

tpikonen:
Could you try running 'gpodder -v' in a terminal to get a verbose debug log?

savannidgerinel:
Here's an entire session up until the moment I dismiss the popup:

```
[] savanni@garnet:~ $ gpodder -v
1615302251.940930 [gpodder.log] DEBUG: ==== gPodder starts up (ui=gtk) ===
1615302251.994480 [gpodder.registry] DEBUG: Registering download_url resolver: <function youtube_real_download_url at 0x7f8e0eeb43a0>
1615302251.995791 [gpodder.registry] DEBUG: Registering download_url resolver: <function escapist_real_download_url at 0x7f8e0ee8d820>
1615302251.997540 [gpodder.registry] DEBUG: Registering download_url resolver: <function vimeo_real_download_url at 0x7f8e0eeb8b80>
1615302252.131492 [eyed3] DEBUG: eyeD3 version 0.9.6
1615302252.172795 [gpodder.gtkui.shownotes] INFO: No WebKit2 gobject bindings, so no HTML shownotes
1615302252.188097 [gpodder.extensions] DEBUG: Found extension "audio_converter" in /nix/store/s81y7155hfpqrqx7c74plqdsbh1za5mx-gpodder-3.10.17/share/gpodder/extensions/audio_converter.py
1615302252.190142 [gpodder.extensions] DEBUG: Found extension "command_on_download" in /nix/store/s81y7155hfpqrqx7c74plqdsbh1za5mx-gpodder-3.10.17/share/gpodder/extensions/command_on_download.py
1615302252.190496 [gpodder.extensions] DEBUG: Found extension "concatenate_videos" in /nix/store/s81y7155hfpqrqx7c74plqdsbh1za5mx-gpodder-3.10.17/share/gpodder/extensions/concatenate_videos.py
[...]

PR #975, opened 2021-03-09
Fix connectivity checking

tpikonen:
Here's a proposed fix for #974.

I'm not 100% sure on second commit which uses 'ip addr show scope global up' instead of 'ip link' to get a list of candidate interfaces, but reading the man page I think a global interface is what we're after.

BSDs and MacOS do not have 'ip' installed by default, so they should fall back to 'ifconfig'.

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1824,17 +1824,22 @@ def run_in_background(function, daemon=False):
 
 
 def linux_get_active_interfaces():
-    """Get active network interfaces using 'ip link'
+    """Get active network interfaces using 'ip addr'
 
-    Returns a list of active network interfaces or an
-    empty list if the device is offline. The loopback
-    interface is not included.
+    A generator function yielding network interface
+    names with an inet (or inet6) and a broadcast
+    address, indicating an active network connection
     """
-    process = Popen(['ip', 'link'], close_fds=True, stdout=subprocess.PIPE)
-    data, _ = process.communicate()
-    for interface, _ in re.findall(r'\d+: ([^:]+):.*state (UP|DORMANT|UNKNOWN)', data.decode(locale.getpreferredencoding())):
-        if interface != 'lo':
-            yield interface
+    process = Popen(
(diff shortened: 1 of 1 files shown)


auouymous:
Doesn't fix the issue, see https://github.com/gpodder/gpodder/issues/974#issuecomment-794510475.

tpikonen:
You're right @auouymous, it didn't work on busybox. I pushed a new version which checks for interfaces with an inet and broadcast address, which apparently is a sign of active connection. This works at least with 'busybox ip' and 'ip' in Debian.

elelay:
Merged, thanks!

auouymous:
Little late, but the patch works fine and looks like it should solve the issue.

One problem I see, that existed before this patch, is that all three `*_get_active_interfaces` methods yield the result. Is there a reason for this? They are only called once and returning the result would be more efficient, right?

tpikonen:
I also don't see the point for using generator functions in this case, but a patch fixing all these functions would have been a lot more invasive (and would have to be tested on all platforms).

elelay:
yes, no need for a generator, and no point, really, since all the command output is gathered before analyzing it.

auouymous:
I wrote a patch the other day that renamed those three functions and returned a bool. But are those functions usable by extensions? If so, do we care about breaking extensions not in gpodder? They are only called once per manual or automatic update, so performance isn't that important. There are also some functions in util.py not used anywhere in gpodder.

elelay:
> But are those functions usable by extensions? If so, do we care about breaking extensions not in gpodder? 
yes on both counts. Are those used, though? Not likely. An extension would use the `connection_available()` function.
Personally I wouldn't change them, but if you find it irksome, you may change them.

> There are also some functions in util.py not used anywhere in gpodder.
which ones?

auouymous:
They are rarely called so I'm fine leaving them alone.

`parse_time`, `isabs`, `write_m3u_playlist`, `check_command` and `website_reachable` are unused.

PR #978, opened 2021-03-16
Subscription statistics extension

auouymous:
Is anyone interested in an extension that shows the interval between episodes per channel? It adds a dialog to the Extras menu that sorts channels into daily, weekly, monthly and yearly categories. And might be useful for finding and removing dead subscriptions. Or showing that you have too many frequently updated podcasts. ;)

### DIFF
--- a/share/gpodder/extensions/subscription_stats.py
+++ b/share/gpodder/extensions/subscription_stats.py
@@ -1,0 +1,157 @@
+# -*- coding: utf-8 -*-
+# Show publishing statistics for subscriptions.
+# Released under the same license terms as gPodder itself.
+
+import time
+
+import gpodder
+
+import gi  # isort:skip
+gi.require_version('Gtk', '3.0')  # isort:skip
+from gi.repository import Gtk, Pango  # isort:skip
+
+_ = gpodder.gettext
+
+__title__ = _('Subscription Statistics')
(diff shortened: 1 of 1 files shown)


elelay:
Nice :-)

anewuser:
Thank you for this extension, but could you add a simple list showing when each subscription was last updated too?  

With the current algorithm, these two podcasts seem to be equally "dead", while in fact Aluku Rebels was updated last month:

gPodder

* https://jodiedenis.podomatic.com/rss2.xml - last update: June 2022
* https://feeds.feedburner.com/Controlo - last update: October 2019

auouymous:
Well, 90 days isn't technically dead, it is just a podcast with a very low update frequency. I have two that release once a year.

But a last updated column is a good idea and I'll add it after I get the release out.

anewuser:
The second show was last updated 2 years and 8 months ago. It *is* dead. The thing is that, while some podcasts get scarce updates, this one used to be updated frequently and then suddenly stopped for good. Comparing averages can be misleading.

Anyway, thank you for adding this idea to your to-do list!

ndias:
@anewuser  You can try this https://github.com/gpodder/gpodder/pull/1590

auouymous:
Implemented by #1590.

PR #979, opened 2021-03-16
Update soundcloud.py

thisfred:
Fix usage of a variable that was renamed in this commit: https://github.com/gpodder/gpodder/commit/3babb86c70cd136abcca7e035eaa154d4526c9f5 (bare except was hiding the error, but flake8 or pylint would have spotted this.)


elelay:
Merged, thanks! Good catch!

PR #983, opened 2021-03-19
Show channel or episode titles in delete dialog

auouymous:
Provides a second, and better, indicator of which channel or episode is being deleted. Only 8 titles up to 60 characters each are shown to avoid excessively large dialogs.

The titles can blend with the final delete message. Should the final message be bold? Or should titles have a prefix such as '• '?

Should the 8 and 60 be adjusted? The dialog appears to self limit its width (wrapping lines), and 4-5@120 would allow fewer double line titles to be shown. It is less likely someone would read more than 4-5 titles when mass deleting, and showing more of long titles might be better.


elelay:
looks good :+1:

auouymous:
I think it looks better without the blank lines between each title and doesn't drown out the message.

@elelay Does it still look good to you?

elelay:
yes, it's clearer now :+1:

elelay:
Cherry-picked commits because merge seemed overly complex

elelay:
Thanks

Issue #922, opened 2020-12-12
 $DISPLAY is not set.


theorbitofmercury:
I cloned the git latest. However when I try to run I got  "DISPLAY is not set". (RPi3 with latest Raspberry OS & VNC). Couldn't find anything about this on Google re gpodder. I followed https://gpodder.github.io/docs/run-from-git.html. instructions. I wasn't sure how to deal with the dependencies:
Gtk3
Python 3 (already there)
PyGObject 3
python bindings: gir1.2-gtk-3.0 python3-cairo python3-gi-cairo on Debian
Git (already there)
do I have to install these or are they taken care  of in the clone?

deleted the clone and installed the older version on the Debian repo. Same result
Thanks

auouymous:
The Gtk version of gpodder (`gpodder` binary) requires an X11 server or wayland, and the DISPLAY environment variable indicates the server and screen. DISPLAY is automatically set by terminal emulators and ssh with X forwarding. `gpo`, the command line version, lacks features but doesn't require Gtk or X.

theorbitofmercury:
I am using Raspberry OS desktop version(buster). So I think I am using an X server?

auouymous:
You can try running `export DISPLAY=":0.0"` before running gpodder.

athohz:
I use gpodder under wayland, without xwayland. I think in this case DISPLAY should not be tested (in bin/gpodder, line 151).

`$ gpodder`
`[__main__] ERROR: Cannot start gPodder: $DISPLAY is not set.`

`$ DISPLAY="any-random-string" gpodder ` 
and gpodder starts just fine.

tpikonen:
There are probably people using wayland without any freedesktop.org software, so XDG_SESSION_TYPE is maybe not foolproof. The WAYLAND_DISPLAY variable should be always set under wayland though.

And what about the macOS version? The test above runs also when gpodder.ui.osx is true, so apparently macOS has DISPLAY set.

elelay:
on macos the check is [removed](https://github.com/gpodder/gpodder-osx-bundle/blob/master/modulesets/patches/gpodder_dont_check_display.patch). It's easier than guessing if it's on macOS with X11 backend or Quartz backend.

auouymous:
The DISPLAY check is needed because gpodder segfaults without a DISPLAY. Adding a check for WAYLAND_DISPLAY will work because wayland requires one of them.

```
[...]

PR #984, opened 2021-03-19
Require either DISPLAY or WAYLAND_DISPLAY being set

tpikonen:
Allows running gPodder under pure Wayland.

Doesn't check for these variables under macOS, which will hopefully allow a smaller delta in the OSX version.

Fixes at least some aspects of #922.

### DIFF
--- a/bin/gpodder
+++ b/bin/gpodder
@@ -124,6 +124,12 @@ def main():
     from gpodder import log
     log.setup(options.verbose, options.quiet)
 
+    if (not (gpodder.ui.win32 or gpodder.ui.osx) and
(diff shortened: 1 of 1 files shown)


auouymous:
LGTM, but it would be better if the change was moved up above the dbus section.

What part of #922 doesn't it fix?

tpikonen:
Ah, I missed the part about dbus in #922. Moved.

The original reporter in #922 was trying to run gPodder under VNC. I'm not sure if this fixes that particular situation.

auouymous:
Merged, thanks.

elelay:
Thanks @tpikonen and @auouymous :+1:

PR #990, opened 2021-03-23
Remove skip_locked to avoid accidental deletion of episodes

### LINKED ISSUE #989: Lock doesn't prevent deletion of single episodes
Locking episodes prevents accidental deletion, for downloaded or undownloaded episodes. It works fine when multiple episodes are selected, but the code below clearly bypasses the lock when only one episode is deleted. The lock even remains on the episode after deleting it. I don't see a reason to bypass the lock, because if someone really wants to delete the episode, they can take the time to unlock it first.

Is there a reason for the current useless behavior? I propose the skip_locked parameter be removed entirely.

```
    def on_btnDownloadedDelete_clicked(self, widget, *args):
        episodes = self.get_selected_episodes()
        if len(episodes) == 1:
            self.delete_episode_list(episodes, skip_locked=False)
        else:
            self.delete_episode_list(episodes)
```

auouymous:
Fixes #989

### DIFF
--- a/bin/gpo
+++ b/bin/gpo
@@ -828,21 +828,20 @@ class gPodderCli(object):
             line = input(msg)
             return not line or (line.lower() == _('yes'))
 
-        def _delete_episode_list(episodes, confirm=True, skip_locked=True, callback=None):
+        def _delete_episode_list(episodes, confirm=True, callback=None):
             if not episodes:
                 return False
 
-            if skip_locked:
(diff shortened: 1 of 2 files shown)


elelay:
Merged, thanks!

kb6ojs:
> Fixes #989

I didn't even know you COULD lock an episode!  Thanks for the fix and for calling my attention to an undiscovered (on my part) feature of the program!

// Steve //

PR #991, opened 2021-03-24
Properly toggle deleted episodes to new.

auouymous:
The Episodes menu would sometimes require two clicks to mark a deleted episode as new. And the context menu would never mark it as new because the methods it called don't support that. The context menu would also set the checkbox to an incorrect state, making it appear as if a deleted episode was new.


elelay:
Merged, thanks

PR #994, opened 2021-03-27
Bump deps for release.

auouymous:
@elelay Do the tools/win_installer/bootstrap.sh changes look good?


elelay:
the addition of modules to pip install in bootstrap.sh shouldn't be necessary.
Let me try another branch with it removed entirely

PR #995, opened 2021-03-27
WIP: try remove pip install in bootstrap

elelay:


auouymous:
The Zstd fix worked but fails because https://github.com/enthought/pywin32-ctypes no longer exists. https://github.com/jaraco/pywin32-ctypes has the f27d6a0 commit and the build completes with it, should we use that repo?

elelay:
Yes, that's fine. Thanks! It builds then...

PR #1005, opened 2021-04-10
Switch from filemagic to python-magic for github releases.

auouymous:
https://github.com/aliles/filemagic was last updated in 2014, has 5 contributors and is used by 265. https://github.com/ahupp/python-magic was updated this year, has 42 contributors and is used by 12522.

### DIFF
--- a/tools/github_release.py
+++ b/tools/github_release.py
@@ -8,7 +8,7 @@ import os
 import re
 import sys
 
-import magic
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks!
Another would have been a mini mapping from extensions we push to mime types but this way it's ready for a new file.

PR #1015, opened 2021-04-23
Fix connection checking and add checkbox to disable it

### LINKED ISSUE #1014: "No network connection" when checking for new episodes on 3.10.19
Also doesn't download episodes. This was working up until about a week ago.

I tried running the `linux_get_active_interfaces()` method in `util.py`, and it returns an empty generator. If I run the command it execs myself, `ip addr show scope global up`, I get:
```
4: wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 58:96:1d:5b:47:08 brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.5/24 scope global noprefixroute wlan0
       valid_lft forever preferred_lft forever
```
My `ip` command is from iproute2 5.11.0

auouymous:
The new `Check connection (if supported)` checkbox is in the updating tab. Thoughts on a better label for it?

Fixes #1014

### DIFF
--- a/src/gpodder/config.py
+++ b/src/gpodder/config.py
@@ -91,6 +91,8 @@ defaults = {
         'retries': 3,  # number of retries when downloads time out
     },
 
+    'check_connection': True,
+
     # Software updates from gpodder.org
     'software_update': {
         'check_on_startup': True,  # check for updates on start
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -563,6 +563,26 @@
                     <property name="position">5</property>
                   </packing>
                 </child>
+                <child>
+                  <object class="GtkSeparator" id="hseparator_updating3">
+                    <property name="orientation">horizontal</property>
+                    <property name="visible">True</property>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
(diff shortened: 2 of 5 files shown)


src/gpodder/config.py
```diff
@@ -228,6 +230,7 @@
     'auto_remove_unfinished_episodes': 'auto.cleanup.unfinished',
     'auto_remove_unplayed_episodes': 'auto.cleanup.unplayed',
     'max_episodes_per_feed': 'limit.episodes',
+    'check_connection': 'check_connection',
```
elelay, line 233: is this really needed?

elelay (changes requested):
legacy support shouldn't be necessary.
Except from that, LGTM.
Thanks,

elelay:
"Check connection before refresh (if supported)" might be clearer

src/gpodder/config.py
auouymous: It was needed for a checkbox I added a week or two ago but apparently removing it seems to work for this one. Why do the other checkboxes need to be there?

elelay: It's to map the old short 'max_episodes_per_feed' key to the 'episodes' key in the new 'limit' object.

elelay:
Merged, thanks!

PR #1018, opened 2021-04-25
build gPodder bundle from tools/mac-osx and downloaded generic bundle

elelay:
gpodder-osx-bundle now builds a more generic bundle with native deps and a working pip.


auouymous:
@elelay Oops, I think this PR broke linter. I fixed it with 9f02f7b925b61db81ef05c64f95b0aa81c610510, but maybe you should enable a pre-commit hook to run `make lint` since TravisCI no longer works.

elelay:
Oops indeed! Thanks for the fix. Yes, a pre-commit or pre-push might help

PR #1019, opened 2021-04-28
Do not show 'All episodes' on podcast search

tpikonen:
Currently when searching podcasts, the 'All episodes' channel proxy and sections are always displayed on the results. This is confusing, since they do no match the search string.

This patch extends the keyword matching to PodcastChannelProxy titles, so they are only shown in search results when the title matches.

EDIT: This patch now just removes the 'All episodes' item from the podcast list during search.

### DIFF
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -590,7 +590,7 @@ class PodcastListModel(Gtk.ListStore):
         # If searching is active, set visibility based on search text
         if self._search_term is not None:
             if isinstance(channel, PodcastChannelProxy):
-                return True
+                return not channel.ALL_EPISODES_PROXY
             key = self._search_term.lower()
             columns = (model.get_value(iter, c) for c in self.SEARCH_COLUMNS)
             return any((key in c.lower() for c in columns if c is not None))


elelay:
ok, but are podcasts matching the query separated by the sections?

tpikonen:
No, the sections are not displayed if their title does not match the search string.

auouymous:
Hiding `All episodes` and empty sections would be nice but I like seeing section headers even if they don't match the key. The channels also look confusing without sections because some are in alphabetical order and others aren't. Maybe it is possible to highlight a section that matches for anyone wanting to search for them?

tpikonen:
Actually just hiding the 'All episodes' item during search would be fine for me, as then the first matching podcast is selected on the list. I'll change the PR accordingly.

auouymous:
LGTM. But all-episodes remains selected for me. If a channel was selected before searching, the next channel or section after it does become selected if it gets hidden.

What are your thoughts on resetting the channel/episode views if the search key is cleared with backspace? There is an option to always show find entries (which I don't use) and users must manually click the clear button or press Esc to reset the view. I'm not sure if that is a feature but it is confusing to me and this patch keeps all-episodes hidden when field is cleared with backspace. The following in both _filter_visible_func methods would reset channel and episode lists when their entry is cleared with backspace.

```
-        if self._search_term is not None:
+        if self._search_term is not None and self._search_term != '':
```

auouymous:
The following resets channel/episode views when cleared with backspace, hides all-episodes when searching, and hides sections that don't have matched channels. I'm not sure why, but it seems to change selection from all-episodes to a channel or section. This does not fix the pre-existing #1023 found during testing.

```
diff --git a/src/gpodder/gtkui/model.py b/src/gpodder/gtkui/model.py
index dd78b248..0b15a81b 100644
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -216,7 +216,7 @@ class EpisodeListModel(Gtk.ListStore):
 
     def _filter_visible_func(self, model, iter, misc):
         # If searching is active, set visibility based on search text
-        if self._search_term is not None:
+        if self._search_term is not None and self._search_term != '':
             episode = model.get_value(iter, self.C_EPISODE)
             if episode is None:
                 return False
@@ -588,10 +588,12 @@ class PodcastListModel(Gtk.ListStore):
         channel = model.get_value(iter, self.C_CHANNEL)
 
         # If searching is active, set visibility based on search text
-        if self._search_term is not None:
-            if isinstance(channel, PodcastChannelProxy):
-                return True
+        if self._search_term is not None and self._search_term != '':
             key = self._search_term.lower()
+            if isinstance(channel, PodcastChannelProxy):
+                if channel.ALL_EPISODES_PROXY:
+                    return False
+                return any(key in getattr(ch, c).lower() for c in PodcastListModel.SEARCH_ATTRS for ch in channel.channels)
             columns = (model.get_value(iter, c) for c in self.SEARCH_COLUMNS)
             return any((key in c.lower() for c in columns if c is not None))
 
```

tpikonen:
The patch above works fine and is a better solution to this. @auouymous, do you want to submit a separate PR, so I can close this one?

elelay:
Replaced by #1030.
Thanks,

PR #1021, opened 2021-05-02
Disable circleci for adaptive branch.

auouymous:
There is no reason to generate mac builds until gpodder-osx-bundle gets libhandy, if ever.

@elelay Does this look like it would work?


elelay:
Let's try ;-)

PR #1028, opened 2021-05-07
Create config file with group/world permissions revoked

juergenhoetzel:
The config file contains sensible information (user/password).

Also remove unused import.

### DIFF
--- a/src/gpodder/config.py
+++ b/src/gpodder/config.py
@@ -27,7 +27,6 @@
 import atexit
 import logging
 import os
-import shutil
 import time
 
 import gpodder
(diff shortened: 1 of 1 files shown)


auouymous:
I run my entire system with 077 and fully agree with this change. The patch looks good and appears to work fine, but will it cause any issues on Windows?

juergenhoetzel:
> I run my entire system with 077 and fully agree with this change. The patch looks good and appears to work fine, but will it cause any issues on Windows?

Posix `umask` is also awailable on Windows: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-umask?view=msvc-160

elelay:
Thanks for the patch!

Tested on windows:
 - it doesn't crash
 - it doesn't change permissions.
 
withumask

I can't find information on python's os.umask, but testing results and this commit https://git.lclutz.de/ghstars/certbot-certbot/commit/50fa04ba0cb739eb0eafec2286cb08d9caae1b16 leads me to believe it's stubbed out on windows...

elelay:
More direct test:
```
C:\Users\IEUser>C:\gpodder-3.10.17-portable\data\bin\python
Python 3.8.9 (default, Apr  3 2021, 09:29:25)  [GCC 10.2.0 32 bit] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.umask(0)
0
>>> os.umask(0o077)
0
>>> os.umask(0)
0
```

Also cpython tests don't assert modified persmissions on windows https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Lib/test/test_os.py#L1582

@juergenhoetzel please add a simple comment that it's a noop on windows and let's merge, to improve it on other platforms.

juergenhoetzel:
> @juergenhoetzel please add a simple comment that it's a noop on windows and let's merge, to improve it on other platforms.

Thank you for your research. I just amended the commit and added your suggestion.

elelay:
Merged, thanks!

PR #1031, opened 2021-05-10
Allow a narrower podcast directory window

tpikonen:
I'm trying to reduce the delta between adaptive and master. This PR moves the 'Select All' and 'Select None' buttons in gPodderPodcastDirectory window to a separate box (or row), so that the window can be made narrower. After this patch the only change needed to gPodderPodcastDirectory in adaptive is to replace GtkPaned with HdyFlap.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpodcastdirectory.ui
+++ b/share/gpodder/ui/gtk/gpodderpodcastdirectory.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.38.2 -->
 <interface>
-  <requires lib="gtk+" version="3.0"/>
+  <requires lib="gtk+" version="3.12"/>
   <object class="GtkDialog" id="gPodderPodcastDirectory">
     <property name="can-focus">False</property>
     <property name="border-width">6</property>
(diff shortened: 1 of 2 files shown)


elelay:
Trouble is it's less optimized for wider screens (like a phone in landscape mode or a tiled layout).
How can we reconcile both?

tpikonen:
[GtkFlowBox](https://developer.gnome.org/gtk4/stable/GtkFlowBox.html) would neatly solve this problem. It just happens to be a GTK4 widget.

Incidentally, what are your thoughts on porting gPodder to GTK4?

elelay:
Nice: just what I looked for in gtk3: a reflow would adapt to both landscape-ish and portrait-ish displays!
Regarding Gtk4, I'm eager to fix compatibility, but in no rush to push it as a hard requirement by using new components

tpikonen:
Bah, I was wrong: GtkFlowBox has been in Gtk since version 3.12. The PR is now updated to use it. The toggles in podcastdirectory are also made wider to be a little more touch friendly.

auouymous:
I don't use that dialog but I think this looks so much better. The only problem I see is the increased gap below buttons

tpikonen:
The gap was the unused action area in GtkDialog. That cannot be removed from GtkDialog, so I changed the window to a normal GtkWindow.

elelay:
looks good, except the *Subscriptions* > *delete podcasts* dialog, where for some reason *Delete* and *Cancel* stay on the 2nd line...

tpikonen:
'Delete podcasts' uses the episode selector dialog. I don't mind converting it also though :)

auouymous:
LGTM.

The episode selector dialog might be more difficult to get on a single line. The select buttons have a selection count label on the line, and there is a `Mark as old` button on the second line. Also, the selection count label is directly above the download button, so you can quickly see how many will be downloaded. I've never used it, but it would be nice if the delete podcast dialog also showed selected count to indicate how many would be deleted.

elelay:
> 'Delete podcasts' uses the episode selector dialog. I don't mind converting it also though :)

ha ha indeed it's the episode selector dialog.

elelay:
Merged, thanks!
I'm open on @auouymous suggestions, but on another PR.

PR #1034, opened 2021-05-13
Fix "GtkSelectionData has no 'data'" exception in drag_data_received().

auouymous:
This happens when dragging an image URL to the cover art in channel dialog.

I don't know if this is the correct fix, but it allows an image to be dragged from a web browser.


elelay:
Merged, thanks!

PR #1035, opened 2021-05-14
Make refreshing channel image more discoverable and easier to do.

### LINKED ISSUE #1033: [Feature idea] Option to automatically or manually (re)download/refresh podcast pictures; globally and per podcast
**Current state**
When adding a podcast, gPodder downloads the podcast's cover picture. This is great.

**Problem**
However, podcast pictures/covers are changed from time to time by the podcast providers, especially with YouTube channel's pictures this happens quite often. But gPodder does not refresh/re-download those cover pictures periodically. So those podcast pictures/covers must be refreshed manually _per podcast_ which is quite time consuming and tedious.

**Feature idea**
- Please add an on/off toggle option in gPodder preferences that makes gPodder automatically re-download/refresh the podcast pictures/covers when searching for new episodes.
- In addition to that, there should also be an entry in the menu "Podcasts" in the menu bar called "Download cover pictures" so that users can manually initiate the cover refresh of _all podcasts globally_.

auouymous:
Should the `Refresh image` menu label be changed? "cover" and "cover art" might confuse users. "icon"? "thumbnail"? Or stick with "image" since the RSS &lt;image&gt; tag declares it?

Fixes #1033.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -1715,6 +1715,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
             item.connect('activate', self.on_channel_toggle_lock_activate)
             menu.append(item)
 
+            item = Gtk.ImageMenuItem(_('Refresh image'))
+            item.connect('activate', self.on_itemRefreshCover_activate)
+            menu.append(item)
+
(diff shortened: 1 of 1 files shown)


elelay:
+1 for image

PR #1040, opened 2021-05-18
Show section if any of its channels have an update error.

auouymous:
A channel with an update error might not any have visible episodes. When its section has no other visible channels, the section must still be displayed to prevent the channel with an error from appearing to be in a different section.


elelay:
Good point! Thanks

PR #1058, opened 2021-06-02
Update Norwegian Nynorsk translation for next release

huftis:
Here’s a translation update for the next release. (But I couldn’t find the new ‘Open’ string mentioned in https://github.com/gpodder/gpodder/issues/518#issuecomment-851989454.)


auouymous:
Thank you. I thought "Open" was new, but it just moved to a new location in the file.

PR #1059, opened 2021-06-03
Return an audio mime-type from Youtube-DL when given a video extension.

auouymous:
This uses audio icon and player when audio formats are downloaded with youtube-dl. Both youtube-dl and built-in download a weba file, but youtube-dl uses its `ext` field (which contains webm) and produces the wrong mime-type.

This also guarantees all supported youtube-dl sites will produce an audio mime-type if youtube-dl has the wrong `ext` value.


elelay:
Merged, thanks!

Issue #1026, opened 2021-05-04
Cannot use sock5 proxy (Tor connection)


realSConway:
Hello experts,

I'm trying to force gPodder to tunnel downloads through my proxy server.

Running gPodder flatpack version: 3.10.19

My env variable is set with:
```
$ echo $all_proxy
socks5://127.0.0.1:9050
```

When attempting to download a podcast error:
```
Traceback (most recent call last):
  File "/app/lib/python3.8/site-packages/gpodder/download.py", line 790, in run
    headers, real_url = downloader.retrieve_resume(self.tempname, self.status_updated)
  File "/app/lib/python3.8/site-packages/gpodder/download.py", line 340, in retrieve_resume
    headers, real_url = downloader.retrieve_resume(url,
  File "/app/lib/python3.8/site-packages/gpodder/download.py", line 268, in retrieve_resume
    with session.get(url,
  File "/app/lib/python3.8/site-packages/requests/sessions.py", line 555, in get
    return self.request('GET', url, **kwargs)
  File "/app/lib/python3.8/site-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/app/lib/python3.8/site-packages/requests/sessions.py", line 655, in send
    r = adapter.send(request, **kwargs)
  File "/app/lib/python3.8/site-packages/requests/adapters.py", line 412, in send
    conn = self.get_connection(request.url, proxies)
  File "/app/lib/python3.8/site-packages/requests/adapters.py", line 309, in get_connection
    proxy_manager = self.proxy_manager_for(proxy)
  File "/app/lib/python3.8/site-packages/requests/adapters.py", line 182, in proxy_manager_for
    manager = self.proxy_manager[proxy] = SOCKSProxyManager(
  File "/app/lib/python3.8/site-packages/requests/adapters.py", line 47, in SOCKSProxyManager
    raise InvalidSchema("Missing dependencies for SOCKS support.")
requests.exceptions.InvalidSchema: Missing dependencies for SOCKS support.
```

socks is already installed via pip
```
$ pip install requests[socks]
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: requests[socks] in /usr/lib/python3.8/site-packages (2.25.1)
Requirement already satisfied: certifi>=2017.4.17 in /usr/lib/python3.8/site-packages (from requests[socks]) (2020.12.5)
[...]

PR #1060, opened 2021-06-03
Add tools/requirements.txt

tpikonen:
There seems to be no machine readable dependency list for the Python packages used by gPodder. This PR adds a pip-compatible [requirements](https://pip.pypa.io/en/stable/cli/pip_install/#requirements-file-format) file, which can be used to get the correct dependencies. I hope I did not miss anything.

I needed this when fixing the dependencies in the flatpak [package](https://github.com/flathub/org.gpodder.gpodder). There's a tool called [flatpak-pip-generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/pip) which converts the requirements file to the format understood by flatpak-builder.

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -1,0 +1,13 @@
+# PyPI / pip requirements for Linux
+# For the benefit of e.g. flatpak-pip-generator.
+#
+mygpoclient==1.8
+podcastparser==0.6.6
+requests[socks]==2.25.1
+urllib3==1.26.4
+html5lib==1.1
+mutagen==1.45.1
+dbus-python
+youtube_dl
+# eyed3 is optional and pulls in a lot of dependencies, so disable by default
+# eyed3


auouymous:
I know nothing about this but should it be in the root directory like other projects?

Should it have the following dependencies?

```
webencodings==0.5.1
certifi==2020.11.8
chardet==4.0.0
idna==3.1
PySocks==1.7.1
```

tpikonen:
I don't mind putting the requirements to the root dir, but since gPodder is not in PyPI, I don't think it needs to be there. The other installer related files are all under tools/, including tools/mac-osx/release_on_mac.sh and tools/win_installer/_base.sh which contain a variant of this list.

The packages you list are all pulled in by pip as dependencies for requests or html5lib (webencodings). PySocks was missing, but is now pulled in by 'requests[socks]'. There's no direct imports of these in the code, so letting pip handle the versioning should be fine.

tpikonen:
BTW, there should perhaps also be a 'requirements-text.txt' file, which would have the testing dependencies listed in the README.

elelay:
Merged, thanks. That would solve the Tor user issue #1026

PR #1065, opened 2021-06-08
Prevent soundcloud from downloading existing episodes each update.

### LINKED ISSUE #1064: recently gpodder started to download the same soundcloud episodes over and over again
This seems to just affect soundcloud podcasts. 
I also tried setting them manually as "old", but this didn't help either. 
I am pretty sure this only started in the recent few weeks. Maybe Since May. 

I am on gPodder 3.10.19 (2021-04-15) using the Fedora package.

auouymous:
GUIDs are stored in the database as strings, but soundcloud creates them as integers. Python no longer succeeds when comparing these integers to the string keys from the existing database entries. This produces new episodes on every update which then collide with the existing GUIDs.

Fixes #1064.

### DIFF
--- a/src/gpodder/plugins/soundcloud.py
+++ b/src/gpodder/plugins/soundcloud.py
@@ -153,7 +153,7 @@ class SoundcloudUser(object):
                     'url': url,
                     'file_size': int(filesize),
                     'mime_type': filetype,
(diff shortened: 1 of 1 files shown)


christofdamian:
It does seem to have fixed it. On first run it downloaded every available podcast, but the second run just got the new ones.

PR #1070, opened 2021-06-15
Add Slovak translation

neodyne:
Written from scratch, hopefully there are only a few typos :)


elelay:
Merged, thanks!
If you want to continue maintaining the translation, you can subscribe to #518 where we post a notification before each release.
Cheers,

PR #1071, opened 2021-06-15
Narrow channel dialog

tpikonen:
A rewrite of the channel dialog UI, allowing a narrower width, and a 'modern' look.

I'm on the fence on whether the dialog should have 'Cancel' and 'Save' buttons instead of just 'Close'. Probably, but I did not add them yet.

### DIFF
--- a/share/gpodder/extensions/filter.py
+++ b/share/gpodder/extensions/filter.py
@@ -135,6 +135,13 @@ class gPodderExtension:
         box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=10)
         box.set_border_width(10)
 
+        # note about Cancel
+        note = Gtk.Label(use_markup=True, wrap=True, label=_(
+            '<b>Note:</b> The Cancel button does <b>not</b> return the '
+            'filter settings to the values they had before. '
+            'The changes are saved immediately after they are made.'))
+        box.add(note)
+
         # block widgets
         self.block_widget = BlockExceptFrame(value=block,
                                              enable_re=self.key('block_re') is not False,
@@ -163,9 +170,14 @@ class gPodderExtension:
         box.add(label)
 
         # re-filter
-        button = Gtk.Button(_('Filter episodes now (undoes any episodes you marked as old)'))
+        separator = Gtk.HSeparator()
+        box.add(separator)
+        button = Gtk.Button(_('Filter episodes now'))
         button.connect('clicked', self.refilter_podcast)
         box.add(button)
+
+        label2 = Gtk.Label(_('Undoes any episodes you marked as old.'))
+        box.add(label2)
 
         box.show_all()
         return box
--- a/share/gpodder/ui/gtk/gpodderchannel.ui
+++ b/share/gpodder/ui/gtk/gpodderchannel.ui
@@ -1,39 +1,39 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.38.2 -->
 <interface>
-  <!-- interface-naming-policy toplevel-contextual -->
+  <requires lib="gtk+" version="3.16"/>
   <object class="GtkDialog" id="gPodderChannel">
-    <property name="can_focus">False</property>
-    <property name="border_width">10</property>
+    <property name="can-focus">False</property>
+    <property name="border-width">10</property>
     <property name="title" translatable="yes">gPodder Podcast Editor</property>
     <property name="modal">True</property>
-    <property name="transient-for">parent_widget</property>
-    <property name="window_position">center-on-parent</property>
-    <property name="default_width">500</property>
-    <property name="default_height">400</property>
-    <property name="type_hint">dialog</property>
+    <property name="window-position">center-on-parent</property>
+    <property name="default-width">500</property>
+    <property name="default-height">400</property>
+    <property name="type-hint">dialog</property>
     <signal name="destroy" handler="on_gPodderChannel_destroy" swapped="no"/>
     <child internal-child="vbox">
       <object class="GtkBox" id="vboxChannelEditorMain">
         <property name="visible">True</property>
-        <property name="can_focus">False</property>
-        <property name="spacing">5</property>
+        <property name="can-focus">False</property>
         <property name="orientation">vertical</property>
+        <property name="spacing">5</property>
         <child internal-child="action_area">
-          <object class="GtkHButtonBox" id="hboxButtons">
+          <object class="GtkButtonBox" id="hboxButtons">
             <property name="visible">True</property>
-            <property name="can_focus">False</property>
-            <property name="layout_style">end</property>
+            <property name="can-focus">False</property>
+            <property name="layout-style">end</property>
             <child>
               <object class="GtkButton" id="btnOK">
                 <property name="label">gtk-close</property>
+                <property name="use-action-appearance">False</property>
                 <property name="visible">True</property>
-                <property name="can_focus">True</property>
-                <property name="has_focus">True</property>
-                <property name="can_default">True</property>
-                <property name="has_default">True</property>
-                <property name="receives_default">False</property>
-                <property name="use_action_appearance">False</property>
-                <property name="use_stock">True</property>
+                <property name="can-focus">True</property>
+                <property name="has-focus">True</property>
+                <property name="can-default">True</property>
+                <property name="has-default">True</property>
+                <property name="receives-default">False</property>
+                <property name="use-stock">True</property>
                 <signal name="clicked" handler="on_btnOK_clicked" swapped="no"/>
               </object>
               <packing>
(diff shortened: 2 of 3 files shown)


auouymous:
It looks good...

I mostly use the channel dialog to set sections and filters. Now I am forced to scroll to set the section. Some podcasts have really long descriptions, is it possible to clip the description to a couple lines and add a show more button to see the full description? The update error is appended to description, and if clipped, it might need its own fully visible label in the dialog (with an error icon).

Is it possible to add non-scrolling notebook-like buttons along the top that jump to description, settings, advanced settings or extensions? My trackpad lacks scrolling and the dialog doesn't scroll with arrow keys, forcing me to reach for the scroll keys on edge of keyboard or the mouse. A button that jumps to settings would be an extra step but not that big of a deal, and the description wouldn't need clipping.

Please remove the bogus tab stops on the container widgets, they require two tab presses for most widgets.

It also broke the filter extension (and maybe others) because you removed the notebook they add pages to. Do you propose extensions append to the end of the scroll forever window?

Close button has no margin around it.

auouymous:
Please don't override the GtkNotebook theme, maybe it looks good on your system with your theme but it is unusable on mine.

The Cancel and Ok buttons still don't have a margin.

I like the Cancel button but it doesn't work for the filter extension, and I doubt it is possible. Should the filter extension have a label informing users that changes are final?

Should the section combo box and new section button have a gap between each other?

The username and password labels have no gap between the input box.

The upper settings look good but the username and password don't. The box uses the same background color as the inputs, and they stand out and look nicer on the window's background color. And hovering or clicking the inputs changes the row inside the box to the highlight background color. This is fine for switches and combo boxes but makes entering text harder. I realize the row can be clicked to focus the input box, and that might make touch easier, but it worsens the usability for others.

Do you think the re-filter button should have a horizontal separator above it so the new label better associates with the button?

auouymous:
The current dialog centers on parent, but this patch does not, even though it should. Do you know what could have broken that?

The channel dialog should also save its size and position. It went from a landscape window to a portrait window which makes the description a little harder to read, requires scrolling to see the URL and reduces the visible content in filter input boxes. I would set my window manager to remember the size but the dialog has a different title for each channel. Using a fixed dialog title would work, except for users on Windows.

tpikonen:
Thanks for the comments @auouymous, I think I addressed all of them in the latest push, except saving the dialog size and position. I think that's a separate PR, if we decide to do that. The dialog is now always called 'Channel Editor', if that helps.

I run the default GTK theme, and the white ListBoxRows looked pretty bad on white Notebook background. I removed the frame around the rows, which makes the rows blend into the background when not highlighted and looks better.

auouymous:
LGTM, @elelay do you approve?

I think we should save and restore the window size and position. There are users who depend on it, and I'm sure they'd appreciate it now that this dialog is narrower.

elelay:
Looks good to me except the description is not selectable. This must be fixed (I'd use a text area for that).

auouymous:
Ah missed that, the update error was added to the description's textarea so it would be copyable. A selectable property like the two URLs have should be good enough for description.

auouymous:
Another issue is the website URL that is no longer selectable.

And the old username/password entries were aligned to each other, is that not possible with the new layout?

tpikonen:
Here you are. Let me know if there's anything still to be fixed. If not, I'll squash this to something more reasonable.

elelay:
looks good to me, thanks!

tpikonen:
Squashed and rebased to master.

auouymous:
LGTM. Does it save and restore the dialog position? If done in another PR, it would be nice if that PR was ready before this is merged.

elelay:
See 834106e67c6ba379c6b0b1b3fb666c037c6e3888
it's good to save/restore positions for windows and macOS

auouymous:
Does the website link work for you? The "Open Link" menu item does nothing and clicking the link places a cursor on it and changes its color to from blue to white. Double-clicking selects text within the URL and then right-clicking opens a "text" context menu to select all or copy.

auouymous:
The About dialog links also don't work but #1095 fixes them and could be applied here as well.

auouymous:
```patch
From ab2219cb95fd2b9aa9545ab1cb77b59db16b0b31 Mon Sep 17 00:00:00 2001
From: auouymous <au@qzx.com>
Date: Sat, 17 Jul 2021 21:50:20 -0600
Subject: [PATCH] Save and restore channel dialog state.

---
 src/gpodder/config.py                | 5 +++++
 src/gpodder/gtkui/desktop/channel.py | 2 ++
 2 files changed, 7 insertions(+)

diff --git a/src/gpodder/config.py b/src/gpodder/config.py
index 45381bc0..f92189c8 100644
--- a/src/gpodder/config.py
+++ b/src/gpodder/config.py
@@ -130,6 +130,11 @@ defaults = {
                     'height': -1,
                     'x': -1, 'y': -1, 'maximized': False,
                 },
+                'channel_editor': {
+                    'width': -1,
+                    'height': -1,
+                    'x': -1, 'y': -1, 'maximized': False,
+                },
                 'episode_selector': {
                     'width': 600,
                     'height': 400,
diff --git a/src/gpodder/gtkui/desktop/channel.py b/src/gpodder/gtkui/desktop/channel.py
index ad252857..ea2d1238 100644
--- a/src/gpodder/gtkui/desktop/channel.py
+++ b/src/gpodder/gtkui/desktop/channel.py
@@ -98,6 +98,8 @@ class gPodderChannel(BuilderWidget):
         self.imgCoverEventBox.connect('button-press-event',
                 self.on_cover_popup_menu)
 
+        self._config.connect_gtk_window(self.gPodderChannel, 'channel_editor', True)
+
         gpodder.user_extensions.on_ui_object_available('channel-gtk', self)
 
         result = gpodder.user_extensions.on_channel_settings(self.channel)
-- 
2.31.1

```

auouymous:
These two patches should resolve all issues and allow this to be merged. @elelay Should we do it now or wait until after release?

```patch
From a0888b0eeb78789a6e6f9942824726c21abd266a Mon Sep 17 00:00:00 2001
From: auouymous <au@qzx.com>
Date: Sat, 17 Jul 2021 22:21:16 -0600
Subject: [PATCH] Fix link in channel dialog by connecting signal and opening
 website.

---
 src/gpodder/gtkui/desktop/channel.py | 1 +
 1 file changed, 1 insertion(+)

diff --git a/src/gpodder/gtkui/desktop/channel.py b/src/gpodder/gtkui/desktop/channel.py
index 9e1fb388..1183139a 100644
--- a/src/gpodder/gtkui/desktop/channel.py
+++ b/src/gpodder/gtkui/desktop/channel.py
@@ -67,6 +67,7 @@ class gPodderChannel(BuilderWidget):
         self.website_label.set_markup('<a href="{}">{}</a>'.format(
             self.channel.link, self.channel.link)
             if self.channel.link else '')
+        self.website_label.connect('activate-link', lambda label, url: util.open_website(url))
 
         if self.channel.auth_username:
             self.FeedUsername.set_text(self.channel.auth_username)
-- 
2.31.1

```

elelay:
@auouymous links are working in podcast settings and in about dialog without your fix.
You use activate-link to override the default, whatever it does...
https://developer.gnome.org/gtk3/stable/GtkLabel.html#id-1.3.8.2.10.9

On my machine links are opened in the current firefox window. I have the `BROWSER=firefox` environment variable set.

Feel free to commit the *Save and restore channel dialog state.* on this branche and merge before release.

auouymous:
@elelay Do the about links work on Mac and Windows for you? I set BROWSER=w3m with patch disabled and the links still don't work in Linux.

elelay:
They work on macOS.
on Linux(Archlinux) I have gtk 3.24.30-1

elelay:
Seems to call [gtk_show_uri_on_window](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkshow.c#L159) by default

elelay:
You might want to try `http://lazka.github.io/pgi-docs/#Gtk-3.0/functions.html#Gtk.show_uri_on_window`:
From a python shell
```
>>> import gi
>>> from gi.repository import Gtk
<stdin>:1: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.
>>> Gtk.show_uri_on_window(None, 'https://gpodder.github.io', 0)
True
```

(and a tab opens)

auouymous:
I have 3.24.29 on Linux and the Windows artifact I installed has 3.24.30, and I'll be testing the patched artifact tomorrow to verify it fixed the problem there. Did you set BROWSER on Mac? Does it work on Linux without BROWSER? We shouldn't require environment variables for features to work.

From that https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkshow.c#L159:

> The uri must be of a form understood by GIO (i.e. you  * need to install gvfs to get support for uri schemes such as http://  * or ftp://, as only local files are handled by GIO itself)

I don't have gvfs installed and neither does the Windows build.

auouymous:
```
>>> Gtk.show_uri_on_window(None, 'https://gpodder.github.io', 0)
  File "<stdin>", line 1, in <module>
gi.repository.GLib.Error: g-io-error-quark: Operation not supported (15)
```

elelay:
Ah, that's it then: gvfs is indeed installed on my machine. It's not installed on macOS but links work.
I can't check on windows right now.
So go for your patches :+1: to podcast and about dialogs. Maybe search for a href markup elsewhere?

auouymous:
> Maybe search for a href markup elsewhere?

What?

elelay:
> > Maybe search for a href markup elsewhere?

I meant `git grep href= src` to look for any other use of hyperlinks in label markup. But you already got them all.

PR #1073, opened 2021-06-18
Fix support for automatically accepting GDPR youtube consents

18928172992817182:
Fixes the # TODO: watch URL does not work in europe due to GDPR cookie consent added in https://github.com/gpodder/gpodder/commit/a2fc81ee829a883b34501ed75d2bfd0ff7fdb3dd

### DIFF
--- a/src/gpodder/youtube.py
+++ b/src/gpodder/youtube.py
@@ -203,7 +203,6 @@ def get_real_download_url(url, allow_partial, preferred_fmt_ids=None):
         if not r.ok:
             logger.warning('Youtube get_video_info: %d %s' % (r.status_code, r.reason))
 
-            # TODO: watch URL does not work in europe due to GDPR cookie consent
             url = 'https://www.youtube.com/watch?bpctr=9999999999&has_verified=1&v=' + vid
             r = requests.get(url)
             if not r.ok:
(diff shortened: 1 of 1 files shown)


auouymous:
Thank you! The number of videos returning 404 on old endpoint increased from one a day to all a day over the past week and I imagine this has caused problems for Europeans, even though the videos became available on the endpoint within a day. Now I don't have to fix it this weekend. I will look at your patch later today or tomorrow.

tpikonen:
Thanks for the patch. Youtube downloads are indeed broken (#1079) and this patch unfortunately does not fix them. It's not the fault of this code though, as the URL used in the unpatched gpodder (https://www.youtube.com/get_video_info?html5=1&el=detailpage&video_id= + id) gives a 404 response, at least where I am.

auouymous:
The `get_video_info` endpoint is 404 for the first ~24 hours after a release for everyone. The code then falls back to the new `watch` endpoint and works outside of Europe (and inside with this patch), even though users see a 404 warning. Once merged, I will flip the endpoints so `watch` is fetched first, and `get_video_info` will remain in case `watch` fails.

@tpikonen Thanks for testing. I'll probably just merge this tomorrow and fix the above issues myself to get this released.

auouymous:
Thanks for the patch. It has been applied along with 14bf8b44b7831d7f57b1a8d54b20e39a1a63488c to fix the above issues.

PR #1083, opened 2021-07-04
Add option to rename_download.py to put the sortdate after the podcast title

tkurtbond:
I find it very useful to have the option when renaming download files to put the sortdate after the podcast title.  This allows me to easily sort by podcast and then date within podcast.

### DIFF
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -24,7 +24,7 @@ __category__ = 'post-download'
 DefaultConfig = {
     'add_sortdate': False,  # Add the sortdate as prefix
     'add_podcast_title': False,  # Add the podcast title as prefix
-    'sortdate_after_podcast_title': False, # put the sortdate after podcast title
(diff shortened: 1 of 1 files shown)


auouymous:
The code looks good but that comment needs two spaces before the `#`. Try running `make lint` to see why.

I assume gpodder uses title by default for filenames and this extension gives you the option to use {title}.ext, {date}.ext or {date}-{title}.ext. And this patch also allows {title}-{date}.ext, correct?

tkurtbond:
> The code looks good but that comment needs two spaces before the `#`. Try running `make lint` to see why.

Done.
 
> I assume gpodder uses title by default for filenames and this extension gives you the option to use {title}.ext, {date}.ext or {date}-{title}.ext. And this patch also allows {title}-{date}.ext, correct?

Yes, that is correct.

auouymous:
Thanks for the patch.

tkurtbond:
> Thanks for the patch.

You are welcome!

PR #1084, opened 2021-07-07
Try new youtube endpoint first and fallback to the old endpoint.

auouymous:
@tpikonen Could you verify this works in Europe?


tpikonen:
Seems to work. No problems with streaming and downloading from Youtube with the youtube-dl extension disabled.

auouymous:
Thank you!

PR #1092, opened 2021-07-17
Fix soundcloud API change when adding new channels.

auouymous:
The API no longer uses the username to query user info. This requires an extra page fetch to find the user ID when adding new users.

### DIFF
--- a/src/gpodder/plugins/soundcloud.py
+++ b/src/gpodder/plugins/soundcloud.py
@@ -90,11 +90,29 @@ class SoundcloudUser(object):
         global CONSUMER_KEY
         key = ':'.join((self.username, 'user_info'))
         if key in self.cache:
-            return self.cache[key]
+            if self.cache[key].get('code', 200) == 200:
+                return self.cache[key]
 
         try:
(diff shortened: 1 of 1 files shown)


src/gpodder/plugins/soundcloud.py
```diff
@@ -90,11 +92,29 @@ def get_user_info(self):
...
-            return self.cache[key]
+            if self.cache[key].get('code', 200) == 200:
+                return self.cache[key]
 
         try:
-            json_url = 'https://api.soundcloud.com/users/%s.json?consumer_key=%s' % (self.username, CONSUMER_KEY)
-            user_info = util.urlopen(json_url).json()
+            # find user ID in soundcloud page
+            url = 'https://soundcloud.com/' + self.username
+            r = requests.get(url)
```
elelay, line 101: why not use **util.urlopen**?

elelay:
otherwise LGTM

src/gpodder/plugins/soundcloud.py
auouymous: Oops! I read it as *urllib* and blindly replaced it with requests. I'll fix that and change the youtube.py code to also use it.

PR #1093, opened 2021-07-17
Reset shownotes to top when changing episodes.

auouymous:
This prevents the title from being partially clipped when moving between episodes after scrolling the shownotes for one of them.


elelay:
LGTM

PR #1094, opened 2021-07-17
Create an html description when none provided.

auouymous:
The youtube-dl extension code is moved into gPodder to provide this feature for all feeds lacking an html description. This enables clickable links inside non-html descriptions. And if available, an image tag for the episode art URL is prepended to the html description (currently only available in html shownotes).


elelay:
LGTM

Marrib:
Does this fix also work for non-YouTube feeds?

PR #1095, opened 2021-07-17
Fix links in About dialog by connecting signal and opening website.

auouymous:


elelay:
But it was already working!

auouymous:
It wasn't working for me in Linux, and today I tested on Windows and it also didn't work.

PR #1105, opened 2021-07-23
gtkui: fix loading of cached thumbnails

ollieparanoid:
Due to this bug, thumbnails did not get loaded from the sqlite database,
but instead were regenerated on each start. Depending on the device this
leads to significant startup slowdown (30 podcast -> ~20s slowdown on my
PinePhone).


auouymous:
Thank you! I have 79 channels and it reduced startup time from 3s to 2s.

elelay:
Amazing, thanks!
It seems to startup instantly now (2s vs 8s).

PR #1106, opened 2021-07-23
gtkui: properly scale cover/pill on high resolution displays

ollieparanoid:
Fix the cover images and "pill" looking blurry on high resolution displays. For example, when the adaptive version of gPodder is used with Phosh on the PinePhone.

Get the scale parameter from the gtkTreeView object and pass it down to the cover thumbnail size and to draw_text_pill. If the scaling is not 1, then use the new draw_iconcell_scale() function to draw the pixbuf with high resolution.

## Screenshots

Before and after, with scaling factor 2:

### Regular UI
<details>

</details>
<details>

</details>

### Adaptive UI (@tpikonen)
<details>

</details>

<details>

</details>

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -48,7 +48,7 @@ from .desktop.welcome import gPodderWelcome
 from .desktopfile import UserAppsReader
 from .download import DownloadStatusModel
 from .draw import (cake_size_from_widget, draw_cake_pixbuf,
-                   draw_text_box_centered)
+                   draw_iconcell_scale, draw_text_box_centered)
 from .interface.addpodcast import gPodderAddPodcast
 from .interface.common import BuilderWidget, TreeViewHelper
 from .interface.progress import ProgressIndicator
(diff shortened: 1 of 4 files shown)


src/gpodder/gtkui/main.py
```diff
@@ -678,14 +678,17 @@ def on_find_podcast_activate(self, *args):
 
     def init_podcast_list_treeview(self):
         size = cake_size_from_widget(self.treeChannels) * 2
-        self.podcast_list_model.set_max_image_size(size)
+        scale = self.treeChannels.get_scale_factor()
```
ollieparanoid, line 681: Apparently this API is only in GTK 3.10 and above, not sure which GTK version is the minimum supported version for gPodder. If needed I can test if this is available. Suggestions welcome :smiley:

elelay: looks good to me.
Indeed min gtk+ version is not tracked (macOS bundle is 3.24). Let's add a Gtk+ requirement line in the readme (after pygobject).

ollieparanoid: Done

elelay:
Merged, thanks!

PR #1107, opened 2021-07-23
gtkui: refactor resize_pixbuf_keep_ratio

ollieparanoid:
Do all image size calculation before calling scale_simple, so it only
needs to be called once.

Add a debug log message, since scaling cover images actually takes some
time and it's good to know the target resolution. (On my PinePhone,
scaling these cover images delays startup significantly and thumbnails
are too small when the program is scaled in the Phosh UI).

While at it, return early if no resize is needed so the changed and
result variables can be removed.

Calling scale_simple twice happens with the previous code, if the height
of the source image is greater than max_image_side and greater than the
width. In that case, the resulting image has less quality since the
second scale_simple already works with a scaled down image and scaling
takes twice as long. In practice most podcasts have square artwork so
I assume this didn't have much of a practical effect. But I care about
the log message and figured this would be the most elegant way to add it.

### DIFF
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -663,31 +663,27 @@ class PodcastListModel(Gtk.ListStore):
         Returns None if the pixbuf does not need to be
         resized or the newly resized pixbuf if it does.
         """
-        changed = False
-        result = None
-
         if url in self._cover_cache:
             return self._cover_cache[url]
 
-        # Resize if too wide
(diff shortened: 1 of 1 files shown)


auouymous:
Other than that, it LGTM, I just need to test it.

ollieparanoid:
> Other than that, it LGTM, I just need to test it.

Other than what? :) Sounds like a review comment was not submitted yet?

auouymous:
Thanks for the patch, it will improve the performance of a feature I'm working on.

PR #1108, opened 2021-07-23
bin/gpodder: add -c / --close-after-startup option

ollieparanoid:
Group command-line options into rest / logging / advanced, and add a new option to advanced that helps with profiling startup performance. (I used this together with a profiler to figure out that it spends most of the time resizing covers, which lead to #1105).

### DIFF
--- a/bin/gpodder
+++ b/bin/gpodder
@@ -113,10 +113,14 @@ def main():
                            action="store_true", dest="quiet", default=False,
                            help=_("reduce warnings on the console"))
 
+    grp_advanced = OptionGroup(parser, "Advanced")
+    parser.add_option_group(grp_advanced)
+
+    grp_advanced.add_option("--close-after-startup", action="store_true",
+                            help=_("exit once started up (for profiling)"))
+
     # On Mac OS X, support the "psn" parameter for compatibility (bug 939)
     if gpodder.ui.osx:
-        grp_advanced = OptionGroup(parser, "Advanced")
-        parser.add_option_group(grp_advanced)
         grp_advanced.add_option('-p', '--psn', dest='macpsn', metavar='PSN',
                                 help=_('Mac OS X application process number'))
 
--- a/bin/gpodder
+++ b/bin/gpodder
@@ -33,7 +33,7 @@ import os.path
 import platform
 import subprocess
 import sys
-from optparse import OptionParser
+from optparse import OptionGroup, OptionParser
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


auouymous:
I like the look of the groups, but it hides `--subscribe`, the only real option, because those groups standout  more. Maybe if subscribe was in a group...

I definitely will be using `--close-after-startup` to benchmark attempts at reducing startup time. However, I'm not sure if it should ship with gpodder, and if it did, only the long option name should be available in case another option needs the `-c` later on, but also to make it harder to accidentally use.

But if added, it might be good to call `self.core.db.close()` before exiting.

@elelay What do you think?

elelay:
> @elelay What do you think?

agreed with all your comments

ollieparanoid:
Patches updated,

* --subscribe is in a group now so it stands out more
* short option name -c is removed
* `self.core.db.close` is called before exit

New help output:
```
$ bin/gpodder -h
Usage: gpodder [options]

gPodder enables you to subscribe to media feeds (RSS, Atom, YouTube,
Soundcloud and Vimeo) and automatically download new content.

This is the gPodder GUI. See gpo(1) for the command-line interface.

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit

  Subscriptions:
    -s URL, --subscribe=URL
                        subscribe to the feed at URL

  Logging:
    -v, --verbose       print logging output on the console
    -q, --quiet         reduce warnings on the console

  Advanced:
    --close-after-startup
                        exit once started up (for profiling)
```

> However, I'm not sure if it should ship with gpodder, [...]

Personally I'd find it useful. It would allow to use this from any branch without cherry picking the patch(es) first, and it would make it available in distro packages (I'll probably test some time soon how much I can reduce startup time by not installing all plugins at once - in Alpine Linux / postmarketOS it's not uncommon to split out plugins into extra packages since the overhead is very small). So for that it it would be helpful too if the patch was in the main repository.

But in the end it's your (the maintainers) call, and if you decide that you'd rather not have these two patches in master then that's also fine. :+1:

ollieparanoid:
Cool that there's a linter in CI now! however the failure seems unrelated? it just says `Error: ` (or am I reading it wrong?)

I've tried to follow PEP-8 in the patches.

thp:
> Cool that there's a linter in CI now! however the failure seems unrelated? it just says `Error: ` (or am I reading it wrong?)
> 
> I've tried to follow PEP-8 in the patches.

It shows it in the output: https://github.com/gpodder/gpodder/pull/1108/checks?check_run_id=3212955349 and even suggests a fix:

```
-from optparse import OptionParser, OptionGroup
+from optparse import OptionGroup, OptionParser
```

auouymous:
The first patch is a good addition. It is the second patch that has very limited use and could be done by adding two lines into the source when needed.

Can the groups be rearranged? If so, do you think Logging(top), Advanced(middle), Subscriptions(bottom) would be better?

ollieparanoid:
They can be re-arranged, I could put them in the suggested order. However, shouldn't we put the most important group on top? I'd expect users to read the help output from top to bottom (and this is how I read it myself) :)

> It is the second patch that has very limited use and could be done by adding two lines into the source when needed.

Another nice thing would be that people could replicate the startup time test easily, without the issue of possibly inserting lines at the wrong place and therefore doing a different measurement.

bin/gpodder
```diff
@@ -95,21 +95,34 @@ def main():
...
+
+    grp_logging.add_option("-v", "--verbose",
+                           action="store_true", dest="verbose", default=False,
+                           help=_("print logging output on the console"))
+
+    grp_logging.add_option("-q", "--quiet",
+                           action="store_true", dest="quiet", default=False,
+                           help=_("reduce warnings on the console"))
+
+    advanced = OptionGroup(parser, "Advanced")
```
tpikonen, line 116: Name the variable grp_advanced?

tpikonen (approved):
I think this is a good addition to gpodder.

bin/gpodder
ollieparanoid: Right, that seems more consistent. Done!

elelay:
merged, thanks!

PR #1114, opened 2021-07-30
Switch from Travis CI to Github Actions.

auouymous:
@elelay Should we try this out?


auouymous:
It seems to be working in this PR, but failed in `make unittest` for `make releasetest`.

```
    LC_ALL=C PYTHONPATH=src/ $(PYTEST) --ignore=tests --ignore=src/gpodder/utilwin32ctypes.py --doctest-modules src/gpodder/util.py src/gpodder/jsonconfig.py
```

```
ERROR src/gpodder/util.py - ModuleNotFoundError: No module named 'gi'
```

thp:
Likely needs https://packages.debian.org/de/sid/python3-gi installed.

auouymous:
It appears python3-gi was already installed.

```
python3-gi is already the newest version (3.36.0-1).
```

I don't know if it is visible but the log is at https://github.com/auouymous/gpodder/actions.

elelay:
Thanks for the link: I was indeed looking for job output!
I think gi has been added for gio sync => it was not there on travis ci.
Maybe import gi only in functions using it?

elelay:
or maybe we are in a virtualenv with setup-python, without access to sytem site-packages?

auouymous:
Ahh, thank you. I actually imported gi first with `copy_text_to_clipboard`, and was confused as to why I imported it in the function if it was already imported for the file. Good to know it wasn't there, and I agree, we could move it inside the function like is done with `copy_text_to_clipboard`.

I don't know what virtualenv or setup-python is.

auouymous:
Importing gi inside functions worked, and the tests are passing now.

thp (approved):

elelay:
Cool! Let's merge :-)

PR #1119, opened 2021-08-03
Allow re-download of failed episodes without manually canceling them.

### LINKED ISSUE #1117: Unable to download an episode that has failed, without first cancelling it
Subscribe to https://www.youtube.com/channel/UCX4Vp6cm35OcPTQVN5_zeHQ and attempt to download the age-restricted video. It will fail because gpodder doesn't support age-gated content and youtube-dl is currently broken. Now try to download it again, the error icon disappears but nothing happens. You have to open progress and cancel the failed download before attempting to try the download again.

@blushingpenguin Prior to #1054, failed episodes could be re-downloaded without manually cancelling them.

auouymous:
Fixes #1117.

@blushingpenguin Is there a better way to handle this?

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -460,6 +460,8 @@ class DownloadQueueManager(object):
     def queue_task(self, task):
         """Marks a task as queued
         """
+        if task.status != DownloadTask.QUEUED:
+            self.tasks.queue_task(task)
         self.__spawn_threads()
 
 
--- a/src/gpodder/gtkui/download.py
+++ b/src/gpodder/gtkui/download.py
@@ -161,6 +161,10 @@ class DownloadStatusModel:
         iter = self.list.append()
         self.request_update(iter, task)
 
+    def queue_task(self, task):
+        task.status = task.QUEUED
+        self.work_queue.add_task(task)
+
     def register_task(self, task):
         self.work_queue.add_task(task)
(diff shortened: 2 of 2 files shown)


auouymous:
I guess this isn't the correct way to handle it because the episode icon isn't set when re-queued.

blushingpenguin:
It still races, but I think you'd find it much harder to hit the problem than before.

 def queue_task(self, task):
        self.work_queue.add_task(task)
        task.status = task.QUEUED
        
Once the task is added to the work queue, then the task could start and finish immediately (on a different thread). After that it could have the status set to queued again. This isn't likely to happen if the task takes some time to run, it might happen if it fails immediately for some reason.

I will give this another go at the weekend. I usually design this kind of thing by having an event queue that all asynchronous events are posted to, then a single threaded loop which processes the events and updates the objects (so starting/stopping a download in the ui would just post a start/stop request then the event processor would update the actual task object). This means that the object state is only mutated in one place which I generally easier to reason about since there's no need to worry about concurrent state changes. I will try something like that and see if it works out.

In the worst case I can add a patch to revert my earlier fix attempts (they aren't related to the gvfs changes, they were just bugs I came across while making them).

auouymous:
This PR or any other PR that fixes #1117 should not be applied until #1147 is fixed.

auouymous:
Obsoleted by #1149.

PR #1120, opened 2021-08-04
Allow '+' as a version number separator

tpikonen:
The adaptive branch currently has version numbers like '3.10.21+1', which breaks when trying to split() the version string on periods. This PR fixes or removes the version string splitting.

The unused variable `__version_info__` is removed, and re.split() is used in util.get_update_info() which allows splitting also on '+'.

### DIFF
--- a/src/gpodder/__init__.py
+++ b/src/gpodder/__init__.py
@@ -26,7 +26,10 @@ __copyright__ = '© 2005-2021 The gPodder Team'
 __license__ = 'GNU General Public License, version 3 or later'
 __url__ = 'http://gpodder.org/'
 
-__version_info__ = tuple(int(x) for x in __version__.split('.'))
+# Use public version part for __version_info__, see PEP 440
+__public_version__, __local_version__ = next(
+    (v[0], v[1] if len(v) > 1 else '') for v in (__version__.split('+'),))
+__version_info__ = tuple(int(x) for x in __public_version__.split('.'))
 
 import gettext
 import locale
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1853,7 +1853,11 @@ def get_update_info():
     days_since_release = (datetime.datetime.today() - release_parsed).days
 
     def convert(s):
-        return tuple(int(x) for x in s.split('.'))
+        # Use both public and local version label, see PEP 440
+        pubv, locv = next(
+            (v[0], v[1] if len(v) > 1 else '') for v in (s.split('+'),))
+        return tuple(int(x) if x.isdigit() else x.lower()
+            for x in pubv.split('.') + (locv.split('.') if locv else []))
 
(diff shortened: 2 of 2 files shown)


elelay:
why not... `__version_info__` is found in some other packages (https://softwareengineering.stackexchange.com/questions/375922/whats-the-version-info-function-in-a-pypi-package) to do clean comparison, but not used indeed.

tpikonen:
Obviously, re.split() can also be used to parse `__version_info__` if it's needed. Your call.

elelay:
I'm ok with removing it.

thp:
The __version_info__ is useful for doing tuple comparisons and using
gpodder as a library, as you can do e.g. gpodder.__version_info__ >= (3,
10, 4).

As for the adaptive branch, why use “+” and not just e.g. a fourth
component?

Eric Le Lay ***@***.***> schrieb am Do. 5. Aug. 2021 um 20:46:

tpikonen:
@thp, the idea was to visually separate the adaptive revisions from the master branch releases, but as seen, this may cause more problems than it solves.

The reasonable options are to use re.split('\.|\+') for splitting the version string everywhere (see the latest push) or I start to use '.' as the adaptive revision separator. Either way is fine for me, but I would prefer the first option.

elelay:
Using a **+adaptive-N** suffix would make it a local version, according to https://www.python.org/dev/peps/pep-0440/#development-release-separators. Is it what we're after?

tpikonen:
Interesting link. The latest push now handles the local version label as required in PEP 440 and leaves the public version part handling as it was.

tpikonen:
BTW, there's a Python library function for version number parsing called packaging.version.parse (also included in setuptools as pkg_resources.packaging.version.parse), but using that requires adding either packaging or setuptools as dependency.

elelay:
Adding the extra dependency seems a bit too much.

Regarding versioning, what format do you intend to use: `3.10.21+1` or `3.10.21+adaptive-1`?

It seems to work as expected anyway:

````
>>> convert('3.10.21+1')
(3, 10, 21, 1)
>>> convert('3.10.21')
(3, 10, 21)
>>> (3, 10, 21, 1) >= (3, 10, 21)
True
>>> (3, 10, 21, 1) >= (3, 10, 22)
False
>>> convert('3.10.21+adaptive-1')
(3, 10, 21, 'adaptive-1')
>>> convert('3.10.21+adaptive-1') >= convert('3.10.21')
True
>>> convert('3.10.21+adaptive-1') >= convert('3.10.22')
False

tpikonen:
I'll leave the adaptive branch versioning as it is, ('3.10.21+1' etc.), unless there is good reason to change it.

thp (approved):
Works for me, but it's @elelay's call :)

elelay:
That's it then :-)

elelay:
Thanks

PR #1124, opened 2021-08-12
fix #1099 folder not escaped when podcast renamed from dialog

### LINKED ISSUE #1099: [Enhancement Idea] Automatically replace unsupported characters in podcast name with custom character and keep the original name inside of gPodder
**Current state**
I was trying to add a YouTube channel to gPodder which has a colon ":" in its name like "Super:Duper:Channel". I am on Windows so colons (and some other special characters) are not allowed in file or folder names in the file system (which is NTFS). Because of this gPodder does not accept the name "Super:Duper:Channel" so I had to rename it to something without the colon. This is not optimal because it forces users to alter the way of writing the original name.

**Idea**
It would be cool if the original name with the special characters could be preserved _inside_ gPodder in the podcast list. As for the folder and file names on the hard drive, gPodder could automatically replace unsupported characters with "_" or "-" or some other character that the user can specify somewhere in the preferences. This way the original podcast name would be preserved, at least in the podcast list.

elelay:


auouymous:
LGTM.

PR #1127, opened 2021-08-12
Double-click or press Enter to open channel settings.

auouymous:
This provides similar behavior to opening shownotes for episodes.


elelay:
Nice addition, thanks!

PR #1129, opened 2021-08-14
Remove GTK stock items and other deprecations

tpikonen:
This PR removes all usage of deprecated GTK stock features. I used Glade to fix the ui-files, which caused quite a bit of churn, but hopefully this makes future UI changes easier.

One instance of (deprecated) Gtk.Action was also replaced by Gio.SimpleAction.

I went through all the dialogs and they seem to work, but I mostly use the adaptive version, so I may have missed some of the breakage caused by saving the ui-files with Glade. Please test.

### DIFF
--- a/share/gpodder/extensions/concatenate_videos.py
+++ b/share/gpodder/extensions/concatenate_videos.py
@@ -38,8 +38,8 @@ class gPodderExtension:
         dlg = Gtk.FileChooserDialog(title=_('Save video'),
                 parent=self.gpodder.get_dialog_parent(),
                 action=Gtk.FileChooserAction.SAVE)
-        dlg.add_button(Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL)
-        dlg.add_button(Gtk.STOCK_SAVE, Gtk.ResponseType.OK)
+        dlg.add_button('_Cancel', Gtk.ResponseType.CANCEL)
+        dlg.add_button('_Save', Gtk.ResponseType.OK)
 
         if dlg.run() == Gtk.ResponseType.OK:
             filename = dlg.get_filename()
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ PyPI. With this, you get a self-contained gPodder CLI codebase.
 ### GTK3 UI - Additional Dependencies
 
 - [PyGObject](https://wiki.gnome.org/PyGObject) 3.22.0 or newer
-- [GTK+3](https://www.gtk.org/) 3.10 or newer
+- [GTK+3](https://www.gtk.org/) 3.16 or newer
 
 
 ### Optional Dependencies
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -37,7 +37,7 @@
                 <property name="is-important">True</property>
                 <property name="label" translatable="yes">Download</property>
                 <property name="use-underline">True</property>
-                <property name="stock-id">gtk-go-down</property>
+                <property name="icon-name">go-down</property>
                 <signal name="clicked" handler="on_download_selected_episodes" swapped="no"/>
               </object>
               <packing>
@@ -51,7 +51,8 @@
                 <property name="sensitive">False</property>
                 <property name="can-focus">False</property>
                 <property name="is-important">True</property>
-                <property name="stock-id">gtk-media-play</property>
+                <property name="label" translatable="yes">Play</property>
+                <property name="icon-name">media-playback-start</property>
                 <signal name="clicked" handler="on_playback_selected_episodes" swapped="no"/>
               </object>
               <packing>
@@ -67,7 +68,7 @@
                 <property name="is-important">True</property>
                 <property name="label" translatable="yes">Cancel</property>
                 <property name="use-underline">True</property>
-                <property name="stock-id">gtk-cancel</property>
+                <property name="icon-name">gtk-cancel</property>
                 <signal name="clicked" handler="on_item_cancel_download_activate" swapped="no"/>
               </object>
               <packing>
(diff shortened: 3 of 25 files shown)


share/gpodder/extensions/concatenate_videos.py
```diff
@@ -38,8 +38,8 @@ def _get_save_filename(self):
         dlg = Gtk.FileChooserDialog(title=_('Save video'),
                 parent=self.gpodder.get_dialog_parent(),
                 action=Gtk.FileChooserAction.SAVE)
-        dlg.add_button(Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL)
-        dlg.add_button(Gtk.STOCK_SAVE, Gtk.ResponseType.OK)
+        dlg.add_button('_Cancel', Gtk.ResponseType.CANCEL)
```
elelay, line 41: aren't you breaking localization of the buttons? Or is '_Cancel' automatically translated by gtk?

auouymous: I don't think it is translated, see https://stackoverflow.com/questions/36805505/gtk-stock-is-deprecated-whats-the-alternative/36811163#36811163.

I haven't had time to look at this PR yet, but we should also fix the menu items that currently use _('_Cancel') and _('_Open'), if not already done by this PR.

thp: Yeah, `_()` has to be used for gettext (`_ = gpodder.gettext` is there in the global scope for this).

auouymous: Well, according to that link, we should be used dgettext. Declaring `D_ = t.dgettext` and then using `D_('gtk3', '_Cancel')` to translate strings known to gtk.

tpikonen: The localization breaks, although it should not according to docs. Using `D_('gtk30', '_Blah')` for all the GTK strings works, but is a bit of an eyesore.

Here's a patch on top of this PR which makes the translation of the Close button in the About dialog work again. Should we use this style, or maybe we can just replace all the GTK strings with local translations, like `_('Blah')`?

```--- a/src/gpodder/__init__.py
+++ b/src/gpodder/__init__.py
@@ -129,6 +129,8 @@ if ui.win32:
     from gpodder.utilwin32locale import install
     install(textdomain, locale_dir)
 
+dgettext = gettext.dgettext
+
 t = gettext.translation(textdomain, locale_dir, fallback=True)
 
 gettext = t.gettext
diff --git a/src/gpodder/gtkui/app.py b/src/gpodder/gtkui/app.py
index b77baa05..0d8913d4 100644
--- a/src/gpodder/gtkui/app.py
+++ b/src/gpodder/gtkui/app.py
@@ -45,6 +45,7 @@ logger = logging.getLogger(__name__)
 
 _ = gpodder.gettext
 N_ = gpodder.ngettext
+D_ = gpodder.dgettext
 
 
 def parse_app_menu_for_accels(filename):
@@ -202,7 +203,7 @@ class gPodderApplication(Gtk.Application):
     def on_about(self, action, param):
         dlg = Gtk.Dialog(_('About gPodder'), self.window.gPodder,
                 Gtk.DialogFlags.MODAL)
-        dlg.add_button('_Close', Gtk.ResponseType.OK).show()
+        dlg.add_button(D_('gtk30', '_Close'), Gtk.ResponseType.OK).show()
         dlg.set_resizable(True)
 
         bg = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=6, margin=16)
```

auouymous: _Cancel and _Open already use _() and are untranslated for most languages, but using D_ will give them translations for all languages, no matter how verbose they will become.

tpikonen: We could define `Dgtk_(s) = lambda: s D_('gtk30', s)` and use `Dgtk_('_Close')` for GTK strings. Would this be ok?

elelay:
Also don't forget the `_` in `"_Close"` is for accelerator support (Pressing Alt shows the accelerator, Alt-C activates it).

share/gpodder/extensions/concatenate_videos.py
tpikonen: The Gtk.STOCK label replacement strings are now translated with GTK translations.

tpikonen:
@auouymous, I can't replicate the missing menubar with GTK_CSD=0 here. Can you provide more details or bisect the patch stack? Can you test without GTK_CSD=0?

My guess is that this is related to setting the 'application' property in the code in fc6177a7b9. Could you try if putting it back to gpodder.ui with this patch helps?
```
diff --git a/share/gpodder/ui/gtk/gpodder.ui b/share/gpodder/ui/gtk/gpodder.ui
index f2faf177..d3e3e0c7 100644
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -15,6 +15,7 @@
   </object>
   <object class="GtkApplicationWindow" id="gPodder">
     <property name="name">gPodder</property>
+    <property name="application">app</property>
     <property name="title">gPodder</property>
     <property name="window-position">center</property>
```

auouymous:
That patch fixed the menubar.

Why did you replace checkboxes with toggle buttons?

auouymous:
The quit button on toolbar is missing an icon.

The close icon in channel and episode find entries are missing.

What are all the placeholders in progress tab?

The `Select None` button on episode selector dialog is missing an icon.

auouymous:
The menubar is still missing.

tpikonen:
Ok, one more try. @auouymous, could you check if the latest push fixes the menubar?

auouymous:
That fixed the menubar.

The `General` tab in preferences is missing icons on the two configure buttons.

The `gpodder.net` tab in prefs is missing padding for the container.

Using tab key to focus is broken in all preference tabs.

tpikonen:
More fixes, thanks @auouymous.

Since this PR is getting ridiculously long, I'm going to squash and split it to smaller parts.

I'm also having second thoughts about bd77467c3e54, i.e. using the GTK gettext domain. Would it be better to just add the standard dialog button strings ('OK, 'Close', etc.) to be translated in gpodder?

auouymous:
Using GTK's translations would be better since gpodder has almost none.

PR #1130, opened 2021-08-15
Query duration for youtube episodes when not using youtube-dl.

auouymous:
The duration is not available in the youtube feed and requires a request for each episode. It was set when downloading but the user had no way of knowing how long an episode was before downloading it. Live streams do not have a duration until they end, and remain blank until downloaded or the next update.

This will increase the time it takes to update feeds, with new subscriptions possibly taking 16x longer to update due to the 16 requests vs the previous single request. Updating existing feeds will only have an additional request for each new episode.


elelay:
OK for it, but it will be one more youtube trick to maintain.

PR #1131, opened 2021-08-15
Move scale_pixbuf to util.py and apply #1107 refactoring.

auouymous:
The function was moved because I have another patch that calls it from many files.


elelay:
ok :-)

PR #1132, opened 2021-08-17
Allow distributions to disable the update check.

### LINKED ISSUE #1103: Include a method for packagers to disable update check
See: https://bugzilla.redhat.com/show_bug.cgi?id=1984726

auouymous:
Package managers can install `/usr/share/gpodder/no-update-check` to disable the automatic check at startup, and the `Software update` menu item.

Fixes #1103.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -251,7 +251,8 @@ class gPodder(BuilderWidget, dbus.service.Object):
             diff = time.time() - self.config.software_update.last_check
             if diff > (60 * 60 * 24) * self.config.software_update.interval:
                 self.config.software_update.last_check = int(time.time())
-                self.check_for_updates(silent=True)
+                if not os.path.exists(gpodder.no_update_check_file):
+                    self.check_for_updates(silent=True)
 
     def create_actions(self):
         g = self.gPodder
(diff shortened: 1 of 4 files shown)


src/gpodder/gtkui/main.py
```diff
@@ -251,7 +251,8 @@ def changed_cb(spinbutton):
             diff = time.time() - self.config.software_update.last_check
             if diff > (60 * 60 * 24) * self.config.software_update.interval:
                 self.config.software_update.last_check = int(time.time())
-                self.check_for_updates(silent=True)
+                if not os.path.exists(gpodder.no_update_check_file):
```
oturpe, line 254: Would it be cleaner if this check was made earlier?
The results of time diff calculation does not affect this check
so it would be possible to short circuit earlier.

Of course this is a matter of style,
and gpodder is not my project in any way,
so it may be that I am just used to different conventions.

auouymous: The time check is a fast calculation that only succeeds once every 5 days, unless the user changes it. Where as the file check must access a potentially slow storage device, and increase startup time.

oturpe: Makes sense, being fast at runtime is more important than tiny readability gains.

elelay:
Merged, thanks!

PR #1134, opened 2021-08-18
Gtk UI: Use the built-in Gtk3 about dialog

thp:
The Gtk3 About dialog (compared to the Gtk2 one) is actually good, no need to roll our own. We do lose the bug tracker link, but it's readily available on the website ("Get help" / "Report problems at the issue tracker").

### DIFF
--- a/src/gpodder/gtkui/app.py
+++ b/src/gpodder/gtkui/app.py
@@ -200,42 +200,24 @@ class gPodderApplication(Gtk.Application):
         self.menu_popover.popup()
 
     def on_about(self, action, param):
-        dlg = Gtk.Dialog(_('About gPodder'), self.window.gPodder,
-                Gtk.DialogFlags.MODAL)
-        dlg.add_button(Gtk.STOCK_CLOSE, Gtk.ResponseType.OK).show()
-        dlg.set_resizable(True)
+        ABOUT_DIALOG_DATA = {
+            'program-name': 'gPodder',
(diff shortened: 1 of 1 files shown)


thp:
Before:

Screenshot from 2021-08-18 20-38-21

After:

Screenshot from 2021-08-18 20-38-39

(arguably this isn't strictly an improvement, but it uses more standard widgets)

auouymous:
#946 just narrowed the About dialog for mobile and this appears to fatten it back up, @tpikonen is this an issue?

tpikonen:
@auouymous, yes and no. The mobile linux distros (pmOS, mobian, pureOS) patch GTK to fix (among other things) the built-in about dialog, but a lot apps are now distributed with flatpak, which comes with it's own unpatched GTK.

The unpatched GTK about dialog is not totally unusable on mobile screens, but the current one in gPodder is better.

auouymous:
The native dialog assumes gvfs which results in unclickable links when not installed, the fix is at https://github.com/gpodder/gpodder/pull/1071#issuecomment-881995546.

auouymous:
And my about dialog (with patch) has a close button in the lower right corner, and it doesn't work.

thp:
Let's keep it as-is for now.. Once Gtk/Flatpak/mobile distros have figured out a way to have the stock dialog properly in all cases, including Flatpak, we can probably revisit that...

PR #1135, opened 2021-08-18
Gtk UI: Use a TextView for dialogs with multiple text lines

thp:
Also fixes the secondary text formatting in the normal message dialog.


thp:
Before:

Screenshot from 2021-08-18 20-12-02

After:

Screenshot from 2021-08-18 20-20-26

auouymous:
The OK button should have default focus.

thp:
Updated PR, the OK button now has default focus, and pressing Enter discards the dialog.

auouymous:
LGTM, thanks!

PR #1136, opened 2021-08-19
Gtk UI: Use GtkStack and put switcher in header bar

thp:
This uses `GtkStack` and `GtkStackSwitcher` to be more in line with the GNOME HIG. Still have to figure out where to put all the items from the main menu + toolbar (those should most likely not appear in the final UI. Also made sure that the non-CSD (`GTK_CSD=0`) UI gains access to the main menu again and that everything is workable there as well.

I haven't looked at the adaptive branch; maybe this becomes obsolete with the adaptive branch, but it still looks nice in the Linux Desktop situation (with Ubuntu/GNOME).

Without CSD (note the added "gPodder" menu, which contains the app menu):

Screenshot from 2021-08-19 16-19-45

With CSD (again, the menu bar is only temporary until those items can find a better spot, most likely in the header bar like Nautilus has):

Screenshot from 2021-08-19 16-19-30

We could replace this with `HdyViewSwitcher` in the future: https://developer.gnome.org/hig/patterns/nav/view-switchers.html

### DIFF
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -31,8 +31,9 @@
     <property name="urgency_hint">False</property>
     <signal handler="on_gPodder_delete_event" name="delete-event"/>
     <child>
-      <object class="GtkGrid" id="vMain">
+      <object class="GtkVBox" id="vMain">
         <property name="visible">True</property>
+        <property name="spacing">0</property>
         <property name="orientation">vertical</property>
         <child>
           <object class="GtkToolbar" id="toolbar">
@@ -140,27 +141,26 @@
               </packing>
             </child>
           </object>
+          <packing>
+            <property name="expand">False</property>
+          </packing>
         </child>
         <child>
           <object class="GtkGrid" id="hboxContainer">
-            <property name="border_width">5</property>
+            <property name="border_width">0</property>
             <property name="visible">True</property>
             <property name="orientation">horizontal</property>
             <property name="hexpand">True</property>
             <property name="vexpand">True</property>
             <child>
-              <object class="GtkNotebook" id="wNotebook">
+              <object class="GtkStack" id="wNotebook">
                 <property name="visible">True</property>
                 <property name="can_focus">True</property>
-                <property name="show_tabs">True</property>
-                <property name="show_border">True</property>
-                <property name="tab_pos">GTK_POS_TOP</property>
-                <property name="scrollable">False</property>
-                <property name="enable_popup">False</property>
-                <signal handler="on_wNotebook_switch_page" name="switch_page"/>
+                <property name="transition-type">slide-left-right</property>
+                <signal handler="on_wNotebook_switch_page" name="notify::visible-child"/>
                 <child>
                   <object class="GtkPaned" id="channelPaned">
-                    <property name="border_width">5</property>
+                    <property name="border_width">0</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="orientation">horizontal</property>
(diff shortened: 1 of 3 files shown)


tpikonen:
This looks great, but I'm kind of missing the gPodder text in the headerbar. Since the progress view is clearly secondary, maybe a toggle button with a download symbol in the header bar would be enough for the switcher? Sorry for bikeshedding. 

The adaptive branch rewrites gpodder.ui anyway, so no problems there. The menus in adaptive are squeezed under two menubuttons plus a search button. The view controls are in the episode header bar, so they would need a separate menubutton here.

auouymous:
I've wanted the GtkNotebook gone for a long time, and this looks great for non-CSD.

Could the gap above the "Check for new episodes" button be removed?

Tab focus is broken. In the old UI it goes from notebook tabs -> channel list -> find channel -> check for new episodes button -> episode list -> find episode. Now tab cycles between podcasts button -> progress button -> ?.

auouymous:
Actually, that gap should probably not be removed, and a gap added on the other three sides of that button. When updating podcasts, the progress bar touches the left and bottom window borders and doesn't look good. So there should be a few pixels of gap around the widgets down there.

thp:
> Actually, that gap should probably not be removed, and a gap added on the other three sides of that button. When updating podcasts, the progress bar touches the left and bottom window borders and doesn't look good. So there should be a few pixels of gap around the widgets down there.

In theory we could put the refresh button from the CSD case to the top left (in the same "row" as the stack switcher buttons) and remove the "check for new episodes" button below the podcast list. The progress could probably be a spinner on the button or something, not sure, but would be great to combine all this UI into the refresh button in the top left (CSD) or left of the stack switcher button row (non-CSD case).

thp:
Marked as "draft", since this needs some more work :)

auouymous:
Moving the button would be nice, as it would provide more space for channels and fills in some of the empty space up top. Displaying progress as a "spinner" sounds like a removal of information. The button could simply change to the current set of widgets during an update. Otherwise we lose the channel name currently being updated, the overall progress bar and the cancel button.

thp:
Updated and moved the whole refresh UI next to the update button. During a feed update, the update button changes into a cancel button. After the update is completed, if "no new episodes" are found, the progress bar will disappear after 3 seconds (it was always a pet peeve of mine that if you get back to the window you have to press twice - once to confirm the "no new episodes", and then once more to refresh again).

This also moves the app menu button (burger icon) to the same row as the stack switcher, avoiding the need to put the app menu into the main menu.

With client-side decorations:

Screenshot from 2021-08-28 10-30-40

Without client-side decorations:

Screenshot from 2021-08-28 10-30-59

thp:
What is kind of annoying is the progress bar "jumping" in width as the text changes (this is also the reason why I moved it into its own row for the non-csd case, as the header bar seems to deal better with keeping the stackswitcher widget "centered"). Ideally I'd like to make the `Gtk.ProgressBar` fixed-width, but it seems Gtk doesn't provide a nice method to do this.

thp:
And here's updated screenshots while a update is NOT in progress (note the hamburger menu button for the non-CSD case):

Screenshot from 2021-08-28 10-43-49

Screenshot from 2021-08-28 10-44-23

thp:
One option might also be to just make the feed update part of the "progress" tab, and not special case it. (Display its progress like sync and downloads and allow cancelling like there, not showing anything in the main UI)

auouymous:
Thank you for making it automatically close if no new episodes, that has always annoying me too. I update my test feed a lot and must always cancel the progress before updating again. I think it is in my todo list to one day fix.

The hamburger menu is a CSD feature that doesn't belong in the non-CSD UI. It also broke accelerators such as ^Q to quit and ^P to open preferences.

Placing the progress bar below looks bad, and causes UI shifting when it opens and closes. The jumping is not an issue if the stack switcher buttons are right aligned.

And I think the progress bar should remain with the refresh button. It can be seen when either stack is active and compliments the download progress in titlebar. I rarely switch to the progress stack because channel updating is visible on main window and download status in titlebar.

elelay:
> One option might also be to just make the feed update part of the "progress" tab, and not special case it. (Display its progress like sync and downloads and allow cancelling like there, not showing anything in the main UI)

that could be nice. It may also allow for  progress report at the podcast level: on youtube-dl feeds it seems the update is stuck because it's parsing each episode's page.

PR #1145, opened 2021-08-28
Delay restoring window position until after it has been shown.

auouymous:
The window manager might not honor the move request until after the window is shown. This can cause window contents to be rendered at a shifted offset inside the window borders.


thp (commented):
Could this cause the window "jumping" from original mapping to when it's loaded fully? Wonder if we should do it twice? (once so that window managers that honor the request will move the window, then second time with delay so that other window managers deal with it properly) -- is there a list of window managers that have one behavior vs the other?

auouymous:
Thanks for bringing that up, the window can jump, so I change it to move before and after. There is a split second where the window contents are shifted but the delayed move quickly corrects that.

I doubt there is a list.

PR #1148, opened 2021-08-30
Use a GtkLabel for text shownotes

### LINKED ISSUE #1097: Links in text shownotes can't be copied
Label links in other parts of gPodder can be copied from the context menu, and is useful when emailing links or opening them in a non-default browser.

I have been unable to find a solution to fix this and am posting it here in case anyone else has ideas.

1) Replacing `self.text_buffer.insert_with_tags_by_name(self.text_buffer.get_end_iter(), text, 'hyperlink')` with `insert_link(text)` and adding the following function inside the `update` function mostly fixes it. The problem is that the links don't wrap, causing some to be moved to the next line and turning on the horizontal scrollbar. There was also an issue in one episode were I couldn't scroll to the last line. Enabling line wrap for the label does not work as expected.

```
        def insert_link(url):
            anchor = Gtk.TextChildAnchor()
            label = Gtk.Label()
#            label.set_line_wrap(True)
            label.set_markup('<a href="%s">%s</a>' % (url, url))
            label.connect('activate-link', lambda label, url: util.open_website(url))
            label.show()
            self.text_buffer.insert_child_anchor(self.text_buffer.get_end_iter(), anchor)
            self.text_view.add_child_at_anchor(label, anchor)
```

2) The `populate-popup` signal is called with `textview` and `menu` parameters when right-clicking in the text buffer. If there was a way to get the cursor position where the menu is opened, it would be possible to add items to the menu for opening or copying a link. I was able to listen for `button-press-event` (event.button==3) and inject a fake left button click to move cursor before opening the menu but it was far too hacky and would likely cause issues for some users.

tpikonen:
Define gPodderShownotesLabel class and use it for text shownotes. This is ported from adaptive, where setting 'selectable' to False makes swipeable containers (HdyLeaflet) work on touchscreens. The selectable property is set to True in this PR however. 

This should be mostly identical to gPodderShownotesText in look and function. 

Fixes #1097.


tpikonen:
Ok, as commented by @auouymous in #1097, setting selectable to True makes the label scroll to the top when clicking the text. Obviously this should be fixed.

auouymous:
The `#%X%X%X` should be  `#%02X%02X%02X` and it should only be done inside the label condition since html shownotes doesn't use it.

The link colors are correct and they have a hover color, even the title brightens on hover. What is the reason you change them to the GtkTextView colors? The label shownotes have a different background color than textview and might have different link colors to go with that background on some themes.

And self.foreground_color is not correct, it is not the color used by GtkLabel or GtkTextView. I'm not even sure how the title gets brighter on hover because it is using #FFFFFF for the non-hover color. Using self.label for dummy_tv instead of creating a GtkTextView gets the correct label color. I imagine html shownotes would also be using the wrong foreground color, the textview widget might need to be added to the window in order to correctly detect colors.

Selecting all of the shownotes text on focus is a little annoying, but not being able to use up and down arrows to move is worse. What does the keyboard_callback patch in adaptive branch do?

The self.status tooltip widget needs to be hidden because it has a different background color in my theme which leaves a small box in the lower left corner.

tpikonen:
The latest push fixes the color text format and tries to hide the status tooltip (I did not see the box here).

The colors are correct here, they are set on purpose to match the treeview colors. E.g. the default text color in Label is slightly lighter than in the treeview. Since I'm not seeing the same effects as you are, can you provide a patch where the colors work correctly for you? 

The keyboard_callback in adaptive was probably meant to be used with a TextView, which eats more keyboard events than a Label. It does not seem to be necessary in adaptive with ShownotesLabel. The lack of arrow key navigation in shownotes annoys me as well, so I'll see if it can be fixed with it.

tpikonen:
Rebased to master, added a fix for link title formatting.

@auouymous, it seems that setting colors via code is deprecated in GTK and CSS is the way to go with custom widget styling. I'll have a look if it can be made to work.

tpikonen:
Now uses the 'view' CSS class for the shownotes label and its background, which should have the correct colors without defining them in a separate CSS style.

Issue #1137, opened 2021-08-19
AttributeError: type object 'gPodderSyncUI' has no attribute 'DL_ADDED_ONE'


thp:
Introduced by #1054.

Steps to reproduce:

1. Download a file
2. Pause the download
3. Try to resume the download

```
Traceback (most recent call last):
  File ".../gpodder/src/gpodder/gtkui/main.py", line 1622, in <lambda>
    item.connect('activate', lambda item: self._for_each_task_set_status(tasks, status, force_start))
  File ".../gpodder/src/gpodder/gtkui/main.py", line 1567, in _for_each_task_set_status
    self.set_download_list_state(gPodderSyncUI.DL_ADDED_ONE)
AttributeError: type object 'gPodderSyncUI' has no attribute 'DL_ADDED_ONE'

```

auouymous:
Fixed by #1149.

Issue #1147, opened 2021-08-30
No such file or directory while downloading "{title}": {file}


auouymous:
Introduced by #1054 with #1119 applied.

Downloading DRM content with built-in youtube support will fail. Downloading again using the youtube-dl extension will fail again when the download is complete with a `No such file or directory while downloading "{title}": {file}` error. The file was downloaded but renaming failed, requiring the user to manually delete the file outside of gpodder. This only happens when there is a failed task, cancelling it before downloading with youtube-dl works fine.

Sample feed with DRM content https://www.youtube.com/user/BBCRadio1VEVO

PR #1149, opened 2021-08-30
#1117, #1137: fixes for pause/resume/cancelling download tasks

blushingpenguin:
For review/test -- I have tried to fix the race conditions in pause/resume/cancel/restarting downloads here. The main thing that this does is to introduce two new states, pausing and cancelling which allow the download task (which is running on a background thread) to transition itself to the paused/cancelled state, and also locks the task while the state is being manipulated. I think that the idea is basically sound but there are probably some issues left.


auouymous:
Minor linting issue but it seems to fix #1117 and #1137.

It gets stuck forever when cancelling from failed or paused states.

#1147 is still an issue and blocks this PR from being merged.

blushingpenguin:
@auouymous I've fixed the lint issue and I've also fixed a (reintroduced) race when queuing multiple items -- if you had hit that when testing then cancel/pause/resume would have been broken (it wasn't hard to hit). I have also tried the test in #1147 (download from https://www.youtube.com/user/BBCRadio1VEVO, let it fail, enable youtube-dl and retry the download) which worked for me.

blushingpenguin:
Also there is (at least) one issue left here, if you restart a download that gets removed from the active downloads list then it is impossible to cancel/pause it. This can be done by starting two downloads (A+B), cancelling A, waiting for A to be cancelled, cancelling B, quickly right clicking on A, waiting for the downloads list to be cleared and then choosing A from the context menu for the vanished item. The task is no longer in the 'download_tasks_seen' list and this means that attempts to manipulate it fail, but it carries on in the background.

auouymous:
The cancelling issue is fixed but #1147 is not. Are you letting the download finish? It fails both ways for me, using `Download with Youtube-DL` on right click menu (easier to test), or checking `manage_downloads` in config and downloading.

auouymous:
#1152 is another issue.

blushingpenguin:
> The cancelling issue is fixed but #1147 is not. Are you letting the download finish? It fails both ways for me, using `Download with Youtube-DL` on right click menu (easier to test), or checking `manage_downloads` in config and downloading.

Yes, I let it finish. I'm using https://www.youtube.com/user/BBCRadio1VEVO as the feed, then right clicking on 
"Billie Eilish - I'm In The Mood For Love (Julie London Cover) in the Live Lounge" and choosing download, which fails, then right clicking again and choosing "Download with Youtube-DL" which succeeds. I am also able to play the downloaded video afterwards. I have tried some other episodes with the same result

auouymous:
My fault, #1044 is causing #1147. I don't know how it is working for you, or how it was working for me when I reverted the #1054 patch. The failed attempt deletes the empty partial file and the youtube-dl extension calls `os.stat` on it because the DownloadTask is only initialized before the first attempt, and the partial file is only created during initialization.

Issue #1150, opened 2021-08-30
Feature: Add contributing.md


ColdIce1:
Not sure if the development steps mentioned in this issue #520 still apply, in the case they do they should be added to contributing.md to inform people how to get started contributing :)

1. Fork the project (button on top-right of this page).
    this will create a copy of the code at https://github.com/Marundu/gpodder
2. clone your fork to work on the code on your computer
3. `python3 tools/localdepends.py` to install dependencies
3. `./bin/gpodder -v` to show debug messages
3. make the changes, commit in a branch and push the branch to your fork
4. on your forked repository webpage on github, a "Make pull request" banner will appear for your new branch.
5. submit a pull request. I'll review it and help you get it in shape for incorporating in gPodder
6. Thanks and well done ;-)

_Originally posted by @elelay in https://github.com/gpodder/gpodder/issues/520#issuecomment-417210818_

PR #1151, opened 2021-08-31
Add contributing.md

ColdIce1:
See on-going issue #1150

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,23 @@
+# Contributing to this repository <!-- omit in toc -->
+
+## Getting started <!-- omit in toc -->
+
+Before you begin:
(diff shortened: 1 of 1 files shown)


elelay:
Seems good. Maybe add `make lint unittest` before push.
They in turn need `pip3 install pytest-cov minimock pycodestyle isort requests pytest pytest-httpserver`

ColdIce1:
Can I suggest we adding a development.txt for easy pip install in that case? :)

elelay:
yes, linting and unittests are done on the PR. 
I still recommend running `make lint unittest` because it's then easier to fix all errors in one go instead of 1 commit per try if one uses only the CI.

elelay:
Merged, thanks :-)

PR #1154, opened 2021-09-09
Preferences: Remove deprecated GTK features

tpikonen:
Split from #1129. Save the ui-file with Glade to regularize it, fix the collateral damage and remove deprecated GTK features.

Sets the required GTK version to max 3.16 everywhere and updates README.md to mention this.

Also sets 'transient-for' to parent in podcastdirectory, as that was broken in a previous commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ PyPI. With this, you get a self-contained gPodder CLI codebase.
 ### GTK3 UI - Additional Dependencies
 
 - [PyGObject](https://wiki.gnome.org/PyGObject) 3.22.0 or newer
-- [GTK+3](https://www.gtk.org/) 3.10 or newer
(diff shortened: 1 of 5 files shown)


auouymous:
Splitting would have been nice before I spent a week or two reviewing the original, now I must start over. But I will accept it as it is easier to squash smaller PRs into single commits.

Why is the podcastdirectory commit in this PR if something else broke it?

tpikonen:
Oh, sorry if this made things harder. The purpose of the splitting was of course to make the review easier. You can check that the changes to gpodderpreferences.ui are identical to #1129 and that the rest of the commits are trivial.

The podcastdirectory is here because submitting a one-line PR felt silly at the time. This can be split out if needed.

auouymous:
Thanks, merged.

PR #1155, opened 2021-09-12
fixes #1152 and an odd corner case

### LINKED ISSUE #1152: Re-ordering downloads in progress tab no longer works
Set a low download rate if you have a fast connection and set the limit to one download. Add three downloads, drag the bottom (# 3) download to the top, pause the current download (# 1). The top entry (# 3) should begin downloading but doesn't until entry # 2 finishes. The original order is retained even when moving them.

Introduced in #1054.

blushingpenguin:
To fix the task queue re-ordering, this basically reverts the chunk of #1054 that added a second thread safe tasks queue and goes back to using the Gtk.ListStore to contain the tasks work queue. To workaround the problem of the background download threads triggering exceptions when concurrently accessing the list store while the main ui thread is manipulating it, only the main ui thread accesses the list store, and background worker threads queue a request for a task to the main thread by adding an idle function which dequeues a task, then sends the result back to the background thread (via the DequeRequest class).

This also fixes the corner case of downloading a task that no longer exists. That can be triggered by downloading some episodes, then in the tasks list cancelling one of the downloads. Once cancellation has finished, cancel all other downloads, then quickly right click on the cancelled task to get the context menu. Wait until the other tasks have been cancelled and the tasks list has been cleared, then choose to download the cancelled task -- at this point the task is in the queued state, not in the list (and hence not in the queue) but attached to the episode,so it becomes impossible to download the episode that the task is attached to. This is fixed by detecting the case of the task not being in the tasks list and adding it back in if necessary.

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -460,6 +460,7 @@ class DownloadQueueManager(object):
     def queue_task(self, task):
         """Marks a task as queued
         """
+        self.tasks.queue_task(task)
         self.__spawn_threads()
 
 
@@ -469,7 +470,7 @@ class DownloadTask(object):
     You can create a new download task like this:
 
         task = DownloadTask(episode, gpodder.config.Config(CONFIGFILE))
-        task.status = DownloadTask.DOWNLOADING
+        task.status = DownloadTask.QUEUED
         task.run()
 
     While the download is in progress, you can access its properties:
@@ -486,7 +487,8 @@ class DownloadTask(object):
 
     You can cancel a running download task by setting its status:
 
-        task.status = DownloadTask.CANCELLED
+        with task:
+            task.status = DownloadTask.CANCELLING
 
     The task will then abort as soon as possible (due to the nature
     of downloading data, this can take a while when the Internet is
@@ -537,9 +539,9 @@ class DownloadTask(object):
     The same thing works for failed downloads ("notify_as_failed()").
     """
     # Possible states this download task can be in
-    STATUS_MESSAGE = (_('Queued'), _('Downloading'),
-            _('Finished'), _('Failed'), _('Cancelled'), _('Paused'))
-    (QUEUED, DOWNLOADING, DONE, FAILED, CANCELLED, PAUSED) = list(range(6))
+    STATUS_MESSAGE = (_('Queued'), _('Queued'), _('Downloading'),
+            _('Finished'), _('Failed'), _('Cancelling'), _('Cancelled'), _('Pausing'), _('Paused'))
+    (NEW, QUEUED, DOWNLOADING, DONE, FAILED, CANCELLING, CANCELLED, PAUSING, PAUSED) = list(range(9))
 
     # Wheter this task represents a file download or a device sync operation
     ACTIVITY_DOWNLOAD, ACTIVITY_SYNCHRONIZE = list(range(2))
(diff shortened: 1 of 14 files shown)


auouymous:
Is the second issue https://github.com/gpodder/gpodder/pull/1149#issuecomment-909259747?

blushingpenguin:
@auouymous Yes, that is the second issue

src/gpodder/gtkui/main.py
```diff
@@ -1177,7 +1177,7 @@ def update_downloads_list(self, can_call_cleanup=True):
 
                 task = row[self.download_status_model.C_TASK]
                 speed, size, status, progress, activity = task.speed, task.total_size, task.status, task.progress, task.activity
-                logger.info("%s: %f", task.episode.title, progress)
+                # logger.info("%s: %f", task.episode.title, progress)
```
auouymous, line 1180: This should be removed.

src/gpodder/gtkui/main.py
```diff
@@ -2244,6 +2238,7 @@ def play_or_download(self, current_page=None):
                     else:
                         streaming_possible |= self.streaming_possible(episode)
                         can_download = True
+                logger.info('%s - can_download=%s, can_play=%s, can_delete = %s' % (episode.description, can_download, can_play, can_delete))
```
auouymous, line 2253: You removed the other debug line in a later patch but missed this one.

src/gpodder/gtkui/main.py
```diff
@@ -2210,7 +2207,6 @@ def play_or_download(self, current_page=None):
             return (False, False, False, False, False)
 
         (can_play, can_download, can_cancel, can_delete) = (False,) * 4
-        (is_played, is_locked) = (False,) * 2
```
auouymous, line 2213: Removing these unused variables is unrelated to the fix and should happen in a separate PR.

auouymous (changes requested):

src/gpodder/download.py
```diff
@@ -936,14 +936,6 @@ def run(self):
             self.error_message = _('Error: %s') % (str(e),)
 
         with self:
-            if result == DownloadTask.FAILED:
```
auouymous, line 919: Is the moving of this block related to the fix or simply to make the normal path faster than the failure path?

auouymous (changes requested):

auouymous:
I'm unable to cancel failed downloads (the 'ccc' episodes at http://borg.qzx.com/test-rss.php always fail).

src/gpodder/download.py
blushingpenguin: it was meant to make things easier to read by making the paths less complicated, I've further refactored the code so it is now:

if SUCCESS:
  do things
  return True

clean up after pause/cancel/fail

if FAIL:
  things
elif status = PAUSING:
  self.status = PAUSED
elif status = CANCELLING:
  self.status = CANCELLED

I think that's even clearer as there are less paths through the code.  (Happy to change this back if you thought it was clearer before)

src/gpodder/gtkui/main.py
blushingpenguin: I've split this off into https://github.com/gpodder/gpodder/pull/1158

blushingpenguin:
> I'm unable to cancel failed downloads (the 'ccc' episodes at http://borg.qzx.com/test-rss.php always fail).
fixed

blushingpenguin:
Thanks for looking at this

auouymous:
Thanks for fixing these issues.

elelay:
Thank you both for your hard work making things work again after the switch to gio!

PR #1156, opened 2021-09-12
Remove deprecated GTK features from simple dialogs

tpikonen:
Removes some deprecated GTK features from gPodderWelcome, gPodderAddPodcast, gPodderConfigEditor and gPodderExportToLocalFolder. Split from #1129.

Changes from #1129:
 * 20577bd32a1c: gPodderWelcome: Replace padding with margins
 * gpodderaddpodcast.ui: Set can-focus to True where needed
 * 32c5ba476264: gpodderexporttolocalfolder.ui: Set can-focus in Dialog, reorder buttons


auouymous:
The welcome buttons lost their chunkiness, and the ability to tab between them.

Are the icons on buttons deprecated? They made it easy to quickly distinguish groups of buttons without reading the text.

Gpodder has no translations for "Paste" and while it has "Save" from the channel dialog, I didn't see any translations for it. And the gtk domain would guarantee "Cancel", "Add" and "Close" work for all languages. Removing the icons and translations will make it harder for non-English users.

tpikonen:
Thanks again for the review.

The latest push fixes the Welcome dialog and adds back the underscore accelarators to buttons.

The button icons are indeed deprecated, they need to be explicitly added, see https://docs.gtk.org/gtk3/property.Settings.gtk-button-images.html

The problem with using GTK translations in ui-files is that GtkBuilder allows setting the translation domain only on a per file basis. If a translation is not actively maintained, then this will be a regression. However, the number of new translatable strings added when stock strings are removed is likely very small, so at least it's not a big translation task.

auouymous:
Why did you disable the labels in the welcome buttons?
```
+                        <property name="sensitive">False</property>
```

The cancel button on welcome dialog lost some of its margin. Unrelated to this PR, but there is a lot of inconsistent padding around close buttons through out gpodder. Some have none (about, subscribe and edit config) and the others vary. We should probably choose a value and apply it to all dialogs.

Will you be adding the accelerator to the close button on preferences dialog?

I've only seen a couple languages receive translations this year, so it could take quite a while to update these. @elelay Are you okay with dropping the gtk translations, and the button icons?

tpikonen:
Added margins to Welcome Cancel button and accelerators to Preferences dialog buttons.

The button label sensitivity was turned off when I was fixing the tab focusing, does it cause problems?

auouymous:
Ya, it changes the text color of the button to indicate disabled. I tested tab focus yesterday to see if that was why you disabled the labels and they didn't become tab stops when enabled. Were they tab stops for you?

tpikonen:
Ok, the labels now have sensitive=True. They were left to False from trying to fix the tab stops which were broken for some other reason. Also the button label appearance is the same here, whether sensitive or not.

PR #1160, opened 2021-09-14
Use icons when pausing or cancelling tasks.

auouymous:
This avoids shifting the list left and then right for single tasks. It also gives more feedback that the task is being paused or cancelled.


elelay:
LGTM :+1:

PR #1161, opened 2021-09-14
Paused episode icon

auouymous:
This changes the download progress icon to a paused icon when an episode is in the pausing or paused states.

@blushingpenguin Will it cause any issues to not recycle the paused task?

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -776,7 +776,8 @@ class DownloadTask(object):
                     time.sleep(delay)
 
     def recycle(self):
-        self.episode.download_task = None
+        if self.status not in (self.FAILED, self.PAUSED):
+            self.episode.download_task = None
 
     def set_episode_download_task(self):
         if not self.episode.download_task:
--- a/src/gpodder/gtkui/model.py
+++ b/src/gpodder/gtkui/model.py
@@ -378,11 +378,21 @@ class EpisodeListModel(Gtk.ListStore):
         view_show_unplayed = False
         icon_theme = Gtk.IconTheme.get_default()
 
-        if episode.downloading:
-            tooltip.append('%s %d%%' % (_('Downloading'),
-                int(episode.download_task.progress * 100)))
+        task = episode.download_task
 
-            index = int(self.PROGRESS_STEPS * episode.download_task.progress)
+        if task is not None and task.status in (task.PAUSING, task.PAUSED):
+            tooltip.append('%s %d%%' % (_('Paused'),
+                int(task.progress * 100)))
+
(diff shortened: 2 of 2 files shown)


blushingpenguin:
@auouymous I don't think that will cause any issues, but I would suggest moving the check into the recycle function itself. (This wouldn't make any difference currently as recycle is only called with the task in the cancelled state, but if future code is added that calls recycle it might also need the special case)

I also noticed that this patch introduces a change in behaviour in that in the episode list previously you could choose to download a paused task from the context menu and now you can cancel it instead -- I think it should have both really to be consistent.  I've got a chunk of the way through adding that support (it works, but needs more testing -- I can probably finish that off this weekend). I don't think that should block your change but it seems nice to have.

auouymous:
Ok, I'll move the check inside recycle, it will give the function more purpose. Also, is there a reason failed tasks are removed (recycled) from episodes? It seems like any task listed on the progress page, other than cancelled, should still have the task in the episode, correct?

#1162 was meant to resolve that behavior change in a better way. Displaying `Resume` on the menu would make more sense than `Download`. And adding the ability to pause from the menu would be an enhancement.

blushingpenguin:
Yes, I think keeping the task connected to the episode as long as it is in the downloads list makes sense. I'm not sure why it gets disconnected/reconnected currently.

For #1162 are you actively working on that? I don't want to duplicate your efforts. I put the stuff I was playing around with in https://github.com/blushingpenguin/gpodder/tree/pause_resume (it's unfinished!) in case that helps.

auouymous:
It would be a couple weeks before I could get to #1162, so feel free to work on it.

auouymous:
Regarding your pause_resume branch, do the menu items not update quickly because https://github.com/blushingpenguin/gpodder/commit/08c3f110bd8cc092fa5c8041b329601f84418b0f#diff-d613a1eb3f5e3e0fba9c6219fc0e8048f88b23de799b5c6f20a30628067529e2R461 isn't also checking the PAUSING state?

auouymous:
@blushingpenguin Changing the line in my last comment to `return task.status in (task.PAUSED, task.PAUSING)` updates menu and toolbar instantly.

play_or_download() needs to return an additional False.
```patch
-             return (False, False, False, False, False)
+             return (False, False, False, False, False, False)
```

The can_pause would be better between can_download and can_cancel since that is how it appears in the menu.

The patch appears to be working fine after removing the changes in `download.py` and `gtkui/model.py`, and fixing the two issues above. It only needs to replace the download button with a resume button, and add pause/resume buttons to the toolbar.

PR #1163, opened 2021-09-14
Disable `Download with Youtube-DL` menu item while downloading episode.

auouymous:
The menu item will still appear for a selection if an episode is selected that isn't downloaded or downloading. Clicking it will resume paused downloads, but most likely does so using the original downloader.


elelay:
[untested] if multiple episodes are selected, comprising eg 1 paused and 1 never downloaded, it will resume the paused with the original downloader and download the never downloaded with yt-dl, right? Is it ok? Or should the paused download be aborted and a new one started using yt-dl?

auouymous:
It probably shouldn't change downloaders unless canceled because you might accidentally select a paused or downloading episode when wanting to queue some more. If paused episodes are aborted, downloading episodes should also be aborted. And auto-canceling the downloading or paused episode would undo any progress it had. Ideally it wouldn't resume paused downloads, and downloading downloads should behave as they currently do.

But this patch is flawed. The other pause PRs get rid of the download menu item for paused episodes. And a selection with a downloading or paused episode does not enable the download menu item. Therefore, youtube-dl should not display its menu item if the selection contains a downloading or paused item? Then there is no confusion over what happens to downloading or paused episodes in a selection.

elelay:
> . Therefore, youtube-dl should not display its menu item if the selection contains a downloading or paused item? Then there is no confusion over what happens to downloading or paused episodes in a selection.

Agreed.
Thanks,

PR #1172, opened 2021-10-12
Remove deprecated GTK features from episodeselector

tpikonen:
Split from #1129 with additional fixes.


auouymous:
I know about downloading episodes, deleting episodes and deleting channels. What else uses this that shows the `OK` and `Remove` buttons?

Should OK, Remove, Download, Mark as old, Select all and Select none have mneumonics? Some of them have use-underline set but have no underscore in the label.

PR #1173, opened 2021-10-13
Scroll to selected episode when opening shownotes.

auouymous:
This prevents the selected episode from being hidden behind shownotes when double clicking or pressing Enter. Only the first episode in a selection of many is scrolled into view when shownotes are opened from a menu.


auouymous:
I'm not sure why, but double clicking the last three episodes (varies with shownote height) in list scrolls to the episode correctly but then some how unselects it. Pressing `Enter` and double clicking with `xte` generated events works fine, it only fails when a real mouse is used to open shownotes.

PR #1177, opened 2021-10-15
Soundcloud support

StopMotionCuber:


src/gpodder/plugins/soundcloud.py
```diff
@@ -38,12 +38,39 @@
...
+        response = util.urlopen("https://api.soundcloud.com/oauth2/token", method="POST", data={
+            "grant_type": "client_credentials",
+            "client_id": CLIENT_ID,
+            "client_secret": CLIENT_SECRET,
+        })
+        if response.status_code != 200:
+            raise SoundcloudException("Authentication not possible")
+        content = response.json()
+        access_token = content["access_token"]
+        token_expire = datetime.now() + timedelta(seconds=content["expires_in"] - 30)
```
StopMotionCuber, line 67: Here I'm decreasing the grace period by 30 seconds (It's 1h by default) to somewhat reduce racyness here and not make requests with expired tokens ever. I also thought of unsetting the access token and recursively calling this function, but that might have the potential for infinite recursion if something goes wrong

src/gpodder/util.py
```diff
@@ -1234,7 +1234,7 @@ def url_add_authentication(url, username, password):
     return urllib.parse.urlunsplit(url_parts)
 
 
-def urlopen(url, headers=None, data=None, timeout=None, **kwargs):
+def urlopen(url, headers=None, data=None, timeout=None, method="GET", **kwargs):
```
StopMotionCuber, line 1237: Needed to make a POST request, so I added this functionality to urlopen. Should be non-breaking, as the parameter is added

src/gpodder/plugins/soundcloud.py
```diff
@@ -38,12 +38,39 @@
...
+token_expire = datetime.now()
 
 logger = logging.getLogger(__name__)
 
 
+class SoundcloudException(Exception):
+    pass
+
+
+def urlopen_with_token(url):
```
StopMotionCuber, line 54: In theory, this method could have more parameters, but as all occurrences only use 1 parameter, I didn't bother to implement more, unused parameters, that are passed to urlopen then

auouymous:
What needs to be done to get credentials?

auouymous:
We are working on getting credentials...

This PR fixes the oauth issue but there are a couple other API changes that will need to be fixed once we have working credentials. Fixing them now would be wasted effort if we can't get credentials.

auouymous:
Soundcloud doesn't want to give us oauth credentials. They either don't understand their API changes or there is another way to make our consumer key work.

> Please note that our API support is limited currently.
> 
> Could you elaborate why you would need a new key and how that would help you?
> 
> The current key looks to be working fine and I fail to understand why a new one would help. Are you looking to add a redirect URI to the application?

StopMotionCuber:
> Soundcloud doesn't want to give us oauth credentials. They either don't understand their API changes or there is another way to make our consumer key work.
> 
> > Please note that our API support is limited currently.
> > Could you elaborate why you would need a new key and how that would help you?
> > The current key looks to be working fine and I fail to understand why a new one would help. Are you looking to add a redirect URI to the application?

You may want to ask them to supply reset the secret to that key, as keeping that key would be fine

auouymous:
> You may want to ask them to reset the secret to that key, as keeping that key would be fine

How would that help with oauth?

StopMotionCuber:
> > You may want to ask them to reset the secret to that key, as keeping that key would be fine
> 
> How would that help with oauth?

https://developers.soundcloud.com/docs#client-creds
There is the client credentials flow, which would be the thing we would be using here. I think we are only using public information, which would mean that this flow would be sufficient. I'm not too familiar with OAuth, but I think this is not an OAuth authentication flow...?

StopMotionCuber:
Reading on, I noticed their API rate limit on the amount of tokens given out, which would be 50 tokens/12h, which seems like a pretty small number of tokens.
I didn't implement the refresh flow, which would be definitely needed with such a low limit (the refresh token should also be cached across restarts).
All in all that would result in ~3000 new gPodder soundcloud users per month (with full utilization of the token), which seems like a not-so-big number for a public application. Still, soundcloud usage might be a niche feature of gPodder so I think it would be still worth implementing that.

All of that, of course, only holds under the circumstances that we obtain a client secret to work with

Lexora:
whatever happened with the soundcloud integration work? did ya guys end up getting updated key/ creds? i'd be interested to make updates to the code to make this work :)

auouymous:
Nope, no working credentials. We might have to scrape the soundcloud page, if that is possible.

Lexora:
[check out this npm scraper ](https://github.com/discord-player/soundcloud-scraper )
so would the process/options be to: 
a) integrate dynamic/on-the-fly scracping functionality into g-podder? (which might include playing songs) 
b) scrape just podcast listings and add these to some sort of database ( is https://gpodder.net/ even still maintained?) 
c) scrape listings and host some service/api and point gpodder to it?

auouymous:
There is no RSS feed for soundcloud channels, gpodder has a plugin that reads episodes from the soundcloud API and basically creates a dynamic feed. The scraping would replicate that functionality to create a dynamic feed for each subscribed channel. It appears to use AJAX to load the list in chunks, so gpodder should probably only grab the first chunk to avoid a lot of requests.

StopMotionCuber:
I would rather suggest to reverse the API that is also used for the website including token creation (should be possible, as that is also what is done on the website) instead of doing HTML scraping

tpikonen:
There is https://getrssfeed.com/ which gives an RSS feed hosted in soundcloud itself. It probably does not work for all the content in souncloud, but for most [podcasts](https://soundcloud.com/tags/podcast) it gives a working RSS URL.

tpikonen:
[rss-bridge](https://github.com/RSS-Bridge/rss-bridge/blob/master/bridges/SoundcloudBridge.php) has soundcloud support, which seems to work ok ([example](https://rss.nixnet.services/?action=display&bridge=Soundcloud&u=richiehawtin&t=tracks&format=Html)).

StopMotionCuber:
Thanks everyone for your inputs. I have implemented a workflow that fetches a `client_id` similarly to rss-bridge.

It's currently rather slow, as the API is returning a URL that needs to be fetched to get a transient content URL, which takes around 200ms per playlist item. It also needs to HEAD that URL to get the file size, as that's included within the metadata.

Async requests might be an option to speed things up here by a fair amount, but that would need another library for async requests. Also, Python 3.5 async development is PITA (I think), would be nice to raise the required python version to 3.7 or something

StopMotionCuber:
Also I just noticed that I have used f-strings, which are Python 3.6, so either
 - You raise the required version to 3.6
 - I'll replace them with good old `.format()`

tpikonen:
> here! here! it's time to ditch 3.5 support :P

I also would not mind, but I have no idea what this means in terms of platform support. Maybe after the next release?

christofdamian:
The most recent change works pretty good for most of my feeds. 
Only few error out with this
```
1650871983.871987 [gpodder.gtkui.main] ERROR: Error: 'str' object is not callable
  File "/home/cdamian/src/gpodder/src/gpodder/gtkui/main.py", line 2712, in update_feed_cache_proc
    channel.update(max_episodes=self.config.max_episodes_per_feed)
  File "/home/cdamian/src/gpodder/src/gpodder/model.py", line 1178, in update
    self._consume_updated_feed(result.feed, max_episodes)
  File "/home/cdamian/src/gpodder/src/gpodder/model.py", line 1088, in _consume_updated_feed
    new_episodes, seen_guids = feed.get_new_episodes(self, existing_guids)
  File "/home/cdamian/src/gpodder/src/gpodder/plugins/soundcloud.py", line 258, in get_new_episodes
    return self._get_new_episodes(channel, existing_guids, 'tracks')
  File "/home/cdamian/src/gpodder/src/gpodder/plugins/soundcloud.py", line 266, in _get_new_episodes
    tracks = list(self.sc_user.get_tracks(track_type))
  File "/home/cdamian/src/gpodder/src/gpodder/plugins/soundcloud.py", line 213, in get_tracks
    'description': track.get('description') or _('No description available'),
TypeError: 'str' object is not callable
```

Examples (I have no idea why I sometimes use http):
* http://soundcloud.com/pan-pot
* https://soundcloud.com/blackasteroid
* http://soundcloud.com/richiehawtin

StopMotionCuber:
Thanks for the feedback and examples, could reproduce on my end and have fixed the issue

christofdamian:
Seems to work well now. All the feeds are working for me. 
The [Richie Hawtin](http://soundcloud.com/richiehawtin) feed gets some 403s for files, but the rest seems to work.

StopMotionCuber:
> https://soundcloud.com/chris-liebing takes a long time to subscribe

Yeah, I'm aware of that. This is caused by having 2 requests per track. Could be reduced to 1 request when leaving out the file size, but one is needed for download URL. Unless there is some callback for getting a resource URL from a temporary URL (which would be a clean solution here and prevent 403 errors) this just takes a long time.

With `aiohttp` as dependency, multiple requests could be handled asynchronously, allowing for parallel requests and significantly improving

> You didn't change the guid line but https://github.com/gpodder/gpodder/pull/1065 is back when updating the feed, do you know if something else is causing it?

The v1-API doesn't return a `permalink` (only `permalink_url`), but the v2-API does, which is why I assume the id was used before, so I guess this is a one-off error. Let's just go for the `id` then as identifier.

src/gpodder/plugins/soundcloud.py
```diff
@@ -73,12 +82,18 @@ def urlopen_with_token(url):
     else:
         url_without_id = url + "?client_id="
```
auouymous, line 83: This could be `url = url + "?client_id=" + client_id`, or `url = "%s?client_id=%d".format(url, client_id)`, and same for line above it, just with the `&`. The other code in this function can then use `url` instead of all the noisy repeated formatting.

auouymous (changes requested):
I added `print(url)` inside urlopen() and then subscribed to https://soundcloud.com/motik. There are several URLs that get fetched twice in a row, do you know what they are?

```
https://w.soundcloud.com/player/?url=http
https://widget.sndcdn.com/widget-4-b941de79a58a.js
https://widget.sndcdn.com/widget-8-778243b0cfc2.js
https://widget.sndcdn.com/widget-9-96462c4e3388.js
https://soundcloud.com/motik?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://soundcloud.com/motik?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/users/901720?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/users/901720?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/users/901720/tracks?limit=200&client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/users/901720/tracks?limit=200&client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/media/soundcloud:tracks:71552595/514b9f7c-52be-4042-b049-943320b3d07c/stream/progressive?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/media/soundcloud:tracks:71552595/514b9f7c-52be-4042-b049-943320b3d07c/stream/progressive?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://cf-media.sndcdn.com/QX7dz9zhSE2m.128.mp3?Policy=eyJTdGF0ZW1lbnQiOlt7IlJlc291cmNlIjoiKjovL2NmLW1lZGlhLnNuZGNkbi5jb20vUVg3ZHo5emhTRTJtLjEyOC5tcDMqIiwiQ29uZGl0aW9uIjp7IkRhdGVMZXNzVGhhbiI6eyJBV1M6RXBvY2hUaW1lIjoxNjUxMzE0NDM4fX19XX0_&Signature=HjymJF5AjM3WYG4ewJqUqzIT7xuBbcue4WVyr9zjWJZ6AlipAPWfX73UfSj5B~pigB~6NridsfQ1Eebfxcg0xgmIU2sb0cVFSKNWLvSQcupC7NebeO7ZqNTFAgHAHjKn~zuX10SVGoqffuzZD8JgpQARNnH-AIqnuU1sgbSk4Rglu-te7mPFEqO0Af3QDbk4TWvtXtsiAPeyHgptiw5K6s49nvQo1j8-YnemEF41IVYCMZzV1JLq83FBKLvoDy95whzaNsEdDAklQHvVnh~XWdW9k6bVjTgumB2f~6XBGFyuHwoYIssknWGRXaVQITcNzCH4ghs8jjEgXKcULO3XxQ__&Key-Pair-Id=APKAI6TU7MMXM5DG6EPQ
https://api-v2.soundcloud.com/media/soundcloud:tracks:2564505/715990f7-e23a-4691-b78f-4477ee75f79b/stream/progressive?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://api-v2.soundcloud.com/media/soundcloud:tracks:2564505/715990f7-e23a-4691-b78f-4477ee75f79b/stream/progressive?client_id=Iy5e1Ri4GTNgrafaXe4mLpmJLXbXEfBR
https://cf-media.sndcdn.com/f5l5Gl2bmavl.128.mp3?Policy=eyJTdGF0ZW1lbnQiOlt7IlJlc291cmNlIjoiKjovL2NmLW1lZGlhLnNuZGNkbi5jb20vZjVsNUdsMmJtYXZsLjEyOC5tcDMqIiwiQ29uZGl0aW9uIjp7IkRhdGVMZXNzVGhhbiI6eyJBV1M6RXBvY2hUaW1lIjoxNjUxMzEyNjE5fX19XX0_&Signature=bbJwvGwIrVGLo8iiBte8fg4GMeQLyCOEwEs7T717PhAEGYmR1hj3XvC5MBmGqf-W4Yhuai4hTvWi97ZufklLdnr6BYxY3BvsVcdslljjZt5hmeVZiQT4NG7w8MntWhoWRzE1M4uwmS2jPWd-tJAGiw2Cu9ylWRympLz61XerTCDM5VU4OQt5UqouEjU9FqcEFBYDkefgrZKbfjHchp14Wh1hjfQn2cRRK~redVZ2F0o2UuTxyoj9w4qR7pQ2tPZmnxeKIMnntRNXGpHFvARWif2jclN7Zu8SAJx6b6PoQpIzzdHGh5MUckbS8UbGyi-YTSD62feQT3HuYJjasInozA__&Key-Pair-Id=APKAI6TU7MMXM5DG6EPQ
https://i1.sndcdn.com/avatars-nP6nG8fG9PELNOub-fpGppA-large.jpg
```

The get_channel_id_url() in youtube.py fetches an URL to get channel ID, and several other functions call that before fetching another URL. The `from functools import lru_cache` and `@lru_cache(1)` annotation caches the function to prevent back to back calls from re-fetching the URL. Is the soundcloud code doing something that could make use of this caching?

src/gpodder/plugins/soundcloud.py
```diff
@@ -51,15 +52,23 @@ class SoundcloudException(Exception):
 
 
 def get_client_id():
+
     response = util.urlopen("https://w.soundcloud.com/player/?url=http")
+    if not response.ok:
+        raise SoundcloudException("Error when fetching {url}".format(url=response.request.url))
     for result in re.finditer(r"https://widget\.sndcdn\.com/widget-[0-9a-f\-]+\.js", response.text):
         match = re.search(r'client_id.*?\"(.+?)\"', util.urlopen(result.group(0)).text)
```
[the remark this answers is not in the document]
auouymous: I forgot to comment on the other urlopen calls in the code, they too should catch connection errors and throw a soundcloud exception so the user knows what exactly failed.

PR #1179, opened 2021-10-16
Norwegian Bokmål translation updated

comradekingu:
I don't want my translations on Transifex.


auouymous:
I would only merge the translation into the project, but I can't stop someone else from doing so, is that a problem?

comradekingu:
@auouymous Actually it means "current", but that is what we use for ~power in general language, unless talking about power production (when it is "kraft"), and "effekt" for measure of Watt.
Either "informasjonskanal" ~ information channel or "nyhetsflyt" ~flow of information.
"strøm" is not my favourite, and "strømning" is a bit of a lost cause.
Generally I didn't try to change anything that is passable.

auouymous:

PR #1180, opened 2021-10-17
Slovak translation update

neodyne:


po/sk.po
```diff
@@ -2251,10 +2236,8 @@ msgid "Only sync unplayed episodes"
 msgstr "Synchronizovať iba neprehrané epizódy"
 
 #: share/gpodder/ui/gtk/gpodderpreferences.ui.h:32
-#, fuzzy
-#| msgid "Remove episodes deleted on device from gPodder"
 msgid "Remove episodes deleted in gPodder from device"
-msgstr "Odstrániť epizódy vymazané zo zariadenia z gPodderu"
+msgstr "Odstrániť epizódy vymazané z gPodderu zo zariadenia"
```
auouymous, line 2240: Is the `u` after gPodder**u** correct?

neodyne: Yes, it's correct. The `-u` suffix denotes (along with the preposition `z`/`zo`) singular genitive case, masculine gender, in [Slovak declension](https://en.wikipedia.org/wiki/Slovak_declension). 'gPodder' as a word would follow the masculine gender paradigm `dub` (oak) subparadigm `vrch` (peak). Note that the original string had it too; it's present in multiple places in the translation. Alternative way of phrasing would be 'vymazané z program**u** gPodder' - deleted in/from (the program) gPodder - as opposed to 'vymazané z gPodderu' - deleted from/in gPodder. The suffix would be still there, just consumed by a preceding noun (this would also feel a bit awkward to a native speaker, depending on the context).

As a sidenote, I pushed an unrelated minor fix for one string.

auouymous:
Thanks for the translation.

PR #1182, opened 2021-10-18
Use youtube-dl or yt-dlp for the youtube-dl extension.

### LINKED ISSUE #1176: Switch from youtube-dl to yt-dlp?
Youtube-dl hasn't merged any PRs since 1jul2021 and last release was 6jun2021. Age-restricted content hasn't worked for a while, but no one has complained. Both youtube-dl and gpodder's native youtube support are throttled to ~50kbps. It is questionable how long the throttling "fix" will continue working in yt-dlp, so I may or may not try to fix the throttling in gpodder because of that.

The following patch uses yt-dlp if available, and it appears to work with no other changes. So we do have something to fallback to if the youtube API changes and youtube-dl still doesn't update.

```patch
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -10,7 +10,10 @@ import re
 import sys
 import time
 
-import youtube_dl
+try:
+    import yt_dlp as youtube_dl
+except:
+    import youtube_dl
 from youtube_dl.utils import DownloadError, ExtractorError, sanitize_url
 
 import gpodder
```

yt-dlp required half a dozen or so dependencies that would need to be included in the windows and mac builds. Although, one of them was jeepney which may or may not cause issues with the fake dbus.

auouymous:
This changes the youtube-dl extension to use yt-dlp if installed. As of yt-dlp 2021.10.09, the progress bar appears in the console and needs `noprogress` to suppress.

- [x] The mac and windows builds need to install yt-dlp and its dependencies.

Closes #1176.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ PyPI. With this, you get a self-contained gPodder CLI codebase.
 - Clickable links in GTK UI show notes: html5lib
 - HTML show notes: WebKit2 gobject bindings
     (webkit2gtk, webkitgtk4 or gir1.2-webkit2-4.0 packages).
-- Better Youtube support (> 15 entries in feeds, download audio-only): youtube_dl
+- Better Youtube support (> 15 entries in feeds, download audio-only): youtube_dl or yt-dlp
 
 
 ### Build Dependencies
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -10,7 +10,10 @@ import re
 import sys
 import time
 
-import youtube_dl
+try:
+    import yt_dlp as youtube_dl
+except:
+    import youtube_dl
 from youtube_dl.utils import DownloadError, ExtractorError, sanitize_url
 
 import gpodder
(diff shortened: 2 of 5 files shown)


auouymous:
I haven't finished this because yt-dlp has an issue where it correctly writes the whole file to `tempname.ext` but incorrectly writes a little bit of the start to `tempname`. The extension then sees that `tempname` has a non-zero size and uses it instead of finding the actual file at `tempname.ext`. Playback is then cut short and deleting the episode leaves behind the `tempname.ext` file which must be manually removed outside of gpodder.

It has never happened when using yt-dlp outside of gpodder, so either an issue with embedding or the nopart option gpodder passes to it, which is something I don't use outside gpodder. The youtube-dl extension could be altered to always scan for `tempname.ext` files by eliminating the stat and size check lines. Scanning is probably the normal behavior for most downloads, so it shouldn't cause any slowdowns. Does annyone have better ideas or see any issues with this approach?

tpikonen:
The issue with --no-part should be fixed in yt-dlp, or at least reported. But current versions should be supported also.

auouymous:
There are 133 open/closed issues matching 'no-part', is/was there an issue with it that you know of?

It could also be related to using DASH formats which I do in gpodder but rarely outside of it. I haven't filed an issue yet, but cancelling a ytdl/ytdlp download leaves behind the separate video and audio parts. So the youtube-dl extension does need a bit better handling of the files created by ytdl/ytdlp.

tpikonen:
> There are 133 open/closed issues matching 'no-part', is/was there an issue with it that you know of?

Sorry, no. For huge projects like yt-dlp it's usually more productive to just submit a patch rather than a bug report. That's of course easier said than done, and I don't have the time right now.

elelay:
@auouymous 
> The youtube-dl extension could be altered to always scan for tempname.ext

No objection: whatever works with yt-dlp

auouymous:
This should now be ready to merge along with #1219, unless anyone sees something I missed. The only issue I have seen is that yt-dlp dumps a stack trace and three identical messages when it can't download a video without throttling. It would be nice if gpodder could prevent the dump and be notified of the throttling, but it should be fine for now.

PR #1191, opened 2021-11-01
fix #1190 ERROR: Could not create save_dir on windows with gio

### LINKED ISSUE #1190: [gpodder.model] ERROR: Could not create save_dir
Build 41319604 on Windows:

```
1635585349.861915 [gpodder.util] WARNING: Could not create directory C:\Users\df-mi\Documents\gPodder\Downloads\How I Built This with Guy Raz: Vorgang wird nicht unterstützt
1635585349.862913 [gpodder.model] ERROR: Could not create save_dir: C:\Users\df-mi\Documents\gPodder\Downloads\How I Built This with Guy Raz
1635585385.450962 [gpodder.model] WARNING: Unknown external file: C:\Users\df-mi\Documents\gPodder\Downloads\How I Built This with Guy Raz\614d9f37a015be0014a52325.mp3
```

It looks to me like this gio library isn't working properly on Windows. Python3 introduced pathlib which is the pythonic way to interact with filesystems. I don't know if this would solve the problem, but it looks like it to me. Shall I give it a try?

elelay:
`util.new_gio_file` returned a GDummyFile, where every operation failed,
in this case `make_directory_with_parents()`.

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -161,7 +161,8 @@ def is_absolute_url(url):
     """
     try:
         parsed = urllib.parse.urlparse(url)
-        return not not parsed.scheme
+        # fix #1190: when parsing a windows path, scheme=drive_letter, path=\rest_of_path
+        return parsed.scheme and not parsed.path.startswith("\\\\")
     except ValueError:
         return False
 


dreamflasher:
I am testing this, but it's not working. Which type is the method supposed to return?

dreamflasher:
It looks like you just wanna do `return not parsed.path.startswith("\\\\")`

PR #838, opened 2020-07-03
rename youtube-dl partial file when it contains % or $ characters

auouymous:
Fixes #837

PR #1193, opened 2021-11-04
Do not try to rename the outtmpl file in the Youtube-DL extension.

auouymous:
The output file does not contain the escaped percents, resulting in a file not found if renamed.

Youtube-DL also does not use '$' and escaping it is not required. An episode title containing a '$' will produce a file containing two dollar signs, and is then renamed to a single dollar sign. But a channel name containing a dollar sign causes Youtube-DL to create a new channel directory, with two dollar signs, for the output file. This directory remains empty after gpodder moves the output file.

Both cases can be tested by renaming a channel to have '%$' and then downloading episodes with Youtube-DL.

Therefore, dollar signs should not be escaped and renaming outtmpl is not required as it will always match tempname, when assuming percents are unescaped.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -86,14 +86,10 @@ class YoutubeCustomDownload(download.CustomDownload):
         """
         self._reporthook = reporthook
         # outtmpl: use given tempname by DownloadTask
-        # (escape % and $ because outtmpl used as a string template by youtube-dl)
-        outtmpl = tempname.replace('%', '%%').replace('$', '$$')
+        # (escape % because outtmpl used as a string template by youtube-dl)
+        outtmpl = tempname.replace('%', '%%')
         res = self._ytdl.fetch_video(self._url, outtmpl, self._my_hook)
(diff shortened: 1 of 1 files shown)


elelay:
Thanks for the patch.
The rename check seems to be to add an extension when we didn't know what extension to put in the tempname, not so much dealing with % or $.
But I'm not familiar with this code so I'm willing to merge as is :-)

auouymous:
Youtube-dl always creates a second partial file with the extension appended when `ext` is set. This change had to be accounted for when the escaped `outtmpl` was not the same as `tempname`, and the file was renamed. But renaming is no longer required because the actual flow would look something like the following, which does nothing.

```
outtmpl = tempname.replace('%', '%%')  # escape
...download...
outtmpl = outtmpl.replace('%%', '%')  # unescape
if outtmpl != tempname:
    ...rename...
```

The removed renaming code was merely to fix `outtmpl` escaping, and the real renaming (due to `ext` appending) is handled further down in that function. So this patch should be good, unlike the original fix in #838.

elelay:
All clear, thanks!

PR #1201, opened 2021-12-22
Open error dialog if user tries to use not-installed xdg-open

### LINKED ISSUE #1003: Check for xdg-open or player command before running it.
If `xdg-open` is not installed, a fresh gpodder instance will fail to play episodes and not provide a reason why, unless the command line is visible or logs are viewed. It should check for the existence of `xdg-open`, or any command the user has supplied, and display a dialog if not found.

```
  File "/home/test/gpodder/src/gpodder/util.py", line 1433, in gui_open
    Popen(['xdg-open', filename], close_fds=True)
  File "/home/test/gpodder/src/gpodder/util.py", line 2019, in __init__
    super(Popen, self).__init__(*args, **kwargs)  # Python 2 syntax
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1706, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
PermissionError: [Errno 13] Permission denied: 'xdg-open'
```

bwildenhain:
Currently trying to use default player xdg-open when it's not installed
will only cause messages on the console, which might not be visible to
the user. This PR creates an alert box in this case.

Fixes #1003.

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1470,7 +1470,7 @@ def http_request(url, method='HEAD'):
     return conn.getresponse()
 
 
-def gui_open(filename):
+def gui_open(filename, gui=None):
     """
     Open a file or folder with the default application set
     by the Desktop environment. This uses "xdg-open" on all
@@ -1481,13 +1481,30 @@ def gui_open(filename):
     try:
         if gpodder.ui.win32:
             os.startfile(filename)
+            opener = None
         elif gpodder.ui.osx:
-            Popen(['open', filename], close_fds=True)
+            opener = 'open'
         else:
-            Popen(['xdg-open', filename], close_fds=True)
+            opener = 'xdg-open'
+
+        if opener:
+            opener_fullpath = shutil.which(opener)
+            if opener_fullpath is None:
+                raise Exception((_("System default program '%(opener)s' not found"))
+                    % {'opener': opener}
+                )
+            Popen([opener_fullpath, filename], close_fds=True)
         return True
     except:
         logger.error('Cannot open file/folder: "%s"', filename, exc_info=True)
+        if gui is not None:
+            if opener is None:
+                message = _("Cannot open file/folder '%(filename)s' using default program") % {'filename': filename}
+            else:
+                message = _("Cannot open '%(filename)s' using '%(opener)s'") \
+                                        % {'filename': filename, 'opener': opener}
+            gui.show_message_details(_('Cannot open file/folder'),
+                    str(sys.exc_info()[1]), message)
         return False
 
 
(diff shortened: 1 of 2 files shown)


auouymous:
Thank you for working on this.

It might be better to pass `self` to `gui_open` via an optional `gui?` parameter, and open the dialog there. For osx and linux it could check for open or xdg-open, perhaps by storing the open or xdg-open in a variable and set it to None for windows. If the command is found or the command is None (windows), it could use `gui.show_message_details()` and display the exception in the details box to handle other errors. That way it works for all platforms, handles errors besides command not found, and works for the other `gui_open` call in `on_open_download_folder`.

However, the other call isn't trying to play a file, and would need a different error message. Unless `gui_open` doesn't check if the program exists and simply calls `show_message_details`.

src/gpodder/util.py
```diff
@@ -1480,14 +1480,32 @@ def gui_open(filename):
...
-            os.startfile(filename)
+            opener = None
         elif gpodder.ui.osx:
-            Popen(['open', filename], close_fds=True)
+            opener = 'open'
         else:
-            Popen(['xdg-open', filename], close_fds=True)
+            opener = 'xdg-open'
+
+        if opener is None:
```
auouymous, line 1489: It would be better to leave startfile in the windows section above in case additional branches are added above that set opener=None. This section should only handle `if opener:`.

src/gpodder/util.py
```diff
@@ -1480,14 +1480,32 @@ def gui_open(filename):
...
+            os.startfile(filename)
+        else:
+            opener_fullpath = shutil.which(opener)
+            if opener_fullpath is None:
+                raise Exception((_("System default player '%(opener)s' not found"))
```
auouymous, line 1494: Should be `default program` like below, since the directory opener isn't trying to play a file.

src/gpodder/util.py
```diff
@@ -1480,14 +1480,32 @@ def gui_open(filename):
...
         return True
     except:
         logger.error('Cannot open file/folder: "%s"', filename, exc_info=True)
+        if gui is not None:
+            if opener is None:
+                message = _("Cannot open file/folder '%(filename)s' using default program") % {'filename': filename}
+            else:
+                message = _("Cannot open '%(filename)s' using '%(opener)s'") \
+                                        % {'filename': filename, 'opener': opener}
+                gui.show_message_details(_('Cannot open file/folder'),
```
auouymous, line 1507: These two lines need to be dedented to handle both message branches.

src/gpodder/util.py
bwildenhain: Indead, that makes sense

src/gpodder/util.py
bwildenhain: Done

src/gpodder/util.py
bwildenhain: Of course, I somehow overlooked this.

auouymous:
Thank you for fixing this.

PR #1204, opened 2021-12-31
Play from Sonos library (not the URL)

lingfish:
This adds new context entries to search for, and play, episodes from the Sonos library (instead of a URL).

I thought it was odd that gPodder downloads episodes, yet the excellent Sonos extension can't use them.

This PR assumes a few things:

* The podcast publisher sets metadata well
* The user of gPodder has added their downloads location to the Sonos music library
* Each episode title is unique (or, it'll just pick the first one)

There are a few ways I guess to present the choices in the UI, this is what I went with.

It basically works by searching for the gPodder episode title in the "tracks" part of the Sonos library (hence the need for good metadata).  It would be great to use, say, `self.local_filename`, but I don't think Soco can use/see/get that from Sonos.

### DIFF
--- a/share/gpodder/extensions/sonos.py
+++ b/share/gpodder/extensions/sonos.py
@@ -49,7 +49,7 @@ class gPodderExtension:
             if name:
                 self.speakers[uid] = speaker
 
-    def _stream_to_speaker(self, speaker_uid, episodes):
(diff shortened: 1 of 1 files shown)


lingfish:
Hrm, need to do a little more work on this, will push updates.

luzpaz:
Any more traction on this ?

PR #1209, opened 2022-01-07
without the ActivateTab, no progress bar

arne182:
https://stackoverflow.com/questions/17607415/python-tkinter-windows-7-taskbar-progress

### DIFF
--- a/share/gpodder/extensions/taskbar_progress.py
+++ b/share/gpodder/extensions/taskbar_progress.py
@@ -138,8 +138,8 @@ class ITaskbarList3(ITaskbarList2):
                   (['in'], POINTER(tagRECT), 'prcClip'))]
 
 
-assert sizeof(tagTHUMBBUTTON) == 540, sizeof(tagTHUMBBUTTON)
-assert alignment(tagTHUMBBUTTON) == 4, alignment(tagTHUMBBUTTON)
+assert sizeof(tagTHUMBBUTTON) in [540, 552], sizeof(tagTHUMBBUTTON)
+assert alignment(tagTHUMBBUTTON) in [4, 8], alignment(tagTHUMBBUTTON)
 
 
 # based on http://stackoverflow.com/a/1744503/905256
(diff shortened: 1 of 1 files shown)


elelay:
Thanks for the patch :-)

Testing it I got another fatal error in the extension: missing `import ctypes`! Is it an issue you encounter as well?

Here is the stack trace:
````
1641748031.113040 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "C:\gpodder-3.10.21-portable\data\share\gpodder\extensions\taskbar_progress.py", line 165, in callback
    ctypes.pythonapi.PyCapsule_GetPointer.restype = ctypes.c_void_p
NameError: name 'ctypes' is not defined

  File "C:\gpodder-3.10.21-portable\data\share\gpodder\extensions\taskbar_progress.py", line 165, in callback
    ctypes.pythonapi.PyCapsule_GetPointer.restype = ctypes.c_void_p
NameError: name 'ctypes' is not defined
````

I've fixed it in gPodder master branch [a moment ago](https://github.com/gpodder/gpodder/commit/b664e7912987437617d7303f07416b9d5fc59ccf).

Now, regarding your patch, I get an error on the ActivateTab invocation because in `on_load()` the window_handle is not registered yet (`on_load()` is called first, then `on_ui_object_available()`, then the `callback` function as verified by running in a `cmd` and adding `print` statements). How comes it works on your machine? What version are your running from?

Also on windows 10 I already got the progress in taskbar working without your patch.
Adding `self.taskbar.ActivateTab(self.window_handle)` in the callback, after setting `self.window_handle` doesn't result in an error on my side, though. So I'd be ok with adding it you show it's needed for you.

arne182:
This is untested code. Just used this for another project. Putting it in the callback is fine and adding the different lengths for 64 bit so the asset does not fail would be cool

elelay:
OK, I'll bring it in when I have time to test it

elelay:
Closed in favor of #1211

PR #1211, opened 2022-01-15
without the ActivateTab, no progress bar (extended)

elelay:
Further work on #1209 
Still to be tested before merge

### DIFF
--- a/share/gpodder/extensions/taskbar_progress.py
+++ b/share/gpodder/extensions/taskbar_progress.py
@@ -31,6 +31,10 @@ from comtypes import COMMETHOD, GUID, IUnknown, client, wireHWND
 
 import gpodder
 
+import gi  # isort:skip
+from gi.repository import Gtk  # isort:skip
+
+
 _ = gpodder.gettext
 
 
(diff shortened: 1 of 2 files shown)


arne182:
Excellent. Thanks for integrating my change.

auouymous:
Code looks fine. I tested on a Win10 machine today and I don't think it works. I enabled the extension and downloaded an episode but nothing happened on the taskbar.

elelay:
> I enabled the extension and downloaded an episode but nothing happened on the taskbar.
do you use a "classic" theme? It doesn't show with it but with modern theme.
Also gPodder might need to restart before it works

auouymous:
By classic do you mean what win95 looked like? It wasn't my computer so I don't know what theme it had, but its taskbar looked like https://i.insider.com/55b7aebbdd08958f7f8b4600?width=2160 (without that white search box). I closed and killed gpodder (in task manager) at least a dozen times, so pretty sure it was started with the extension.

If it works for you on win10 then it should be good to go, because that machine is heavily configured and maybe something is disabled.

elelay:
> By classic do you mean what win95 looked like? It wasn't my computer so I don't know what theme it had, but its taskbar looked like https://i.insider.com/55b7aebbdd08958f7f8b4600?width=2160 (without that white search box). I closed and killed gpodder (in task manager) at least a dozen times, so pretty sure it was started with the extension.
> 

yes it is the same theme I have. No idea why it didn't work on this system...

PR #1214, opened 2022-01-25
Show 'Syncing' in progress tab instead of 'Downloading' when syncing.

auouymous:
Would it be better to show `Syncing` in the progress tab instead of `Downloading` when syncing files, to distinguish them from files actually being downloaded from feeds?


elelay:
yes!

elelay:
Merged, thanks

PR #1219, opened 2022-02-02
YoutubeDL partial file fixes

auouymous:
The first patch fixes an issue caused by yt-dlp that results in a truncated download and a partial file containing the full download which must be manually renamed or deleted outside of gPodder.

The second patch cleans up YoutubeDL partial files after cancelling a download.


elelay:
LGTM :+1:

PR #1222, opened 2022-02-04
gPodderExportToLocalFolder: Use GtkDialog.add_buttons()

tpikonen:
Currently the gPodderExportToLocalFolder dialog adds the OK and Cancel buttons to the action area explicitly in the ui-file. This causes Gtk warnings on systems which are configured to have dialog the buttons in the window headerbar.

The correct way is to add the buttons with GtkDialog.add_buttons() method, which places the buttons either to the header bar or the action area depending on system settings.

Also fix strings in the 'File already exists' confirmation dialog.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderexporttolocalfolder.ui
+++ b/share/gpodder/ui/gtk/gpodderexporttolocalfolder.ui
@@ -15,7 +15,7 @@
     <property name="window-position">center-on-parent</property>
     <property name="type-hint">dialog</property>
     <property name="action">save</property>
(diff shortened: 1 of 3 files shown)


auouymous:
LGTM, but the `Save` button changed to `OK`, was that intentional or a mistake? The `save page as` dialog in Firefox uses `Cancel` and `Save` buttons.

tpikonen:
That was a pure mistake. Fixed now.

auouymous:
Thanks for the fix.

PR #1223, opened 2022-02-04
Translation update (sk)

neodyne:
Fixes changed new/fuzzy strings in the Slovak translation of gPodder.


neodyne:
Looking at the added typo fixes in #1222, corresponding translations will probably get marked as `fuzzy` after rerunning `make messages`.  The translation in Slovak is correct for both strings even with the fix, so any resulting `fuzzy` markings can be removed.

auouymous:
Thank you for the translation, and thanks for the heads up.

PR #1228, opened 2022-02-08
Create partial file when reusing cancelled tasks.

auouymous:
The partial file for a cancelled task is deleted, but the task is not removed from progress list while other tasks are downloading. Downloading the cancelled episode again would reuse the existing task without creating a new partial file. This prevented resuming the download after restarting gPodder, and caused the youtube-dl extension to fail if used to re-download it.


elelay:
LGTM

Issue #1234, opened 2022-03-02
NotFound error when trying sync with email instead of user name


JuanCanham:
When logging in with email rather than username, login competes but subsequent operations will fail

tpikonen:
Did #1235 fix this?

auouymous:
Please re-open if #1235 didn't fix this.

PR #1235, opened 2022-03-02
Make error message more verbose on NotFound sync error

JuanCanham:
Quick fix, 

I think ideally we would handle this when credentials are entered but the UI, but the UI doesn't seem to validated them.

#1234


elelay:
Should the message start with "Server doesn't recognize your mygpo account" or similar, and then offer the advice about username not email?

auouymous:
The message should probably also be a single translatable string instead of two.

auouymous:
The message could use some punctuation and tweaks.

```
Could not find your device.

Check login is a username (not email)
and the device name matches one in your account.
```

We can run `make messages` after merging, so anything you didn't manually modify shouldn't be included in the commit.

auouymous:
I just tested this and the """ quoting not only indents the strings in translations but also on the dialog. It would be better to wrap each line in a string.

```
                    message = _(
                        'Could not find your device.\n'
                        '\n'
                        'Check login is a username (not an email)\n'
                        'and that the device name matches one in your account.'
                    )
```

I have never used gpodder.net, do you know if this message happens anywhere other than when clicking the replace on server button on the gpodder.net preferences page?

JuanCanham:
This is only for error handling on the set_subscriptions api call, and only an email / bad deviceid can cause this

auouymous:
Thank you.

PR #1248, opened 2022-03-16
Allow failed downloads to be cancelled from episodes list.

auouymous:
This not only removes the failed download from the progress list, but also removes the error icon from the episode list.


elelay:
Merged, thanks

PR #1249, opened 2022-03-16
Set episode toolbar and menu actions.

auouymous:
Items in the Episode menu were not disabled when the progess tab was opened. This consolidates the enabling and disabling of buttons in the toolbar and items in the Episodes menu, and fixes that issue.

It also changes the label and icon for the toolbar play button to be Open, Play, Preview or Stream.

I haven't yet figured out how, but if possible, this new function could change the Play menu item label as well, and no longer need the Open menu item.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -1966,7 +1966,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
             menu = Gtk.Menu()
 
-            (can_play, can_download, can_pause, can_cancel, can_delete, open_instead_of_play) = self.play_or_download()
+            (open_instead_of_play, can_play, can_download, can_pause, can_cancel, can_delete) = self.play_or_download()
 
             if open_instead_of_play:
                 item = Gtk.ImageMenuItem(_('Open'))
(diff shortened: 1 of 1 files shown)


elelay:
Good!

Still a minor bug when
1. selecting an episode
2. choosing download (then the CANCEL action and tool button are enabled),
3. switching to the progress tab
4. expected: both menu item and button are disabled
5. actual: only the menu item is disabled
6. clicking the tool button has no effect, but it's not the bug I'm concerned with

auouymous:
The cancel button is disabled when opening progress tab, but it then gets turned back on each time a task updates. Switching back and forth turns it off each time progress tab is opened, and the following code is what causes it.

```
    def play_or_download(self, current_page=None):
        if current_page is None:
            current_page = self.wNotebook.get_current_page()
        if current_page > 0:
            self.toolCancel.set_sensitive(True)
            return (False,) * 6
```

on_item_cancel_download_activate() has code to cancel the selected task in progress tab. So play_or_download() and on_wNotebook_switch_page() need to check for a selected task and toggle the cancel button to the correct state using set_episode_actions() so the episode menu and ^C can also cancel tasks. And a future PR could enable the download and pause buttons to resume and pause selected tasks.

elelay:
So, nothing to fix within current PR then.
Thanks for the investigation and you're welcome to a fix :+1:

PR #1252, opened 2022-03-20
Refactor play_or_download() episode action code.

### LINKED ISSUE #1250: Failed downloads can not be re-downloaded or deleted.
#1248 allowed failed downloads to be cancelled but broke re-downloading and deleting for them. I am working on a fix that replaces the code in play_or_download() with something that is easier to reason about.

auouymous:
Streaming was always possible (#867) for all audio and video episodes if an audio player was configured. Now it is only possible if a player is configured for the episode's type.

Methods have been added to PodcastEpisode that are either now used by actions or replicate the conditions used by actions to filter selections. This will help prevent mismatched conditions between play_or_download() and episode actions. And play_or_download() is easier to reason about and should prevent future bugs such as #1250.

Fixes #1250.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2155,42 +2155,20 @@ class gPodder(BuilderWidget, dbus.service.Object):
         self.update_podcast_list_model(set(e.channel.url for e in episodes))
         self.db.commit()
 
-    def episode_player(self, episode):
-        file_type = episode.file_type()
-        if file_type == 'video' and self.config.player.video \
-                and self.config.player.video != 'default':
-            player = self.config.player.video
-        elif file_type == 'audio' and self.config.player.audio \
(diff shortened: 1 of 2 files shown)


elelay (approved):
LGTM

elelay:
Still an issue with the Cancel button when the Progress tab is active after cancelling a task:
1. start downloading
2. go to progress tab
3. select downloading task
4. click cancel
5. task shown as cancelled in the list
6. Cancel is not disabled

auouymous:
This PR is only for the episode list actions. I have a partial fix for the progress tab cancel button/menutiem but needed to shelve it while fixing my mistake from #1250. There is also an issue with the Open/Play/Preview/Stream button/menuitems when using a mixed selection that will need to be dealt with later.

elelay:
ok, so merge it when you want.
Thanks,

PR #1256, opened 2022-03-24
Allow ctrl-C to cancel episode downloads.

auouymous:
^C is unused and represents cancel in most apps.

Many of the accelerators don't make much sense (^L to subscribe), or are bound to seldomly used actions, while common actions (download, pause and lock) lack one. Should we make ^V download (looks like a down pointing arrow, and ^D toggles description which I don't understand why that needs an accelerator)? ^Z could pause because 'zzz' for sleep, and ^P and ^S are already taken. Lock could use ^A which makes sense for 'archive', but I think there was discussion about one day renaming it to lock everywhere in the UI. Thoughts?

### DIFF
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -123,6 +123,7 @@
         <item>
           <attribute name="action">win.cancel</attribute>
           <attribute name="label" translatable="yes">Cancel</attribute>
+          <attribute name="accel">&lt;Primary&gt;c</attribute>
         </item>
         <item>
           <attribute name="action">win.delete</attribute>


thp:
^L = "location", as it's also used in most browsers to focus the URL bar to enter a new URL to go to, so this is where this comes from ;)

^V is also "paste" in other apps, not sure if we want to overload this one
^C is "copy", but maybe it makes sense to be "cancel" in this case

The "menu button" () can be used to bring up the pop-up menu (I hope, haven't actually checked), so maybe that'd be an easier and more visual way of accessing the UI, and then have accelerator keys on the pop-up menu (again, not sure if Gtk supports this).

I personally wouldn't mind if somebody added a "vi keybindings" mode to the GUI, you could do more advanced commands prefixed with ":", and some commands (e.g. "dd" to delete the current episode). We would probably have to move the "start to type to filter" feature to e.g. "/", but this is also used by e.g. the GMail web interface (and also the Github web interface here). GMail and Github's web UI also bring up keyboard shortcuts using "?", if we have many fancy keyboard shortcuts, bringing that up when pressing "?" might be a nice touch.

auouymous:
^L makes more sense now, and thanks for the reminder on ^C/^V for copy/paste, using either for cancel/download prevents copy/paste in the find fields.

The menu key opens the episode list context menu, and arrows can be used to select an item. Adding mnemonics to the context menu would allow menu+mnemonic to activate the menu item (for those with a menu key). Alt+E opens the Episodes menu and adding mnemonics there would allow alt+e+mnemonic to activate those menu items. But not as useful as control combos for common actions.

I have been trying to modify the accelerators programmatically to allow the user to configure them since gtk3 no longer allows that. But a vi-mode might be an easier task, and would be more convenient than control combos. The / is used in many console apps to search, and would be so much nicer than the annoying find-as-you-type. However, the vi bindings are less discoverable than control combos. Maybe a menu item, in addition to ?, to show them (assuming that is what you meant by keyboard shortcuts).

PR #1259, opened 2022-03-25
Remove 'from youtube_dl' imports

tpikonen:
I removed the `youtube_dl` module from my system (but left `yt_dlp`) and got an exception in the youtube-dl extension when running gPodder master:

```
  File "/home/tpikonen/src/gpodder/master-worktree/share/gpodder/extensions/youtube-dl.py", line 17, in <module>
    from youtube_dl.utils import DownloadError, ExtractorError, sanitize_url
``` 

This patch fixes the import error.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -14,7 +14,6 @@ try:
     import yt_dlp as youtube_dl
 except:
     import youtube_dl
-from youtube_dl.utils import DownloadError, ExtractorError, sanitize_url
 
 import gpodder
 from gpodder import download, feedcore, model, registry, youtube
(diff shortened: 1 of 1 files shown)


auouymous:
I can confirm that uninstalling youtube-dl throws a 'youtube_dl module not found' error. But this patch fails when yt-dlp is not installed because youtube-dl stores them in the utils module.

This seems to work for both cases.

```python
try:
    import yt_dlp as youtube_dl
    from yt_dlp.utils import DownloadError, ExtractorError, sanitize_url
except:
    import youtube_dl
    from youtube_dl.utils import DownloadError, ExtractorError, sanitize_url
```

But simply inserting `.utils` into your patch works and eliminates the duplicate imports. So maybe `youtube_dl.utils.*` instead of `youtube_dl.*`?

tpikonen:
Thanks for testing, the exceptions actually are also in `youtube_dl.*` namespace, but `sanitize_url` is only in `youtube_dl.utils.*`.

I changed the patch to use `youtube_dl.utils.*` for both.

auouymous:
Thanks for fixing this.

PR #1262, opened 2022-03-30
youtube_dl.is_supported_url(): Don't fail on None

tpikonen:
RSS episodes which have no media and have an empty or non-existing link element cause an exception in youtube_dl.is_supported_url(). In this case None is passed as the url argument.

This patch makes is_supported_url() return False when url is None.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -399,6 +399,8 @@ class gPodderYoutubeDL(download.CustomDownloader):
         return None
 
     def is_supported_url(self, url):
+        if url is None:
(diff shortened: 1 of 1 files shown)


tpikonen:
Fixed the class name in the commit message.

tpikonen:
See [these lines](https://github.com/gpodder/gpodder/blob/8e7c8f1448408d7173bdc62154316079608b064b/src/gpodder/model.py#L309) in PodcastEpisode.from_podcastparser_entry(). The url is temporarily switched to the link when testing the link for compatible media.

auouymous:
Ah, thanks for the fix.

PR #1265, opened 2022-04-04
Fix exceptions on mpris-listener extension

tpikonen:
Some media players (notably firefox on linux) do not support / have the 'Position' property, which makes the MPRIS extension throw an exception when pause is toggled.

This patch makes the MPRIS property query functions return None when the queried properties do not exist.

Also, fix exceptions caused by log string formats, when the properties are None.


auouymous:
LGTM, thanks for the fix.

PR #1266, opened 2022-04-07
Turkish translation update

serkan-maker:


po/tr.po
```diff
@@ -1513,26 +1497,20 @@ msgid "Remove"
...
+msgstr "Tümünü _seç"
 
 #: src/gpodder/gtkui/desktop/episodeselector.py:286
 #: share/gpodder/ui/gtk/gpodderepisodeselector.ui.h:6
 #: share/gpodder/ui/gtk/gpodderpodcastdirectory.ui.h:5
-#, fuzzy
-#| msgid "Select none"
 msgid "Select _none"
-msgstr "Hiçbirini seçme"
+msgstr "Hiçbirini _seçme"
```
auouymous, line 1513: > msgstr "Tümünü _seç"
> msgstr "Hiçbirini _seçme"

The letter after the underscore must be unique on the window where used. This would try to bind Alt+S to both actions. That dialog has `Select _all`, `Select _none`, `_Mark as old`, `_Cancel` and `_Download`. So `_Hiçbirini seçme` and `_Tümünü seç` would be better choices.

auouymous (changes requested):

po/tr.po
serkan-maker: Thank you for the warning. I corrected the relevant line **Seçimleri _kaldır**

PR #1261, opened 2022-03-28
preferences: Use GtkStack instead of GtkNotebook

tpikonen:
Replace GtkNotebook in the preferences dialog with a GtkStack and a vertical GtkStackSwitcher.

This results in a (IMO) cleaner look for the preferences dialog.

If this PR is accepted, I can port (and fix) the changes from adaptive branch which make the individual preference tab pages narrower to master. After that, the delta between adaptive and master (for the preferences dialog) will be reduced to placing the StackSwitcher to a foldable HdyFlap widget.

Issue #1263, opened 2022-03-31
Spreaker feeds not working - [EDIT - youtube-DL extension is the culprit]


joesmith222:
any chance to fix it?

it works on another app i have called Tapin Radio fwiw.

thx

mhussaincov93:
hi,
could you give us a feed for testing?
this will enable us to check what errors are spawned.
could you conferm:
does this feed function in your browser?
what version of gpodder are you rocking?
are you running the gui /gtk ui or the gpo cli interface?
a thought could this be a user agent issue?
post us a feed and we'll give it a look.
Majid

tpikonen:
The feed above works for me on gPodder master, but only if the youtube-dl extension is off. With the extension on, episode downloads fail. This looks like a bug in the extension.

Do you have the youtube-dl extension enabled, @joesmith222 ?

joesmith222:
> Do you have the youtube-dl extension enabled, @joesmith222 ?

yes

joesmith222:
changed title of thread so if the maker of the youtube-DL extension browses issues, he/she/they may notice it

auouymous:
Any site youtube-dl supports causes the `manage_downloads` setting to fail if `youtube.preferred_fmt_ids` doesn't have a format for it. You can manually download with youtube-dl by right clicking episodes and using the `Download with Youtube-DL` option. Or add `mp3` to `youtube.preferred_fmt_ids`, along with any comma-separated youtube formats, and then set preferred youtube format to custom and re-enable manage_downloads.

tpikonen:
Should we just add a default fallback format 'best' to `add_format()` the youtube-dl extension? It would prevent downloads failing with this site and probably many others as well. Something like this:

```
diff --git a/share/gpodder/extensions/youtube-dl.py b/share/gpodder/extensions/youtube-dl.py
index a5d903c2..7fed3945 100644
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -277,7 +277,7 @@ class gPodderYoutubeDL(download.CustomDownloader):
             logger.debug('no stdout, setting YoutubeDL logger')
             self._ydl_opts['logger'] = logger
 
-    def add_format(self, gpodder_config, opts, fallback=None):
+    def add_format(self, gpodder_config, opts, fallback='best'):
         """ construct youtube-dl -f argument from configured format. """
         # You can set a custom format or custom formats by editing the config for key
[...]

PR #1268, opened 2022-04-10
Add Youtube-DL extension to preferences dialog.

auouymous:
Extensions with a UI for their preferences increase the dialog width but that won't be an issue once #1261 is merged.


tpikonen:
Looks ok, but the strings on checkboxes are a bit long and they don't wrap by default. I'm now checking if there is an easy way to make them wrap, or if the checkboxes should be replaced with something else, like label + switch.

auouymous:
I was hoping to have them as descriptive as possible, and even shortened them because they were too wide for my prefs window. The switch + label would hurt accessibility, unless there is a way to associate them. But it appears you can just do `checkbox = Gtk.Checkbox()` and then `checkbox.add(label)`, where label is a wrappable label. I will update the patch later.

tpikonen:
I have piece of code which sets word-wrapping on for all checkboxes in my soon-ready PR for narrow preferences dialog, so no need to do it manually here.

tpikonen:
Maybe add a default youtube-dl format string here, as discussed in #1263?

The CheckButton line wrapping code is now submitted in #1274.

auouymous:
We can implement the #1263 fix in a separate PR that adds the setting, functionality and UI.

PR #1270, opened 2022-04-10
Turkish translation fix

serkan-maker:


auouymous:
I forgot to mention that `_Close` and `_Delete` are also used along with `Select _none` and `Select _all` on other dialogs. So `_Kapat` would collide with `Seçimleri _kaldır`.

serkan-maker:
now i checked all translation, all letters after (_) sign are unique.

auouymous:
Thanks for the translation.

PR #1272, opened 2022-04-11
Show episodes without downloadable content.

auouymous:
Non-media RSS feeds only have a textual description and can't be subscribed to in gpodder. Some media feeds lack downloadable content and can only be accessed from the website. In both cases, gpodder removes the episodes without any feedback to the user as to why, making it appear to be a problem in gpodder.

This patch adds those episodes but prevents automatic download and displays a proper error when a manual download is attempted. The episodes also get a web browser icon to indicate they lack downloadable content. While not downloadable or transferable to media devices, gpodder can be used to notify the user when new episodes are available and they can be accessed by clicking through to the website.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2805,16 +2805,22 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     # download older episodes first
                     episodes = list(Model.sort_episodes_by_pubdate(episodes))
 
-                if not episodes:
+                # Remove episodes without downloadable content
+                downloadable_episodes = [e for e in episodes if e.url]
+
+                if not downloadable_episodes:
                     # Nothing new here - but inform the user
                     self.pbFeedUpdate.set_fraction(1.0)
-                    self.pbFeedUpdate.set_text(_('No new episodes'))
+                    self.pbFeedUpdate.set_text(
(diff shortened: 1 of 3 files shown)


src/gpodder/gtkui/main.py
```diff
@@ -2805,16 +2805,22 @@ def update_feed_cache_finish_callback():
...
-                if not episodes:
+                # Remove episodes without downloadable content
+                downloadable_episodes = [e for e in episodes if e.url]
+
+                if not downloadable_episodes:
                     # Nothing new here - but inform the user
                     self.pbFeedUpdate.set_fraction(1.0)
-                    self.pbFeedUpdate.set_text(_('No new episodes'))
+                    self.pbFeedUpdate.set_text(
+                        _('No new episodes with downloadable content') if episodes and not downloadable_episodes else _('No new episodes'))
```
tpikonen, line 2815: `and not downloadable_episodes` is redundant here, because of the if-clause above.

tpikonen (changes requested):

tpikonen:
Seems to work fine, and the code is ok, except for the one point in the review.

src/gpodder/gtkui/main.py
auouymous: Oh right, thanks!

elelay:
LGTM, merge at your leisure

PR #1274, opened 2022-04-13
A narrower preferences dialog

tpikonen:
Allows the preferences dialog to be scaled to a width of about 480 px, of which ~180 px is given to the sidebar and the rest goes to the settings pages inside the Stack.

Sets most of the labels to be word-wrapped and splits label+widget combinations to separate lines, or uses FlowBox to do it adaptively. Also adds some vertical space between widgets to make touchscreen use easier.


share/gpodder/ui/gtk/gpodderpreferences.ui
```diff
@@ -866,18 +898,32 @@
                                 <property name="visible">True</property>
                                 <property name="can-focus">True</property>
                                 <property name="receives-default">False</property>
+                                <property name="halign">start</property>
+                                <property name="margin-bottom">8</property>
```
auouymous, line 903: Should this be a top margin? There is a separator below it and the bottom margin creates a visual gap. It seems more logical to have a top margin that moves it away from the combo box above it.

share/gpodder/ui/gtk/gpodderpreferences.ui
```diff
@@ -1015,82 +1015,92 @@
...
-                                    <property name="left-attach">0</property>
-                                    <property name="top-attach">0</property>
-                                  </packing>
                                 </child>
                                 <child>
-                                  <object class="GtkComboBox" id="combobox_preferred_youtube_format">
+                                  <object class="GtkFlowBoxChild">
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
-                                    <property name="hexpand">True</property>
```
auouymous, line 956: On wide dialogs, the combo boxes can be far from the labels, making it harder to match them up. Is it possible to only stretch the combo boxes and not the labels?

The two centered widgets on the updating page have the same problem. Is it possible to have them immediately after the labels and wrap if not enough space?

Can the label+sliders on updating and cleanup pages, and the label+combo boxes on devices page also be made to flow?

auouymous (changes requested):

share/gpodder/ui/gtk/gpodderpreferences.ui
auouymous: It is the last checkbox in the devices playlist section.

auouymous:
Adding the following before the Gtk.Container branch in _wrap_checkbox_labels() will wrap the label inside buttons too.

```python
            elif isinstance(w, Gtk.Button):
                for c in w.get_children():
                    if isinstance(c, Gtk.Label):
                        c.set_line_wrap(True)
                        return
```

The only problem, at least for me, is that the button doesn't resize until hover, causing the wrapped text to partially sit outside of the button. Hiding and showing the button didn't work. But if this could be made to work, we wouldn't need to add the label outside the "Upload local subscriptions" button.

share/gpodder/ui/gtk/gpodderpreferences.ui
```diff
@@ -1037,6 +1051,9 @@
                                 <property name="search-column">1</property>
                                 <signal name="button-release-event" handler="on_treeview_extension_button_released" swapped="no"/>
                                 <signal name="popup-menu" handler="on_treeview_extension_show_context_menu" swapped="no"/>
+                                <child internal-child="selection">
```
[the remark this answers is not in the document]
tpikonen: Glade sets this automatically, since all TreeViews have a TreeSelection object.

share/gpodder/ui/gtk/gpodderpreferences.ui
tpikonen: At least on my system Gtk wraps the CheckButton label so that it expands only downwards (i.e. the label does not move with respect to the checkbox), and the space allocated to the CheckButton does not grow. If there is no free space below, then the label overlaps the separator when wrapped. This is probably a bug in Gtk.

share/gpodder/ui/gtk/gpodderpreferences.ui
tpikonen: I tried various options in the FlowBox and its children, but it looks that extra space is always allocated evenly to both children in a row, so there will be a gap between the label and the combobox.

The combobox and spin button in the Updating page can be put into flowboxes.

The sliders do not work that well with FlowBoxes, the width is again allocated evenly to the label and slider, which results in a gap and wasted space. The 'Update interval' slider could be converted to a ComboBox though.

tpikonen:
> Adding the following before the Gtk.Container branch in _wrap_checkbox_labels() will wrap the label inside buttons too.

I'm not a fan of huge button labels, and would prefer to have the longer explanation in a label and a short action description in a button label as it is now in this PR.

share/gpodder/ui/gtk/gpodderpreferences.ui
auouymous: Changing the auto-download combobox halign from center to fill would stretch it like all other comboboxes.

Can the labels+widgets on devices page be put in flow boxes?

tpikonen: The Devices page label+widgets are now FlowBoxed, and the auto-download combobox set to fill.

auouymous:
The playlist folder on devices causes window to expand for long paths. Everything else looks good.

tpikonen:
Good catch, the label is now ellipsized.

auouymous:
Thanks.

Issue #520, opened 2018-08-13
Strip whitespace around RSS links before adding


Marundu:
I noticed you cannot add podcasts via RSS link if there is a space at the beginning of the link. I tried with two podcasts and an error occurred (see screenshots). Also, gPodder does not pick up the link from the clipboard if it has beginning whitespace. I'm running 64-bit Linux Mint 18. 

![whitespacepodcasterror1](https://user-images.githubusercontent.com/6661880/44024255-7355a7de-9ef5-11e8-982a-ac5c9a68bf35.png)

![whitespacepodcasterror2](https://user-images.githubusercontent.com/6661880/44024256-73b3c508-9ef5-11e8-933e-397e029bd0b7.png)

elelay:
Good point.
That would be a an easy improvement if you're interested.
Everything happens in [addpodcast.py](https://github.com/gpodder/gpodder/blob/master/src/gpodder/gtkui/interface/addpodcast.py).

Marundu:
Seeing this now...
Let me give it a go.

Marundu:
Thanks. Let me look into it. 

How do I get started contributing to the project?

elelay:
1. Fork the project (button on top-right of this page).
    this will create a copy of the code at https://github.com/Marundu/gpodder
2. clone your fork to work on the code on your computer
3. `python3 tools/localdepends.py` to install dependencies
3. `./bin/gpodder -v` to show debug messages
3. make the changes, commit in a branch and push the branch to your fork
4. on your forked repository webpage on github, a "Make pull request" banner will appear for your new branch.
5. submit a pull request. I'll review it and help you get it in shape for incorporating in gPodder
6. Thanks and well done ;-)

PR #1276, opened 2022-04-16
Added logic to trim away leading/trailing whitespaces in user-clipboard auto-inputted urls  

Lexora:
# Added logic to trim away leading/trailing whitespaces in user-clipboard auto-inputted urls  
### Type: feat 

Tested against current gpodder-installed executable for correct/desired 
behaviour. simplified receive_clipboard_text method of class
gPodderAddPodcast in addpodcast.py -- 

Resolves: #520

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -247,6 +247,14 @@ def normalize_feed_url(url):
     """
     if not url or len(url) < 8:
         return None
+    
+    # Removes leading and/or trailing whitespaces - if url contains whitespaces 
+    # in between after str.strip() -> conclude invalid url & return None  
+    url = url.strip()
+    if ' ' in url:
+        return None
+        
+    
 
     # This is a list of prefixes that you can use to minimize the amount of
     # keystrokes that you have to use.
@@ -267,7 +275,7 @@ def normalize_feed_url(url):
 
     # Assume HTTP for URLs without scheme
     if '://' not in url:
-        url = 'http://' + url
+        url = f'http://{url}'  
 
     scheme, netloc, path, query, fragment = urllib.parse.urlsplit(url)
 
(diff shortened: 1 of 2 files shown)


src/gpodder/util.py
```diff
@@ -73,14 +73,16 @@
 try:
     import html5lib
 except ImportError:
-    logger.warn('html5lib not found, falling back to HTMLParser')
+    error_message = "html5lib was not found, fall-back to HTMLParser"
+    logger.warning(f'{error_message}', stack_info=True)
```
tpikonen, line 77: The separate var for `error_message` is not needed, just use a string constant argument in the call and break lines if they are longer than 79 chars.

The minimum Python version for gPodder is 3.5, where f-strings are not supported. Please use other formatting methods.

Also, I'm not sure if `stack_info` is really needed here.

src/gpodder/util.py
```diff
@@ -73,14 +73,16 @@
...
     html5lib = None
 
 if gpodder.ui.win32:
     try:
         import gpodder.utilwin32ctypes as win32file
     except ImportError:
-        logger.warn('Running on Win32 but utilwin32ctypes can\'t be loaded.')
+        error_message = 'Running on Win32: utilwin32ctypes cannot be loaded'
+        logger.warning(f'{error_message}', stack_info=True)
```
tpikonen, line 85: Same here.

src/gpodder/util.py
```diff
@@ -267,7 +275,7 @@ def normalize_feed_url(url):
 
     # Assume HTTP for URLs without scheme
     if '://' not in url:
-        url = 'http://' + url
+        url = f'http://{url}'
```
tpikonen, line 278: Not really needed and uses an f-string.

tpikonen (changes requested):

tpikonen:
Thanks for the patch, see my review for comments.

Lexora:
Thanks for the tips @tpikonen - added commit to update & address requested changes :)

auouymous:
Thank you for the patch.

PR #1280, opened 2022-04-22
Gtk UI: Handle links in HTML shownotes

thp:
Noticed that in HTML shownotes, clicking links (with left mouse button) does nothing. Hook up the handler to open the link in the default web browser. Add a UI confirmation dialog when clicking links in shownotes (both text and HTML) before opening the browser. This can be disabled by setting the advanced config option `ui.gtk.shownotes_confirm_open_browser` to `False`, in which case clicking the link will immediately open the browser.

### DIFF
--- a/src/gpodder/gtkui/shownotes.py
+++ b/src/gpodder/gtkui/shownotes.py
@@ -411,8 +411,9 @@ class gPodderShownotesHTML(gPodderShownotes):
             if req.get_uri() in (self._base_uri, 'about:blank'):
                 decision.use()
             else:
-                logger.debug("refusing to go to %s (base URI=%s)", req.get_uri(), self._base_uri)
+                # Avoid opening the page inside the WebView and open in the browser instead
                 decision.ignore()
+                util.open_website(req.get_uri())
             return False
         else:
             decision.use()


tpikonen:
I'm not sure if the confirmation dialog is needed. The text shownotes have click-to-open already and the behavior should be the same in both widgets.

thp:
Any preferences @elelay @auouymous? I'm happy to remove the confirmation dialog code if it doesn't add anything of value :)

thp:
Gave this another thought and removed the confirmation UI, since as Teemu said, text-only shownotes behave like this.

So the PR is much simpler now. I still kept the `on_link_clicked()` function the base class, as it might make things easier in the future if we want to hook something there (implementation will then be shared by both Text and HTML shownotes), even if right now it's just another layer of indirection.

src/gpodder/gtkui/shownotes.py
```diff
@@ -128,6 +128,9 @@ def hide_pane(self):
         self.pane_is_visible = False
         self.shownotes_pane.hide()
 
+    def on_link_clicked(self, url):
+        util.open_website(url)
+
```
tpikonen, line 133: Is there any value in having this alias to `open_website`? If you remove it, then this PR would be a one-liner.

tpikonen: Ah, you already commented on this. Either way is fine for me...

auouymous: Why is on_link_clicked() stuffed in between pane-related methods?

Regarding the comment outside of this review, couldn't we add the function in the future if something actually needs it? Until then it only makes the code harder to follow. And the title for both text and html shownotes is still using open_website(), so some links are using it but not others.

The confirmation dialog was a good idea for anyone wanting to avoid accidental browser openings, or to see the link they are about to open. It could even be extended to always-open, ask, copy-only. There could be two settings, one for gpodder links and the other for foreign links. The open_website() function could get an optional parameter to pass the foreign link setting from shownotes and channel dialog. If no parameter is given, it would internally use the gpodder link setting. No indirection is required to implement this.

thp: Good point. PR updated, it's now a single-line change :)

auouymous:
LGTM, but I am unable to test it. My only question is why it logs a "refusing to go to ..." message before opening the link?

thp:
> LGTM, but I am unable to test it. My only question is why it logs a "refusing to go to ..." message before opening the link?

Historic reasons. We do `decision.ignore()` so that the web view doesn't navigate to the web page itself. I guess the log message is a bit misleading now, I can remove it.

thp:
Updated with comment.

elelay:
Tested and merged, thanks!

PR #1281, opened 2022-04-22
Gtk UI: Allow streaming any episode using Gtk.AppChooserDialog()

### LINKED ISSUE #228: [Feature Request] Double Click / Enter Expected behaviour
**Current Behaviour**

Double Clicking or pressing Enter on a selected episode opens up the Episode Details segment.

**Expected Behaviour**

Double Clicking or pressing Enter on a selected episode should try and play the downloaded file, otherwise try to stream it.

thp:
Instead of disabling the "Stream" (Play) feature by default until a player is chosen, we can use `Gtk.AppChooserDialog` to show possible applications to open for the mime type of the episode.

Based on a discussion in #228 (this doesn't fix #228, but avoids the "Furthermore, I cannot play a podcast before it has been downloaded." issue).

I haven't checked how this looks on macOS or Windows.

Example for Linux:

Screenshot from 2022-04-22 11-03-55


thp:
Tested this on macOS, and it doesn't show any apps (well, I guess it just finds [Desktop Entry](https://specifications.freedesktop.org/desktop-entry-spec/latest/) files, and macOS doesn't use those):

Screenshot 2022-04-23 at 10 26 47

However, on macOS, the list of apps in the preferences dialog also isn't populated for me, so not sure if maybe that empty dialog is "just fine" for now.

I imagine the situation is similar on Windows, but I haven't done any tests in this direction.

For macOS, a "proper" integration would probably involve [something like this](https://stackoverflow.com/questions/2872994/how-to-launch-choose-application-dialog-on-mac).

auouymous:
We should probably make sure this works well on mac and windows before merging it. Having the stream button appear and then get a blank selection dialog when using it might generate a few bug reports.

Does the selector in preferences show working apps on mac? If so, could this dialog be adapted to use the selector entries for all platforms?

The dialog's `select` button does the same as the `cancel` button for me. Double clicking the app works fine and streams the episode. But the selected app does not get stored in the setting, requiring it to be selected every time.

PR #1284, opened 2022-04-25
Fix #864 add undownload checkbox in delete episode confirm dialog

### LINKED ISSUE #864: Feature Request: add redownload dropdown menu option

elelay:
Closes #864.
Alternative to #868: add a checkbox to the episode delete confirm dialog.

+ more discoverable, more space to explain what undownload does to the user
- users using it routinely will have 2 actions (delete, then check undownload) instead of 1 (undownload)

@auouymous what do you think?

The wording can be improved...

### DIFF
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -115,16 +115,27 @@ class BuilderWidget(GtkBuilderWidget):
         else:
             gpodder.user_extensions.on_notification_show(title, message)
 
-    def show_confirmation(self, message, title=None):
+    def show_confirmation_extended(self, message, title=None, checkbox=None, default_checked=False):
         dlg = Gtk.MessageDialog(self.main_window, Gtk.DialogFlags.MODAL, Gtk.MessageType.QUESTION, Gtk.ButtonsType.YES_NO)
         if title:
             dlg.set_title(str(title))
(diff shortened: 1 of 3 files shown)


auouymous (changes requested):
I like this idea more but it has an issue when undownloading undownloadable episodes. Select a downloaded episode and another that isn't downloaded, then delete them. The checkbox suggests that only one will be marked as new. If multiple downloaded episodes in selection it suggests that possibly all will marked as new. Checking the checkbox and deleting will undownload the downloaded episodes and mark the others as new.

The message should indicate other episodes will not be deleted, only marked as new. Or
it could do something like `_undownload = undownload and episode.can_undownload()` before deleting the file, and use that value to mark as new. The later should indicate that others will be deleted in the checkbox message. The former might be better even though it does nothing more than marking non-downloaded episodes to new, because it does exactly what the message says and the later doesn't list what is deleted and what is marked as new.

elelay:
Thanks for the review: my code was full of mistakes. The issue with marking episodes we don't delete as new is fixed.

auouymous:
LGTM, now to figure out the checkbox message. Does "Mark downloaded episode(s) as new, after deletion, to allow downloading again" make sense?

elelay:
Yes, better wording indeed.

PR #1286, opened 2022-04-28
MPRIS extension fixes

tpikonen:
Two commits:

 1. Some players occasionally send negative 'pos' values to MPRIS. Clamp these values to 0.
 2. The fix for exceptions caused by missing position values in #1265 causes 'pos' to be set to None, which causes further exceptions. If there's no position (or status) property, do nothing. Refactor the code in question slightly.


auouymous:
LGTM, thanks.

PR #1289, opened 2022-05-03
iPod support restored using ctypes and libgpod

thp:
Debian/Ubuntu [haven't built](http://changelogs.ubuntu.com/changelogs/pool/main/libg/libgpod/libgpod_0.8.3-16build1/changelog) the Python bindings for libgpod for a while now (presumably as part of the Python 2 -> Python 3 transition, but I haven't investigated more than this):

```

libgpod (0.8.3-14) unstable; urgency=medium

...

  [ Andrey Rahmatullin ]
  * Drop the Python subpackage.

 -- Andrey Rahmatullin <wrar@debian.org>  Wed, 21 Aug 2019 22:31:15 +0500
 ```

This leaves users without an "easy" way to use iPod sync support, even on Linux.

Apart from that, I noticed that even with `python-gpod` installed, the iPod sync support is probably just broken at the moment.

This patch re-introduces "classic" iPod support using libgpod (which is still available in distros) directly using Python's ctypes FFI module to interact with the C library. It also cleans up some things in `src/gpodder/sync.py` that were unused/dead code.

This actually does more than previous incarnations of iPod support in gPodder ever did:

- "New" status and bookmark position (resume point) on iPod is restored from `current_position` of the episode
- On the next sync, updated bookmark values from the iPod are synchronized to the episode's `current_position`

This means that when listening to an episode on an iPod, and later syncing, the playback position (bookmark) is reflected in gPodder's UI. Deleting episodes on the iPod that have been deleted in gPodder has also been tested and works.

The other great thing about this change is, if we ever get to build `libgpod` and its dependencies for Windows and/or macOS, iPod syncing would work for those devices as well (`libgpod` is basically just writing files on the filesystem for "classic" iPods, nothing magical there).

I tried to be careful not to introduce use-after-free and memory leak bugs, and tried to document the pitfalls I ran into, but there might still be some issues lurking (`valgrind` is mostly happy, though). A "proper" C extension module for Python might be the way to go, but this ctypes-based solution doesn't require development headers or even a compiler (and trying to get the SWIG-based Python bindings re-enabled in Linux distros might be more effort than what it's worth).

### DIFF
--- a/src/gpodder/libgpod_ctypes.py
+++ b/src/gpodder/libgpod_ctypes.py
@@ -1,0 +1,428 @@
+#
+# -*- coding: utf-8 -*-
+#
+# gPodder - A media aggregator and podcast client
+# Copyright (c) 2005-2022 The gPodder Team
+#
+# gPodder is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# gPodder is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+#
+# libgpod_ctypes: Minimalistic ctypes-based bindings for libgpod
+# (Just enough coverage to get podcast syncing working again...)
+# Thomas Perl <m@thp.io>, May 2022
+#
+
+
+import ctypes
+import logging
+import os
+import struct
+
+logger = logging.getLogger(__name__)
+
+
+# libgpod, for iTunesDB access
+libgpod = ctypes.CDLL('libgpod.so.4')
+
+# glib, for g_strdup() and g_free()
+libglib = ctypes.CDLL('libglib-2.0.so.0')
+
+
+# glib/gtypes.h: typedef gint   gboolean;
+gboolean = ctypes.c_int
+
+# glib/gstrfuncs.h: gchar *g_strdup(const gchar *str);
+libglib.g_strdup.argtypes = (ctypes.c_char_p,)
+# Note: This MUST be c_void_p, so that the glib-allocated buffer will
+# be preserved when assigning to track member variables. The reason
+# for this is that Python ctypes tries to be helpful and converts a
+# c_char_p restype to a Python bytes object, which will be different
+# from the memory returned by g_strdup(). For track properties, the
+# values will be free'd indirectly by itdb_free() later.
+libglib.g_strdup.restype = ctypes.c_void_p
+
+# glib/gmem.h: void g_free(gpointer mem);
+libglib.g_free.argtypes = (ctypes.c_void_p,)
+libglib.g_free.restype = None
+
+# XXX: Can we always assume time_t is 64-bit these days?
+time_t = ctypes.c_int64
+
+
+# glib/glist.h: struct _GList
+class GList(ctypes.Structure):
+    ...
+
+
+GList._fields_ = [
+    ('data', ctypes.c_void_p),
+    ('next', ctypes.POINTER(GList)),
+    ('prev', ctypes.POINTER(GList)),
+]
+
+
+# gpod/itdb.h
+class Itdb_iTunesDB(ctypes.Structure):
+    _fields_ = [
+        ('tracks', ctypes.POINTER(GList)),
+        # ...
+    ]
+
+
+# gpod/itdb.h: struct _Itdb_Playlist
+class Itdb_Playlist(ctypes.Structure):
+    _fields_ = [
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by auouymous** [comment]
I'm still reviewing this but you used at least one f-string which isn't supported in 3.5.

File: src/gpodder/libgpod_ctypes.py [hunk 1]
```diff
@@ -0,0 +1,428 @@
...
+# c_char_p restype to a Python bytes object, which will be different
+# from the memory returned by g_strdup(). For track properties, the
+# values will be free'd indirectly by itdb_free() later.
+libglib.g_strdup.restype = ctypes.c_void_p
+
+# glib/gmem.h: void g_free(gpointer mem);
+libglib.g_free.argtypes = (ctypes.c_void_p,)
+libglib.g_free.restype = None
+
+# XXX: Can we always assume time_t is 64-bit these days?
```
auouymous, line 60: You can assume my 32bit system is 64bit, but it would be incorrect.

`platform.architecture()[0]` returns '32bit' or '64bit' and could be used to change the size.

src/gpodder/sync.py
```diff
@@ -307,214 +303,124 @@ def get_free_space(self):
...
-            gpod.itdb_playlist_remove_track(self.podcasts_playlist, track)
-        except:
-            logger.info('Track %s not in playlist', track.title)
-
-        gpod.itdb_track_unlink(track)
-        util.delete_file(filename)
+            self.tracks_list.remove(next((sync_track for sync_track in self.tracks_list
+                                          if sync_track.ipod_track == track), None))
+        except ValueError:
+            ...
```
auouymous, line 366: Is `...` an alias for pass?

auouymous (changes requested):
> Debian/Ubuntu [haven't built](http://changelogs.ubuntu.com/changelogs/pool/main/libg/libgpod/libgpod_0.8.3-16build1/changelog) the Python bindings for libgpod for a while now (presumably as part of the Python 2 -> Python 3 transition, but I haven't investigated more than this):

This comment from the Gentoo ebuild for gpodder suggests the transition is the reason.

```
# As in Fedora: re-enable dev-python/eyeD3 and
# ipod? ( media-libs/libgpod ) once they support python3
```

src/gpodder/libgpod_ctypes.py
thp: Yes and no. See also: https://linux.slashdot.org/story/20/02/15/0247201/linux-is-ready-for-the-end-of-time

I haven't checked with different libc versions what the size of time_t is.

Related glibc documentation: https://www.gnu.org/software/libc/manual/html_node/64_002dbit-time-symbol-handling.html

Just using pointer size for this is probably wrong (but so might be assuming it's 64-bit always, hence the added comment...).

src/gpodder/sync.py
thp: It's the [ellipsis literal](https://docs.python.org/3/library/constants.html#Ellipsis) and can be used instead of `pass` (ellipsis can do a bit more, but here it's used as a `pass` replacement).

src/gpodder/libgpod_ctypes.py
auouymous: Well, time_t **is** 32bit on my 32bit system. I have no plans to use libgpod (no ipod), so we could just let it crash or do bad things for anyone with a 32bit time_t. Maybe at least detect a 32bit platform and log a message that libgpod support might not work if time_t is 32bit.

thp: I'll change it to assume 32-bit time_t on 32-bit platforms, but while looking for a way to maybe "find" the time_t size somewhere in Python, I spotted this in the CPython source code:

```
        /* MS VS2005 changes time_t to a 64-bit type on all platforms */
#       if defined(_MSC_VER) && _MSC_VER >= 1400
#       define SIZEOF_TIME_T 8
#       else
#       define SIZEOF_TIME_T 4
#       endif
```

https://github.com/python/cpython/blob/d284e8b3e351c69406519d5afdcdbed7fbbe0190/PC/pyconfig.h#L301

thp:
This PR in CPython: https://github.com/python/cpython/pull/92870 should take care of exposing `time_t` properly, as it depends on OS, compiler used and other factors whether time_t is 64-bit or 32-bit on 32-bit systems (see https://github.com/python/cpython/issues/92869 for detailed information what I could have gathered -- e.g. OpenBSD and NetBSD use 64-bit time_t also on 32-bit, but FreeBSD on 32-bit i386 uses 32-bit time_t). For now, the current check is probably "good enough" to get it working on 64-bit and 32-bit Linux systems, and if `c_time_t` becomes part of Python's `ctypes`, we can use that eventually.

src/gpodder/libgpod_ctypes.py
auouymous: Looks good, but I still think the 32bit path should log a message to indicate that time_t might be 64bit and might causes issues. Like you mentioned below, there are 32bit platforms that use 64bit time_t, and I read that musl on linux uses it. Being able to ask users if the message exists would be helpful for any issues they open about it.

```python
    logger.info('libgpod may cause issues if time_t is 64-bit on your 32-bit system.')
```

auouymous:
Merged, thanks for the patch.

PR #1293, opened 2022-05-19
Replace context menus with GtkPopovers

tpikonen:
Replaces the context menus in channel, episode and download treeviews ~~(actually just channel for now)~~ with action-based GtkPopover menus. Ported and improved from the adaptive branch. 

I think popovers are a better fit for context menus in gPodder. They look better (IMO) and allow styling with CSS for a larger touch surface, if needed.

~~There is an incomplete and slightly outdated implementation of popover context menus for channel, episode and download treeviews in the adaptive branch. For now, I've ported the channel context menu to the same functionality as the master branch, but if this idea is well received, I will complete the rest of them as well.~~

### DIFF
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -261,9 +261,35 @@
   </menu>
   <menu id="channels-context">
     <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
       <item>
         <attribute name="action">win.updateChannel</attribute>
         <attribute name="label" translatable="yes">Update channel</attribute>
+        <attribute name="verb-icon">view-refresh-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.editChannel</attribute>
+        <attribute name="label" translatable="yes">Channel settings</attribute>
+        <attribute name="verb-icon">preferences-other-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.removeChannel</attribute>
+        <attribute name="label" translatable="yes">Delete channel</attribute>
+        <attribute name="verb-icon">edit-delete-symbolic</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.updateChannel</attribute>
+        <attribute name="label" translatable="yes">Update channel</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.editChannel</attribute>
+        <attribute name="label" translatable="yes">Channel settings</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.removeChannel</attribute>
+        <attribute name="label" translatable="yes">Delete channel</attribute>
       </item>
     </section>
     <section>
@@ -285,19 +311,29 @@
         <attribute name="action">win.refreshImage</attribute>
         <attribute name="label" translatable="yes">Refresh image</attribute>
       </item>
-      <item>
-        <attribute name="action">win.removeChannel</attribute>
-        <attribute name="label" translatable="yes">Delete channel</attribute>
-      </item>
     </section>
-    <section>
-      <item>
-        <attribute name="action">win.editChannel</attribute>
-        <attribute name="label" translatable="yes">Channel settings</attribute>
-      </item>
-    </section>
+    <!-- Extensions -->
   </menu>
   <menu id="episodes-context">
+    <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
+      <item>
+        <attribute name="action">win.play</attribute>
+        <attribute name="label" translatable="yes">Play</attribute>
+        <attribute name="verb-icon">media-playback-start-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Download</attribute>
+        <attribute name="verb-icon">document-save-symbolic</attribute>
+      </item>
+      <!-- Cancel -->
+      <item>
+        <attribute name="action">win.delete</attribute>
+        <attribute name="label" translatable="yes">Delete</attribute>
+        <attribute name="verb-icon">edit-delete-symbolic</attribute>
+      </item>
+    </section>
     <section>
       <item>
         <attribute name="action">win.play</attribute>
@@ -351,6 +387,24 @@
     </item>
   </menu>
   <menu id="downloads-context">
+    <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Start download now</attribute>
+        <attribute name="verb-icon">document-save-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.pause</attribute>
+        <attribute name="label" translatable="yes">Pause</attribute>
+        <attribute name="verb-icon">media-playback-pause-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.cancel</attribute>
+        <attribute name="label" translatable="yes">Cancel</attribute>
+        <attribute name="verb-icon">process-stop-symbolic</attribute>
+      </item>
+    </section>
     <section>
       <item>
         <attribute name="action">win.download</attribute>
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -259,5 +259,128 @@
       </submenu>
     </submenu>
   </menu>
+  <menu id="channels-context">
+    <section>
+      <item>
+        <attribute name="action">win.updateChannel</attribute>
+        <attribute name="label" translatable="yes">Update channel</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.openChannelDownloadFolder</attribute>
+        <attribute name="label" translatable="yes">Open download folder</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.markEpisodesAsOld</attribute>
+        <attribute name="label" translatable="yes">Mark episodes as old</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.channelAutoArchive</attribute>
+        <attribute name="label" translatable="yes">Archive all episodes</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.refreshImage</attribute>
+        <attribute name="label" translatable="yes">Refresh image</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.removeChannel</attribute>
+        <attribute name="label" translatable="yes">Delete channel</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.editChannel</attribute>
+        <attribute name="label" translatable="yes">Channel settings</attribute>
+      </item>
+    </section>
+  </menu>
+  <menu id="episodes-context">
+    <section>
+      <item>
+        <attribute name="action">win.play</attribute>
+        <attribute name="label" translatable="yes">Play</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Download</attribute>
+      </item>
+      <!-- Cancel -->
+      <item>
+        <attribute name="action">win.delete</attribute>
+        <attribute name="label" translatable="yes">Delete</attribute>
+      </item>
+    </section>
+    <!-- Extensions -->
+    <!-- Send To -->
+    <section>
+      <item>
+        <attribute name="action">win.episodeNew</attribute>
+        <attribute name="label" translatable="yes">New</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.episodeLock</attribute>
+        <attribute name="label" translatable="yes">Archive</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.toggleShownotes</attribute>
+        <attribute name="label" translatable="yes">Episode details</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.openEpisodeDownloadFolder</attribute>
+        <attribute name="label" translatable="yes">Open download folder</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.selectChannel</attribute>
+        <attribute name="label" translatable="yes">Select channel</attribute>
+      </item>
+    </section>
+  </menu>
+  <menu id="episodes-context-sendto">
+    <item>
+      <attribute name="action">win.saveEpisodes</attribute>
+      <attribute name="label" translatable="yes">Local folder</attribute>
+    </item>
+    <item>
+      <attribute name="action">win.bluetoothEpisodes</attribute>
+      <attribute name="label" translatable="yes">Bluetooth device</attribute>
+    </item>
+  </menu>
+  <menu id="downloads-context">
+    <section>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Start download now</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.pause</attribute>
+        <attribute name="label" translatable="yes">Pause</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.cancel</attribute>
+        <attribute name="label" translatable="yes">Cancel</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.moveUp</attribute>
+        <attribute name="label" translatable="yes">Move up</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.moveDown</attribute>
+        <attribute name="label" translatable="yes">Move down</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.remove</attribute>
+        <attribute name="label" translatable="yes">Remove from list</attribute>
+      </item>
+    </section>
+  </menu>
 </interface>
 <!-- :noTabs=true:tabSize=2:indentSize=2: -->
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -390,7 +390,7 @@
     <section>
       <attribute name="display-hint">horizontal-buttons</attribute>
       <item>
-        <attribute name="action">win.download</attribute>
+        <attribute name="action">win.forceDownload</attribute>
         <attribute name="label" translatable="yes">Start download now</attribute>
         <attribute name="verb-icon">document-save-symbolic</attribute>
       </item>
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -757,13 +757,13 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         return event.button == 3
 
-    def on_treeview_podcasts_button_released(self, treeview, event):
+    def on_treeview_channels_button_released(self, treeview, event):
         if event.window != treeview.get_bin_window():
             return False
 
         return self.treeview_channels_show_context_menu(event)
 
-    def on_treeview_podcasts_long_press(self, gesture, x, y, treeview):
+    def on_treeview_channels_long_press(self, gesture, x, y, treeview):
         ev = Dummy(x=x, y=y, button=3)
         return self.treeview_channels_show_context_menu(ev)
 
@@ -842,7 +842,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         lp = Gtk.GestureLongPress.new(self.treeChannels)
         lp.set_touch_only(True)
         lp.set_propagation_phase(Gtk.PropagationPhase.CAPTURE)
-        lp.connect("pressed", self.on_treeview_podcasts_long_press, self.treeChannels)
+        lp.connect("pressed", self.on_treeview_channels_long_press, self.treeChannels)
         setattr(self.treeChannels, "long-press-gesture", lp)
 
         # Set up type-ahead find for the podcast list
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -761,19 +761,19 @@ class gPodder(BuilderWidget, dbus.service.Object):
         if event.window != treeview.get_bin_window():
             return False
 
-        return self.treeview_channels_show_context_menu(treeview, event)
+        return self.treeview_channels_show_context_menu(event)
 
     def on_treeview_episodes_button_released(self, treeview, event):
         if event.window != treeview.get_bin_window():
             return False
 
-        return self.treeview_available_show_context_menu(treeview, event)
+        return self.treeview_available_show_context_menu(event)
 
     def on_treeview_downloads_button_released(self, treeview, event):
         if event.window != treeview.get_bin_window():
             return False
 
-        return self.treeview_downloads_show_context_menu(treeview, event)
+        return self.treeview_downloads_show_context_menu(event)
 
     def on_find_podcast_activate(self, *args):
         if self._search_podcasts:
@@ -869,7 +869,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             elif event.keyval == Gdk.KEY_Delete:
                 return False
             elif event.keyval == Gdk.KEY_Menu:
-                self.treeview_channels_show_context_menu(self.treeChannels)
+                self.treeview_channels_show_context_menu()
                 return True
             else:
                 unicode_char_id = Gdk.keyval_to_unicode(event.keyval)
@@ -880,9 +880,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     input_char = chr(unicode_char_id)
                     self._search_podcasts.show_search(input_char)
             return True
-        self.treeChannels.connect('key-press-event', on_key_press)
 
-        self.treeChannels.connect('popup-menu', self.treeview_channels_show_context_menu)
+        self.treeChannels.connect('key-press-event', on_key_press)
+        self.treeChannels.connect('popup-menu',
+            lambda _tv, *args: self.treeview_channels_show_context_menu)
 
         # Enable separators to the podcast list to separate special podcasts
         # from others (this is used for the "all episodes" view)
@@ -1118,7 +1119,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                 else:
                     self.shownotes_object.hide_pane()
             elif event.keyval == Gdk.KEY_Menu:
-                self.treeview_available_show_context_menu(self.treeAvailable)
+                self.treeview_available_show_context_menu()
             elif event.get_state() & Gdk.ModifierType.CONTROL_MASK:
                 # Don't handle type-ahead when control is pressed (so shortcuts
                 # with the Ctrl key still work, e.g. Ctrl+A, ...)
@@ -1132,9 +1133,10 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     input_char = chr(unicode_char_id)
                     self._search_episodes.show_search(input_char)
             return True
-        self.treeAvailable.connect('key-press-event', on_key_press)
 
-        self.treeAvailable.connect('popup-menu', self.treeview_available_show_context_menu)
+        self.treeAvailable.connect('key-press-event', on_key_press)
+        self.treeAvailable.connect('popup-menu',
+            lambda _tv, *args: self.treeview_available_show_context_menu)
 
         self.treeAvailable.enable_model_drag_source(Gdk.ModifierType.BUTTON1_MASK,
                 (('text/uri-list', 0, 0),), Gdk.DragAction.COPY)
@@ -1211,8 +1213,6 @@ class gPodder(BuilderWidget, dbus.service.Object):
         self.treeDownloads.set_model(self.download_status_model)
         TreeViewHelper.set(self.treeDownloads, TreeViewHelper.ROLE_DOWNLOADS)
 
-        self.treeDownloads.connect('popup-menu', self.treeview_downloads_show_context_menu)
-
         # enable multiple selection support
         selection = self.treeDownloads.get_selection()
         selection.set_mode(Gtk.SelectionMode.MULTIPLE)
@@ -1226,11 +1226,13 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         def on_key_press(treeview, event):
             if event.keyval == Gdk.KEY_Menu:
-                self.treeview_downloads_show_context_menu(self.treeDownloads)
+                self.treeview_downloads_show_context_menu()
                 return True
             return False
 
         self.treeDownloads.connect('key-press-event', on_key_press)
+        self.treeDownloads.connect('popup-menu',
+            lambda _tv, *args: self.treeview_downloads_show_context_menu)
 
     def on_treeview_expose_event(self, treeview, ctx):
         model = treeview.get_model()
@@ -1838,7 +1840,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
         # Tell the podcasts tab to update icons for our removed podcasts
         self.update_episode_list_icons(episode_urls)
 
-    def treeview_downloads_show_context_menu(self, treeview, event=None):
+    def treeview_downloads_show_context_menu(self, event=None):
+        treeview = self.treeDownloads
+
         model, paths = self.treeview_handle_context_menu_click(treeview, event)
         if not paths:
             return not treeview.is_rubber_banding_active()
@@ -1894,7 +1898,8 @@ class gPodder(BuilderWidget, dbus.service.Object):
         path = self.podcast_list_model.get_filter_path_from_url(channel.url)
         self.treeChannels.set_cursor(path)
 
-    def treeview_channels_show_context_menu(self, treeview, event=None):
+    def treeview_channels_show_context_menu(self, event=None):
+        treeview = self.treeChannels
         model, paths = self.treeview_handle_context_menu_click(treeview, event)
         if not paths:
             return True
@@ -2033,7 +2038,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         util.run_in_background(lambda: convert_and_send_thread(episodes_to_copy))
 
-    def treeview_available_show_context_menu(self, treeview, event=None):
+    def treeview_available_show_context_menu(self, event=None):
+        treeview = self.treeAvailable
+
         model, paths = self.treeview_handle_context_menu_click(treeview, event)
         if not paths:
             return not treeview.is_rubber_banding_active()
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -50,7 +50,8 @@ from .download import DownloadStatusModel
 from .draw import (cake_size_from_widget, draw_cake_pixbuf,
                    draw_iconcell_scale, draw_text_box_centered)
 from .interface.addpodcast import gPodderAddPodcast
-from .interface.common import BuilderWidget, TreeViewHelper
+from .interface.common import (BuilderWidget, Dummy, ExtensionMenuHelper,
+                               TreeViewHelper)
 from .interface.progress import ProgressIndicator
 from .interface.searchtree import SearchTree
 from .model import EpisodeListModel, PodcastChannelProxy, PodcastListModel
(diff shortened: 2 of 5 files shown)


auouymous:
I am unable to review the patch right now, but it no longer opens with the menu key. I also don't understand the Gtk push to remove icons in menus which aid in quickly finding important items.

tpikonen:
The menu key is now working.

Icons can be specified in the menu model, I put one in for a test, but the icon is not shown, at least in my system.

auouymous:
Do the accelerators work for you? I don't think control keys can be shifted so `^S` either does nothing or does the same as `^s` which syncs. Same with most of the other accelerators.

I like the popover because it doesn't grab the keyboard, allowing focus to change to another window without closing the popover. It also doesn't appear to display a blank menu and require scrolling to see any items like the normal context menus do near the edge of screen. The lack of icons is the biggest issue, and while the popover looks nice, and might fit in with CSD, it is a different menu from those those in the menu bar for non-CSD. However, that might be acceptable due to the other issues they solve.

As for the icons, when using `GTK_DEBUG=interactive` I can set the `visible` and `icon-name` properties of the GtkImage before each GtkLabel and it does display an icon. It looks odd and is hard to read because items without icons don't shift over. https://github.com/GNOME/gtk/blob/main/gtk/gtkpopovermenu.c mentions 'icon' and 'verb-icon' attributes, and a 'display-hint' attribute at top of section will use 'verb-icon' attributes to convert a section into an icon button bar (only the 'horizontal-buttons` value works for me). There might be another section attribute that displays the 'icon' attribute and aligns each label for items without icons.

tpikonen:
> Do the accelerators work for you?

Accelerators set in `menus.ui` do not work, but they can be attached to actions with [gtk_applicaton_set_accels_for_action](https://docs.gtk.org/gtk3/method.Application.set_accels_for_action.html).

The accelerator keys are only shown in GTK4, and there seems to be [little hope of having them in GTK3](https://gitlab.gnome.org/GNOME/gtk/-/issues/1538). But, there are no accelerators shown in current context menus either, so this is not a regression.

I'm not personally missing the icons in context menus, but I suppose the `visible` property of the item icons could be set in code when creating the menu. The alignment could be fixed by setting the `width-request` property to the menu icon width for all images in ModelButtons. This would be a brittle hack though.

auouymous:
> there are no accelerators shown in current context menus either, so this is not a regression.

I know the current context menus lack accelerators, I just meant the patch includes accelerators that don't work, or might collide with existing accelerators if they were made to work (unless control-shift is possible).


> This would be a brittle hack though.

We don't want hacks. Sadly I learned the left margin in the menubar menus is only for radio/check buttons, and adding icons does exactly what the popover does, except it actually works without enabled the visible property. I was really hoping to get a couple icons in the Episodes menu after dynamically generating it. :(

What do you think about adding a horizontal strip of icon buttons for the most common actions at the top of context menus? The named actions could still exist in the menu, they would just have a secondary way to access them via icon at top. This might also be possible in the Episodes menu.

tpikonen:
Added a horizontal button row with common actions and removed the accelerators from the menu definition.

I'm not sure if repeating actions in a menu is good UI design, but at least the common actions are now easy to see.

tpikonen:
Ok, I think this is ready for review now. Rebased to master, now converts the channel, episode and download list context menus to Actions and Popovers.

elelay:
Just tested it: popovers on list right-click on desktop feel strange to me. I understand they are better suited for touch.

tpikonen:
The latest commit removes the transitions which cause at least some of the delay. The looks are a matter of taste of course.

Just by using GMenu models and Actions would get me 90% there in terms of having the menu definitions to be the same in master and adaptive. This is prototyped in #1305.

elelay:
Thanks @tpikonen: now the transition delay is gone. Indeed I'd be happier with #1305. Sorry I missed it 2 months ago.

auouymous:
@elelay Other than looking different, the popovers seem superior. When the window is placed at bottom of screen, and a context menu is opened using menu key for a channel/episode at bottom of window, the menu is completely blank with scroll buttons. Popovers don't have that issue, and they are also part of the window meaning they can be captured in screenshots and don't grab the keyboard when cursor is outside of the window.

src/gpodder/gtkui/main.py
```diff
@@ -3070,6 +3090,16 @@ def on_item_toggle_lock_activate(self, unused, toggle=True, new_value=False):
         self.on_selected_episodes_status_changed()
         self.play_or_download()
 
+    def on_episode_lock_action(self, action, *params):
+        new_state = not action.get_state().get_boolean()
+        for episode in self.get_selected_episodes():
+            episode.mark(is_locked=new_state)
```
auouymous, line 3007: `episode.mark` is flawed, please call on_item_toggle_lock_activate().

src/gpodder/gtkui/main.py
```diff
@@ -3070,6 +3090,16 @@ def on_item_toggle_lock_activate(self, unused, toggle=True, new_value=False):
...
+        self.on_selected_episodes_status_changed()
+        self.play_or_download()
+        action.change_state(GLib.Variant.new_boolean(new_state))
```
auouymous, line 3010: Could the state be set when opening the menu instead of when closing it? Same applies for the "new" toggle below.

src/gpodder/gtkui/main.py
```diff
@@ -3638,6 +3668,16 @@ def get_selected_episodes(self):
     def on_playback_selected_episodes(self, *params):
         self.playback_episodes(self.get_selected_episodes())
 
+    def on_new_action(self, action, *params):
```
auouymous, line 3582: This should probably be `on_episode_new_action`.

src/gpodder/gtkui/main.py
```diff
@@ -1963,125 +1978,57 @@ def treeview_available_show_context_menu(self, treeview, event=None):
...
-            menu.append(item)
-
-            if can_download:
-                item = Gtk.ImageMenuItem(_('Download'))
-                item.set_image(Gtk.Image.new_from_icon_name('document-save', Gtk.IconSize.MENU))
-                item.set_action_name('win.download')
-                menu.append(item)
+                    menu.insert(0, _('Stream'), 'win.play')
+
+            # Download / Pause / Cancel
```
auouymous, line 1961: The Download and Pause items can be shared, but Cancel needs its own item. This is because Download resumes a paused download and Cancel cancels the download if paused or downloading.

src/gpodder/gtkui/main.py
```diff
@@ -1963,125 +1978,57 @@ def treeview_available_show_context_menu(self, treeview, event=None):
...
-                item = Gtk.MenuItem(_('Open download folder'))
-                item.connect('activate', self.on_open_episode_download_folder)
-                menu.append(item)
+            # New and Archive state
+            self.gPodder.lookup_action('episodeNew').change_state(GLib.Variant.new_boolean(any_new))
+            self.gPodder.lookup_action('episodeLock').change_state(GLib.Variant.new_boolean(any_locked))
 
-            menu.attach_to_widget(treeview)
-            menu.show_all()
-            # Disable tooltips while we are showing the menu, so
```
auouymous, line 2079: Could the popover open code disable channel and episode tooltips, and then enable them on close?

src/gpodder/gtkui/main.py
```diff
@@ -1963,125 +1978,57 @@ def treeview_available_show_context_menu(self, treeview, event=None):
...
-            # Ok, this probably makes sense to only display for downloaded files
-            if downloaded:
-                menu.append(Gtk.SeparatorMenuItem())
-                share_menu = self._add_sub_menu(menu, _('Send to'))
-
-                item = Gtk.ImageMenuItem(_('Local folder'))
-                item.set_image(Gtk.Image.new_from_icon_name('folder', Gtk.IconSize.MENU))
-                self._submenu_item_activate_hack(item, self.save_episodes_as_file, episodes)
-                share_menu.append(item)
-                if self.bluetooth_available:
```
auouymous, line 2040: Bluetooth item needs to hide or be disabled if not available.

auouymous (changes requested):

src/gpodder/gtkui/main.py
tpikonen: The state is set (look for 'New and Archive state' in the code) before the popover is shown.

Is there any harm in setting the new state here? The episodes remain selected after the popover closes, so the state is still valid.

auouymous: Ok, I see that state is set when opened, but what is the purpose of setting it immediately before the popover is closed? I've tested and the checkboxes don't update in the split second before it closes.

src/gpodder/gtkui/main.py
```diff
@@ -1963,125 +1978,57 @@ def treeview_available_show_context_menu(self, treeview, event=None):
...
-            item.set_action_name('win.toggleShownotes')
-            menu.append(item)
-
-            if len(self.get_selected_episodes()) == 1:
-                item = Gtk.MenuItem(_('Open download folder'))
-                item.connect('activate', self.on_open_episode_download_folder)
-                menu.append(item)
+            # New and Archive state
+            self.gPodder.lookup_action('episodeNew').change_state(GLib.Variant.new_boolean(any_new))
+            self.gPodder.lookup_action('episodeLock').change_state(GLib.Variant.new_boolean(any_locked))
```
auouymous, line 2001: This should be `can_lock and any_locked` so the item is disabled for deleted episodes.

auouymous (changes requested):

src/gpodder/gtkui/main.py
auouymous: Ignore the any_locked part, it should be disabled if not can_locked. set_episode_actions() handles disabling for actions shared between menus but this uses a different action.

src/gpodder/gtkui/main.py
tpikonen: The episodes remain selected, so the state of the action now represents their current state. There could be some other UI element which shows the state of the action, in a toolbar for example (currently there's not), which would be then updated.

src/gpodder/gtkui/main.py
tpikonen: Should these actions be added to `set_episodes_actions()` then?

src/gpodder/gtkui/main.py
auouymous: I'm thinking that tooltip code was not doing anything and the menus themselves were disabling tooltips. Open a channel or episode context menu (without this patch) and hover over both channel and episode tooltips. So both popovers need to disable tooltips for both treeviews.

src/gpodder/gtkui/main.py
auouymous: Okay.

src/gpodder/gtkui/main.py
auouymous: It only applies to the context menu but the old and new code have no other set_enabled calls, so sure, add it to set_episode_actions() below an `# episode context menu` comment. I've always wanted to merge the two episode menus for consistency, and I think this new code will make that easier. Right after we figure out which menu has the better new/lock behavior. :)

src/gpodder/gtkui/main.py
tpikonen: Ok, I'll disable both.

src/gpodder/gtkui/main.py
```diff
@@ -1963,125 +1978,57 @@ def treeview_available_show_context_menu(self, treeview, event=None):
...
-            self.treeview_allow_tooltips(self.treeAvailable, False)
-            menu.connect('deactivate', lambda menushell: self.treeview_allow_tooltips(self.treeAvailable, True))
             if event is None:
-                func = TreeViewHelper.make_popup_position_func(treeview)
-                menu.popup(None, None, func, None, 3, Gtk.get_current_event_time())
+                area = TreeViewHelper.get_popup_rectangle(treeview)
             else:
-                menu.popup(None, None, None, None, event.button, event.time)
+                area = Gdk.Rectangle()
+                area.x, area.y = event.x, (event.y + 25)  # adjust for treeView header
```
auouymous, line 2009: Is 25 the height of header on your system? Fixed values like that fail when users have different font sizes or themes.

auouymous (changes requested):

src/gpodder/gtkui/main.py
```diff
@@ -1028,6 +1041,8 @@ def on_key_press(treeview, event):
                     self._search_episodes.hide_search()
                 else:
                     self.shownotes_object.hide_pane()
+            elif event.keyval == Gdk.KEY_Menu:
+                self.treeview_available_show_context_menu(self.treeChannels)
```
auouymous, line 1081: Wrong treeview, should be treeAvailable.

auouymous (changes requested):

src/gpodder/gtkui/main.py
auouymous: Also, when opened from menu key, the header height isn't accounted for, causing the menu to be higher up than it should.

src/gpodder/gtkui/main.py
tpikonen: There was another mistake if this type there too. I refactored the code to make them impossible.

src/gpodder/gtkui/main.py
tpikonen: Thanks for catching this, I forgot to fix this hack. Now it should be correct.

tpikonen:
The commit stack needs serious refactoring, so setting to draft again.

share/gpodder/ui/gtk/menus.ui
```diff
@@ -267,5 +267,51 @@
       </item>
     </section>
   </menu>
+  <menu id="episodes-context">
+      <item>
```
auouymous, line 271: Should these items be wrapped in a `<section>`?

auouymous (changes requested):

share/gpodder/ui/gtk/menus.ui
tpikonen: Yes, it does not seem to affect how the menu looks like though.

auouymous: I forgot to say the download menu was also missing a `<section>`. They don't change the look, but they are probably needed to add buttons above the items.

tpikonen: Added, thanks.

tpikonen:
Fixes from review are now mostly included in the patches. Ready for another round.

auouymous:
Is it not possible to create the verb-icon menu items from code?

tpikonen:
> Is it not possible to create the verb-icon menu items from code?

If you're after icons in front of the menu labels, I think the only way is to set the icon visibility of the ModelButtons to true in code. The section `display-hint` property values all produce a horizontal row of icon buttons.

Of course the content of a Popover can also be constructed by hand just as any other widget tree.

tpikonen:
Since there's now a consensus on using the popovers for context menus, let's try to finish this. Rebased to master and added the recently added 'Select channel' item to the episodes context menu.

Seemingly works, but needs testing.

auouymous:
> > Is it not possible to create the verb-icon menu items from code?
> 
> Of course the content of a Popover can also be constructed by hand just as any other widget tree.

Does this construct the episode context menu by hand so verb-icons can be used? I'm assuming it does not seeing as how the episodes menu is graying out some items that don't show in the old menu.

src/gpodder/gtkui/main.py
```diff
@@ -351,11 +357,16 @@ def create_actions(self):
...
+            ('openChannelDownloadFolder', self.on_open_download_folder),
             ('selectChannel', self.on_select_channel_of_episode),
             ('findEpisode', self.on_find_episode_activate),
             ('toggleShownotes', self.on_shownotes_selected_episodes),
             # Extras
             ('sync', self.on_sync_to_device_activate),
+            ('findPodcast', self.on_find_podcast_activate),
+            ('findEpisode', self.on_find_episode_activate),
+            ('markEpisodesAsOld', self.on_mark_episodes_as_old),
+            ('refreshImage', self.on_itemRefreshCover_activate),
```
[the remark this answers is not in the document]
tpikonen: Thanks, there were some duplicates there as well. I removed them and reordered the remaining actions.

tpikonen:
> Does this construct the episode context menu by hand so verb-icons can be used? I'm assuming it does not seeing as how the episodes menu is graying out some items that don't show in the old menu.

No, the context menu popover widget is generated from the menu model. The menu model itself is partially dynamically generated to make the extension submenus, Play/Download/Stream etc. work.

ernestask:
Is there any help needed towards this getting merged? I’ve just started using the app and I’m already seeing nits to pick with the UI, so I might as well use the motivation to review things and whatnot.

tpikonen:
It would be nice to get this merged eventually. Is there anything to fix still?

auouymous:
The progress menu has an issue where "start download now" quickly grays out after opening the menu on a queued episode. I assume it is caused by `update_downloads_list` calling `play_or_download` every 1.5 seconds, which uses can_queue to set the download status, and since the episode is already queued...

The following patch not only fixes the problem but also allows the download button in toolbar to force start it. However, neither one force starts the download.

```patch
@@ -2360,6 +2378,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     can_pause, can_cancel, can_delete, can_lock)
         else:
             (can_queue, can_pause, can_cancel, can_remove) = (False,) * 4
+            can_force = True
 
             selection = self.treeDownloads.get_selection()
             if selection.count_selected_rows() > 0:
@@ -2375,6 +2394,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
                         logger.error('Invalid task at path %s', str(path))
                         continue
 
+                    if task.status != download.DownloadTask.QUEUED:
+                        can_force = False
+
                     # These values should only ever be set, never unset them once set.
                     # Actions filter tasks using these methods.
                     can_queue = can_queue or task.can_queue()
@@ -2382,7 +2404,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     can_cancel = can_cancel or task.can_cancel()
                     can_remove = can_remove or task.can_remove()
 
-            self.set_episode_actions(False, False, can_queue, can_pause, can_cancel, can_remove, False, False)
+            self.set_episode_actions(False, False, can_queue or can_force, can_pause, can_cancel, can_remove, False, False)
 
             return (False, False, False, can_queue, can_pause, can_cancel,
                     can_remove, False)
```


I'm testing it with the following patch to add verb icons. Do you see any issues with this?

```patch
diff --git a/share/gpodder/ui/gtk/menus.ui b/share/gpodder/ui/gtk/menus.ui
index f8c6df94..2e725ed9 100644
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -261,9 +261,35 @@
   </menu>
   <menu id="channels-context">
     <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
       <item>
         <attribute name="action">win.updateChannel</attribute>
         <attribute name="label" translatable="yes">Update channel</attribute>
+        <attribute name="verb-icon">view-refresh-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.editChannel</attribute>
+        <attribute name="label" translatable="yes">Channel settings</attribute>
+        <attribute name="verb-icon">preferences-other-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.removeChannel</attribute>
+        <attribute name="label" translatable="yes">Delete channel</attribute>
+        <attribute name="verb-icon">edit-delete-symbolic</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.updateChannel</attribute>
+        <attribute name="label" translatable="yes">Update channel</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.editChannel</attribute>
+        <attribute name="label" translatable="yes">Channel settings</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.removeChannel</attribute>
+        <attribute name="label" translatable="yes">Delete channel</attribute>
       </item>
     </section>
     <section>
@@ -285,19 +311,29 @@
         <attribute name="action">win.refreshImage</attribute>
         <attribute name="label" translatable="yes">Refresh image</attribute>
       </item>
-      <item>
-        <attribute name="action">win.removeChannel</attribute>
-        <attribute name="label" translatable="yes">Delete channel</attribute>
-      </item>
-    </section>
-    <section>
-      <item>
-        <attribute name="action">win.editChannel</attribute>
-        <attribute name="label" translatable="yes">Channel settings</attribute>
-      </item>
     </section>
+    <!-- Extensions -->
   </menu>
   <menu id="episodes-context">
+    <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
+      <item>
+        <attribute name="action">win.play</attribute>
+        <attribute name="label" translatable="yes">Play</attribute>
+        <attribute name="verb-icon">media-playback-start-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Download</attribute>
+        <attribute name="verb-icon">document-save-symbolic</attribute>
+      </item>
+      <!-- Cancel -->
+      <item>
+        <attribute name="action">win.delete</attribute>
+        <attribute name="label" translatable="yes">Delete</attribute>
+        <attribute name="verb-icon">edit-delete-symbolic</attribute>
+      </item>
+    </section>
     <section>
       <item>
         <attribute name="action">win.play</attribute>
@@ -351,6 +387,24 @@
     </item>
   </menu>
   <menu id="downloads-context">
+    <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
+      <item>
+        <attribute name="action">win.download</attribute>
+        <attribute name="label" translatable="yes">Start download now</attribute>
+        <attribute name="verb-icon">document-save-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.pause</attribute>
+        <attribute name="label" translatable="yes">Pause</attribute>
+        <attribute name="verb-icon">media-playback-pause-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.cancel</attribute>
+        <attribute name="label" translatable="yes">Cancel</attribute>
+        <attribute name="verb-icon">process-stop-symbolic</attribute>
+      </item>
+    </section>
     <section>
       <item>
         <attribute name="action">win.download</attribute>
@@ -365,6 +419,19 @@
         <attribute name="label" translatable="yes">Cancel</attribute>
       </item>
     </section>
+    <section>
+      <attribute name="display-hint">horizontal-buttons</attribute>
+      <item>
+        <attribute name="action">win.moveUp</attribute>
+        <attribute name="label" translatable="yes">Move up</attribute>
+        <attribute name="verb-icon">go-up-symbolic</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.moveDown</attribute>
+        <attribute name="label" translatable="yes">Move down</attribute>
+        <attribute name="verb-icon">go-down-symbolic</attribute>
+      </item>
+    </section>
     <section>
       <item>
         <attribute name="action">win.moveUp</attribute>
diff --git a/src/gpodder/gtkui/main.py b/src/gpodder/gtkui/main.py
index 889d8e75..14bcbf97 100644
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -830,7 +830,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         menu = self.application.builder.get_object('channels-context')
         # Extensions section, updated in signal handler
         extmenu = Gio.Menu()
-        menu.insert_section(3, _('Extensions'), extmenu)
+        menu.insert_section(4, _('Extensions'), extmenu)
         self.channel_context_menu_helper = ExtensionMenuHelper(
             self.gPodder, extmenu, 'channel_context_action_')
         self.channels_popover = Gtk.Popover.new_from_model(self.treeChannels, menu)
@@ -972,7 +972,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         menu = self.application.builder.get_object('episodes-context')
         # Extensions section, updated dynamically
         extmenu = Gio.Menu()
-        menu.insert_section(1, _('Extensions'), extmenu)
+        menu.insert_section(2, _('Extensions'), extmenu)
         self.episode_context_menu_helper = ExtensionMenuHelper(
             self.gPodder, extmenu, 'episode_context_action_')
         # Send To submenu section, shown only for downloaded episodes
@@ -1885,14 +1885,22 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     self.downloads_list_get_selection(model, paths)
 
             menu = self.application.builder.get_object('downloads-context')
-            dsec = menu.get_item_link(0, Gio.MENU_LINK_SECTION)
+            vsec = menu.get_item_link(0, Gio.MENU_LINK_SECTION)
+            dsec = menu.get_item_link(1, Gio.MENU_LINK_SECTION)
 
+            def insert_menuitem(position, label, action, icon):
+                dsec.insert(position, label, action)
+                menuitem = Gio.MenuItem.new(label, action)
+                menuitem.set_attribute_value('verb-icon', GLib.Variant.new_string(icon))
+                vsec.insert_item(position, menuitem)
+
+            vsec.remove(0)
             dsec.remove(0)
             if can_force:
-                dsec.insert(0, _('Start download now'), 'win.download')
+                insert_menuitem(0, _('Start download now'), 'win.download', 'document-save-symbolic')
                 self.download_action.set_enabled(True)
             else:
-                dsec.insert(0, _('Download'), 'win.download')
+                insert_menuitem(0, _('Download'), 'win.download', 'document-save-symbolic')
                 self.download_action.set_enabled(can_queue)
 
             self.gPodder.lookup_action('remove').set_enabled(can_remove)
@@ -2086,34 +2094,44 @@ class gPodder(BuilderWidget, dbus.service.Object):
              can_cancel, can_delete, can_lock) = self.play_or_download()
 
             menu = self.application.builder.get_object('episodes-context')
-            psec = menu.get_item_link(0, Gio.MENU_LINK_SECTION)
+            vsec = menu.get_item_link(0, Gio.MENU_LINK_SECTION)
+            psec = menu.get_item_link(1, Gio.MENU_LINK_SECTION)
+
+            def insert_menuitem(position, label, action, icon):
+                psec.insert(position, label, action)
+                menuitem = Gio.MenuItem.new(label, action)
+                menuitem.set_attribute_value('verb-icon', GLib.Variant.new_string(icon))
+                vsec.insert_item(position, menuitem)
 
             # Play / Stream / Preview / Open
+            vsec.remove(0)
             psec.remove(0)
             if open_instead_of_play:
-                psec.insert(0, _('Open'), 'win.play')
+                insert_menuitem(0, _('Open'), 'win.play', 'document-open-symbolic')
             else:
                 if downloaded:
-                    psec.insert(0, _('Play'), 'win.play')
+                    insert_menuitem(0, _('Play'), 'win.play', 'media-playback-start-symbolic')
                 elif can_preview:
-                    psec.insert(0, _('Preview'), 'win.play')
+                    insert_menuitem(0, _('Preview'), 'win.play', 'media-playback-start-symbolic')
                 else:
-                    psec.insert(0, _('Stream'), 'win.play')
+                    insert_menuitem(0, _('Stream'), 'win.play', 'media-playback-start-symbolic')
 
             # Download / Pause
+            vsec.remove(1)
             psec.remove(1)
             if can_pause:
-                psec.insert(1, _('Pause'), 'win.pause')
+                insert_menuitem(1, _('Pause'), 'win.pause', 'media-playback-pause-symbolic')
             else:
-                psec.insert(1, _('Download'), 'win.download')
+                insert_menuitem(1, _('Download'), 'win.download', 'document-save-symbolic')
 
             # Cancel
             have_cancel = (psec.get_item_attribute_value(
                 2, "action", GLib.VariantType("s")).get_string() == 'win.cancel')
             if not can_cancel and have_cancel:
+                vsec.remove(2)
                 psec.remove(2)
             elif can_cancel and not have_cancel:
-                psec.insert(2, _('_Cancel'), 'win.cancel')
+                insert_menuitem(2, _('Cancel'), 'win.cancel', 'process-stop-symbolic')
 
             # Extensions section
             entries = [(label, None if func is None else lambda a, b: func(episodes))
```

auouymous:
The `_('Open'), 'win.play'` in our patches must be changed to `_('Open'), 'win.open'` so it enables the menuitem.

auouymous:
And the two `self.download_action.set_enabled` calls inside `treeview_downloads_show_context_menu` aren't needed because play_or_download sets the actions and toolbar buttons when a selection happens, before the menu is opened.

auouymous:
The following patch mostly fixes the "start download now" issue. The toolbar button either needs to be grayed out or preferably toggled to send `win.forceDownload`, if possible. I'm going to try to get a release out now so it might be a while until I can look at that.

```patch
diff --git a/share/gpodder/ui/gtk/menus.ui b/share/gpodder/ui/gtk/menus.ui
index 2e725ed9..f55c6330 100644
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -390,7 +390,7 @@
     <section>
       <attribute name="display-hint">horizontal-buttons</attribute>
       <item>
-        <attribute name="action">win.download</attribute>
+        <attribute name="action">win.forceDownload</attribute>
         <attribute name="label" translatable="yes">Start download now</attribute>
         <attribute name="verb-icon">document-save-symbolic</attribute>
       </item>
diff --git a/src/gpodder/gtkui/main.py b/src/gpodder/gtkui/main.py
index 4505c90b..02f57fc3 100644
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -376,6 +376,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             # Episodes
             ('play', self.on_playback_selected_episodes),
             ('open', self.on_playback_selected_episodes),
+            ('forceDownload', self.on_force_download_selected_episodes),
             ('download', self.on_download_selected_episodes),
             ('pause', self.on_pause_selected_episodes),
             ('cancel', self.on_item_cancel_download_activate),
@@ -410,6 +411,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         # Episodes
         self.play_action = g.lookup_action('play')
         self.open_action = g.lookup_action('open')
+        self.force_download_action = g.lookup_action('forceDownload')
         self.download_action = g.lookup_action('download')
         self.pause_action = g.lookup_action('pause')
         self.cancel_action = g.lookup_action('cancel')
@@ -1897,7 +1899,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             vsec.remove(0)
             dsec.remove(0)
             if can_force:
-                insert_menuitem(0, _('Start download now'), 'win.download', 'document-save-symbolic')
+                insert_menuitem(0, _('Start download now'), 'win.forceDownload', 'document-save-symbolic')
                 self.download_action.set_enabled(True)
             else:
                 insert_menuitem(0, _('Download'), 'win.download', 'document-save-symbolic')
@@ -3865,6 +3867,15 @@ class gPodder(BuilderWidget, dbus.service.Object):
                                DownloadStatusModel.C_TASK)) for path in paths]
             self._for_each_task_set_status(selected_tasks, download.DownloadTask.QUEUED)
 
+    def on_force_download_selected_episodes(self, action_or_widget, param=None):
+        if self.wNotebook.get_current_page() == 1:
+            selection = self.treeDownloads.get_selection()
+            (model, paths) = selection.get_selected_rows()
+            selected_tasks = [(Gtk.TreeRowReference.new(model, path),
+                               model.get_value(model.get_iter(path),
+                               DownloadStatusModel.C_TASK)) for path in paths]
+            self._for_each_task_set_status(selected_tasks, download.DownloadTask.QUEUED, True)
+
     def on_pause_selected_episodes(self, action_or_widget, param=None):
         if self.wNotebook.get_current_page() == 0:
             selection = self.get_selected_episodes()
```

tpikonen:
Ok, I'll see if I can make the forced downloads work. Your patches are fine, except the up and down arrows in download context menu look really out of place.

auouymous:
The extensions in channel and episode context menus don't work. It seems like no matter which extension menuitem is clicked, it always activates the last menuitem.


> except the up and down arrows in download context menu look really out of place.

I agree.

auouymous:
> The extensions in channel and episode context menus don't work. It seems like no matter which extension menuitem is clicked, it always activates the last menuitem.

@tpikonen I have looked over the code and each extension entry gets a unique action_id and the correct callback is connected to 'activate', but they all trigger the last callback. Can you see the problem?

tpikonen:
The bug in extension callbacks was a pretty subtle one, but it should be fixed now. I also integrated your patches and rebased to master, so it should be ready for testing (which I did not do much, except for the extension callbacks).

The patch stack is a mess and needs to be squashed before merging.

auouymous:
This patch disables the toolbar download button when the progress list is empty.

```patch
diff --git a/src/gpodder/gtkui/main.py b/src/gpodder/gtkui/main.py
index 8aa13f4e..76582aef 100644
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2406,6 +2406,8 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     can_pause = can_pause or task.can_pause()
                     can_cancel = can_cancel or task.can_cancel()
                     can_remove = can_remove or task.can_remove()
+            else:
+                can_force = False
 
             self.set_episode_actions(False, False, can_queue or can_force, can_pause, can_cancel, can_remove, False, False)
 
```

The toolbar download button is not disabled for queued downloads, but does nothing when clicked. This patch adds a toolbar force download button.

```patch
diff --git a/share/gpodder/ui/gtk/gpodder.ui b/share/gpodder/ui/gtk/gpodder.ui
index 9977f655..50f65217 100644
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -44,6 +44,21 @@
                 <property name="homogeneous">True</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkToolButton" id="toolForceDownload">
+                <property name="visible">False</property>
+                <property name="sensitive">False</property>
+                <property name="can-focus">False</property>
+                <property name="is-important">True</property>
+                <property name="label" translatable="yes">Start download now</property>
+                <property name="icon-name">document-save-symbolic</property>
+                <signal name="clicked" handler="on_force_download_selected_episodes" swapped="no"/>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="homogeneous">True</property>
+              </packing>
+            </child>
             <child>
               <object class="GtkToolButton" id="toolDownload">
diff --git a/src/gpodder/gtkui/main.py b/src/gpodder/gtkui/main.py
index 76582aef..4fdcd57e 100644
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -2161,8 +2161,8 @@ class gPodder(BuilderWidget, dbus.service.Object):
             self.episodes_popover.show()
             return True
 
-    def set_episode_actions(self, open_instead_of_play=False, can_play=False, can_download=False, can_pause=False, can_cancel=False,
-                            can_delete=False, can_lock=False, is_episode_selected=False):
+    def set_episode_actions(self, open_instead_of_play=False, can_play=False, can_force=False, can_download=False,
+                            can_pause=False, can_cancel=False, can_delete=False, can_lock=False, is_episode_selected=False):
         episodes = self.get_selected_episodes() if is_episode_selected else []
 
         # play icon and label
@@ -2184,6 +2184,9 @@ class gPodder(BuilderWidget, dbus.service.Object):
 
         # toolbar
         self.toolPlay.set_sensitive(can_play)
+        self.toolForceDownload.set_visible(can_force)
+        self.toolForceDownload.set_sensitive(can_force)
+        self.toolDownload.set_visible(not can_force)
         self.toolDownload.set_sensitive(can_download)
         self.toolPause.set_sensitive(can_pause)
         self.toolCancel.set_sensitive(can_cancel)
@@ -2191,7 +2194,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
         # Episodes menu
         self.play_action.set_enabled(can_play and not open_instead_of_play)
         self.open_action.set_enabled(can_play and open_instead_of_play)
-        self.download_action.set_enabled(can_download)
+        self.download_action.set_enabled(can_force or can_download)
         self.pause_action.set_enabled(can_pause)
         self.cancel_action.set_enabled(can_cancel)
         self.delete_action.set_enabled(can_delete)
@@ -2374,7 +2377,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
                     can_delete = can_delete or episode.can_delete()
                     can_lock = can_lock or episode.can_lock()
 
-            self.set_episode_actions(open_instead_of_play, can_play, can_download, can_pause, can_cancel, can_delete, can_lock,
+            self.set_episode_actions(open_instead_of_play, can_play, False, can_download, can_pause, can_cancel, can_delete, can_lock,
                                     selection.count_selected_rows() > 0)
 
             return (open_instead_of_play, can_play, can_preview, can_download,
@@ -2409,7 +2412,7 @@ class gPodder(BuilderWidget, dbus.service.Object):
             else:
                 can_force = False
 
-            self.set_episode_actions(False, False, can_queue or can_force, can_pause, can_cancel, can_remove, False, False)
+            self.set_episode_actions(False, False, can_force, can_queue, can_pause, can_cancel, can_remove, False, False)
 
             return (False, False, False, can_queue, can_pause, can_cancel,
                     can_remove, False)
```

Other than that, the PR is ready to merge. I would like to fix #1538 and get out another release first, and then this could be merged.

tpikonen:
I added the patches by @auouymous. Will refactor the patch stack soonish.

tpikonen:
The commit stack is now cleaned up, so as far as I'm concerned, this is ready.

auouymous:
Thanks for all the work you did on this.

auouymous:
Gtk removed icons for menu items, the buttons provide that icon and the textual menu items describe what each icon does. We couldn't find a better way to have both.

PR #1297, opened 2022-06-06
Treat only really new episodes as new after an update

tpikonen:
Currently, when a channel or all channels are updated, all episodes marked as new are treated as being new, which usually means being shown in the episode selector. If one treats the 'new' marking as a kind of a to-do list for interesting episodes, sorting through the old, but marked as new episodes after each update gets tiring. This PR makes gPodder treat as new only episodes which are added in an update.

This change could be added behind a new configuration variable, but I decided not to, because episodes marked as new are easily accessible either from View->Unplayed episodes, or from Podcasts->Download new episodes.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -522,7 +522,7 @@
                                 <child>
                                   <object class="GtkFlowBoxChild">
                                     <property name="visible">True</property>
-                                    <property name="can-focus">True</property>
+                                    <property name="can-focus">False</property>
                                     <child>
                                       <object class="GtkLabel" id="label_episode_limit">
                                         <property name="visible">True</property>
@@ -537,7 +537,7 @@
                                 <child>
                                   <object class="GtkFlowBoxChild">
                                     <property name="visible">True</property>
-                                    <property name="can-focus">True</property>
+                                    <property name="can-focus">False</property>
                                     <child>
                                       <object class="GtkSpinButton" id="spinbutton_episode_limit">
                                         <property name="visible">True</property>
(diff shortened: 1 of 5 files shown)


auouymous:
Would this mean youtube livestreams and premieres require manual downloading when they finally become available? I mark those as new to force a download on the next update, or old if I want to see the duration before downloading.

Marking as old could be used for todo episodes, and deleting when not interested. But I can see how that doesn't work for anyone using the unplayed view mode. Perhaps this could detect the unplayed view setting, or add a custom setting because it might be useful for someone using the other view modes.

tpikonen:
> Would this mean youtube livestreams and premieres require manual downloading when they finally become available?

Probably. Admittedly this is a use case I did not think about, since I never have auto-download or auto-queue on.

> Marking as old could be used for todo episodes, and deleting when not interested.

Well, personally I have (too) many feeds and end up watching or listening only a small fraction of even potentially interesting episodes. Downloading and then deleting would be a waste of bandwidth and storage, and I mostly stream my channels anyway.

So, since there are many ways of using gPodder, should this behavior go behind a config option, with the default being off for backwards compatibility?

auouymous:
Have you tested this with the filter extension? For http://borg.qzx.com/test-rss.php, I blocked "aaa" but updating the channel wants to download the newly seen "aaa" episodes.


> Downloading and then deleting would be a waste of bandwidth and storage

I use the filter extension to unmark (prevent auto-download) of most channels. By 'delete', I meant deleting the undownloaded episode to get rid of it (using the hide deleted episodes view mode), and either manually downloading, marking new to auto-download on next update, or leaving unmarked if unsure or not ready to download.


> should this behavior go behind a config option, with the default being off for backwards compatibility?

That would make it available for all users, no matter what their view mode is. Or it could be a ternary option with off, only if unplayed view, always on. That would allow one who doesn't want it always on to change view mode while updating. Your choice though.

tpikonen:
Rebased to master, added a config variable for this feature and a checkbutton for setting it in the preferences. The filter extension now works as it should.

Also fix the tabbing behavior in preferences by setting 'can-focus' to False in GtkFlowBoxChild containers.

tpikonen:
Rebased to master. I've been using this for a while now without problems. Could you review this again @auouymous ?

auouymous:
Merged, thanks for the enhancement, and thanks for fixing the tab issue.

PR #1298, opened 2022-06-07
Fix undownload dialog layout

### LINKED ISSUE #1292: Need suggestions to fix layout of the new delete dialog
The delete dialog now has an undownload checkbox added by #1284 (when deleting downloaded episodes), and I've discovered a few issues after merging it, oops.

The deleted episode titles are clipped to 60 characters and the "deleting removes files" message is shorter than that, at least in english. The problem is that the checkbox message increases the dialog width, and may or may not be an issue on mobile (@tpikonen). I could insert a newline into the message but translations might not position it properly. Wrapping the checkbox label with the following patch does nothing.

```patch
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -124,6 +124,7 @@ class BuilderWidget(GtkBuilderWidget):
             dlg.set_markup('<span weight="bold" size="larger">%s</span>' % (message))
         if checkbox:
             cb = Gtk.CheckButton.new_with_label(checkbox)
+            cb.get_child().set_line_wrap(True)
             cb.set_active(default_checked)
             dlg.get_message_area().pack_end(cb, False, False, 0)
             dlg.get_widget_for_response(Gtk.ResponseType.NO).grab_focus()
```

Should we use a newline, force dialog to an unresizeable width or does someone have a better solution?

The checkbox is also too close to the message, making it too busy and harder to read. Should we add a separator, somehow move the checkbox out of the message area so it is closer to the buttons with padding above, or does anyone have a better idea?

```patch
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -126,6 +126,7 @@ class BuilderWidget(GtkBuilderWidget):
             cb = Gtk.CheckButton.new_with_label(checkbox)
             cb.set_active(default_checked)
             dlg.get_message_area().pack_end(cb, False, False, 0)
+            dlg.get_message_area().pack_end(Gtk.Separator(orientation=Gtk.Orientation.HORIZONTAL), False, False, 0
             dlg.get_widget_for_response(Gtk.ResponseType.NO).grab_focus()
         dlg.show_all()
         response = dlg.run()
```

auouymous:
This forces the new undownload checkbox to wrap, preventing wide delete dialogs. It also wraps other confirmation dialogs, like the wide "uploading list to gpodder.net" dialog. It would be better if the message textview could be resized to fit its content, but that doesn't appear to be possible. The chosen height should display all dialogs, including translations, without scrolling the message. It should only scroll when deleting multiple episodes with long titles.

The confirmation dialog is used for:
* deleting episodes (with or without undownload checkbox)
* deleting channels
* upload button on gpodder.net preferences
* new version available, download latest
* overwrite saved episode when sending to local folder
* error saving episode when sending to local folder
* URL redirection warning when adding channels
* sync to device when no device is configured
* not enough space when syncing

Fixes #1292.

### DIFF
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -119,14 +119,62 @@ class BuilderWidget(GtkBuilderWidget):
         dlg = Gtk.MessageDialog(self.main_window, Gtk.DialogFlags.MODAL, Gtk.MessageType.QUESTION, Gtk.ButtonsType.YES_NO)
         if title:
             dlg.set_title(str(title))
-            dlg.set_markup('<span weight="bold" size="larger">%s</span>\n\n%s' % (title, message))
+
+            # Long titles will cause an excessively wide dialog.
+            dlg.set_markup('<span weight="bold" size="larger">%s</span>' % title)
+
+            tv = Gtk.TextView()
+            tv.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
+            tv.set_border_width(10)
+            tv.set_editable(False)
+            tb = Gtk.TextBuffer()
+            tb.insert_markup(tb.get_start_iter(), message, -1)
+            tv.set_buffer(tb)
+            tv.set_property('expand', True)
+
+            sw = Gtk.ScrolledWindow()
+            sw.set_size_request(400, 150)
+            sw.set_property('shadow-type', Gtk.ShadowType.IN)
+            sw.add(tv)
+            sw.show_all()
+            dlg.get_message_area().pack_start(sw, True, True, 0)
+
+            # TextView height is not allocated until after it is realized,
+            # preventing its ScrolledWindow from being resized to fit message.
         else:
-            dlg.set_markup('<span weight="bold" size="larger">%s</span>' % (message))
+            # Long messages will cause an excessively wide dialog.
+            # Add a short title to move message to a TextView.
+            dlg.set_markup('<span weight="bold" size="larger">%s</span>' % message)
+
+        # make title or message copy/pastable
+        for lbl in dlg.get_message_area():
+            if isinstance(lbl, Gtk.Label):
+                lbl.set_halign(Gtk.Align.START)
+                lbl.set_selectable(True)
+
         if checkbox:
             cb = Gtk.CheckButton.new_with_label(checkbox)
+            cb_label = cb.get_child()
+            cb_label.set_line_wrap(True)
             cb.set_active(default_checked)
-            dlg.get_message_area().pack_end(cb, False, False, 0)
-            dlg.get_widget_for_response(Gtk.ResponseType.NO).grab_focus()
+
+            sw = Gtk.ScrolledWindow()
+            sw.set_size_request(400, -1)
+            sw.set_property('hscrollbar-policy', Gtk.PolicyType.NEVER)
+            sw.set_property('vscrollbar-policy', Gtk.PolicyType.NEVER)
+            sw.set_property('shadow-type', Gtk.ShadowType.NONE)
+            sw.add(cb)
+            sw.show_all()
+            dlg.get_message_area().pack_start(sw, True, True, 0)
+
+            # Resize ScrolledWindow after CheckButton height is allocated.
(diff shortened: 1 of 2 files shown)


tpikonen:
For some reason the checkbox label does not wrap to small enough size, so the checkbox is not visible and the labels cannot be read completely on a small screen. This happens both on gnome shell and phosh, although on desktop with gnome shell it probably does not matter if the dialog is wider than the app window.

Also the textview alone is making the dialog too wide.

tpikonen:
Actually, just removing the textView and scrolledWindows, setting the message as dialog text (`dlg.set_markup(message)`), packing the checkbox (`dlg.get_message_area().pack_start(cb, True, True, 0)`) and removing the `size-allocate` callback works fine on both desktop and mobile. I don't see why the textView is needed, as long as the labels are set to wrap.

I can push my tree with this simple version of the dialog here if needed.

elelay:
> Actually, just removing the textView and scrolledWindows, setting the message as dialog text (`dlg.set_markup(message)`), packing the checkbox (`dlg.get_message_area().pack_start(cb, True, True, 0)`) and removing the `size-allocate` callback works fine on both desktop and mobile. I don't see why the textView is needed, as long as the labels are set to wrap.

It's to allow copying the message, for bug reports and such. Is it possible with `dialog.set_markup()`

elelay:
Not great: a scrollbar on a big screen.

scroll

elelay:
Sorry I can't offer improvements :-/

tpikonen:
There is this code in this PR which makes copying work:
```
        # make title or message copy/pastable
        for lbl in dlg.get_message_area():
            if isinstance(lbl, Gtk.Label):
                lbl.set_halign(Gtk.Align.START)
                lbl.set_selectable(True)
```

auouymous:
@tpikonen Can you provide your patch for testing? Only the scrolled windows and setting size request to limit size worked for me. I dislike the fixed size textview and would love a better way.

auouymous:
@tpikonen If you subscribe to a bogus feed url, does the show_message_details() dialog cause width issues?

tpikonen:
A tree with the simple dialog I described above is [here](https://github.com/tpikonen/gpodder/tree/deldialog).

The dialog following a bogus subscription URL is also too wide.

(A potential bug: Inputing a string with numbers and with a slash in between like '1234/1234' as a subscription URL is not detected as invalid, and causes 4 x 60 s wait with an unresponsive UI.)

auouymous:
@tpikonen https://github.com/tpikonen/gpodder/commit/738cd5ec274d226a77f00c764dc7b67a6f4d5396 is missing the title inside dialog which is needed because tiling window managers don't have a titlebar, but it also doesn't wrap the dialog for me. I think gtk clamps the dialog to screen width so it works great on a small screen, and the labels wrap. However, when you have a large screen and a narrow gpodder on one side, the dialog can grow wider than gpodder and then centers over gpodder which causes one side to clip off the screen.

tpikonen:
@auouymous That seems like a window management issue. On gnome-shell a narrow gPodder on the side of the screen also pops up a dialog wider than gPodder, but it's offset from the center of the parent window, so that the dialog is fully visible.

auouymous:
> That seems like a window management issue.

Yes and no. My window manager has areas, making the monitor a view into a much larger display. A dialog is not placed off screen when it appears on the edge of the virtual resolution.

elelay:
@auouymous is setting the dialog resizable helping (`dlg.set_resizable(True)`)?
For me under i3 gtk 3.24.34 it makes the dialog twice as high for no reason.

auouymous:
Oops, I never replied to this after testing `set_resizable`. I noticed the same thing when making it resizable. I also tried setting the min/max content properties while resizable and I think they worked a little better for one of the widgets but not the other.

Making the confirmation dialog resizable with wrapping labels and saving its size (and maybe position) like other windows might be the best way. It will be limited to screen size on mobile and anyone with a narrow gpodder window near screen edge can resize it to be smaller, assuming it works. We can also do the same for the `show_message` dialog which currently sets textview width to 400px and would be too wide on mobile.

Issue #1296, opened 2022-05-31
HTTP 400 error when downloading specific podcast files


usmanatron:
Hello,

I've got a few podcast files which continuously fail to download, even after retrying etc.  Looking at the logs, I've got this output:

```
2022-05-31 18:34:25,048 [gpodder.download] ERROR: HTTP 400 while downloading "17 Marr Gan and the Wastes": 400 Client Error: Bad Request for url: https://s3.eu-west-2.amazonaws.com/sanspantsplus/Hey-DnD-Adam/17+Marr+Gan+and+the+Wastes.mp3
Traceback (most recent call last):
  File "C:\Program Files (x86)\gPodder\lib\python3.9\site-packages\gpodder\download.py", line 277, in retrieve_resume
    resp.raise_for_status()
  File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.9/site-packages/requests/models.py", line 943, in raise_for_status
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://s3.eu-west-2.amazonaws.com/sanspantsplus/Hey-DnD-Adam/17+Marr+Gan+and+the+Wastes.mp3

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files (x86)\gPodder\lib\python3.9\site-packages\gpodder\download.py", line 796, in run
    headers, real_url = downloader.retrieve_resume(self.tempname, self.status_updated)
  File "C:\Program Files (x86)\gPodder\lib\python3.9\site-packages\gpodder\download.py", line 343, in retrieve_resume
    headers, real_url = downloader.retrieve_resume(url,
  File "C:\Program Files (x86)\gPodder\lib\python3.9\site-packages\gpodder\download.py", line 279, in retrieve_resume
    raise gPodderDownloadHTTPError(url, resp.status_code, str(e))
gpodder.download.gPodderDownloadHTTPError: ('https://s3.eu-west-2.amazonaws.com/sanspantsplus/Hey-DnD-Adam/17+Marr+Gan+and+the+Wastes.mp3', 400, '400 Client Error: Bad Request for url: https://s3.eu-west-2.amazonaws.com/sanspantsplus/Hey-DnD-Adam/17+Marr+Gan+and+the+Wastes.mp3')
```

I know the files exist - going to that URL in a browser gives me the file and I can play it etc.
[...]

PR #1300, opened 2022-06-08
Add download retry without authentication if auth is set

JKAbrams:
This is a naive workaround that adds another try without authentication if the download fails. This is to work around servers where the .rss file is protected but the .mp3 files are strictly not protected (where sending the authentication causes the download to fail).
Works around bug #1296.

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -231,7 +231,7 @@ class DownloadURLOpener:
 # The following is based on Python's urllib.py "URLopener.retrieve"
 # Also based on http://mail.python.org/pipermail/python-list/2001-October/110069.html
 
-    def retrieve_resume(self, url, filename, reporthook=None, data=None):
+    def retrieve_resume(self, url, filename, reporthook=None, data=None, disable_auth=False):
         """Download files from an URL; return (headers, real_url)
 
         Resumes a download if the local filename exists and
@@ -244,7 +244,7 @@ class DownloadURLOpener:
             'User-agent': gpodder.user_agent
         }
 
-        if self.channel.auth_username or self.channel.auth_password:
+        if (self.channel.auth_username or self.channel.auth_password) and not disable_auth:
             logger.debug('Authenticating as "%s"', self.channel.auth_username)
             auth = (self.channel.auth_username, self.channel.auth_password)
         else:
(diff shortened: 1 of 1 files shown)


src/gpodder/download.py
```diff
@@ -277,7 +277,13 @@ def retrieve_resume(self, url, filename, reporthook=None, data=None):
             try:
                 resp.raise_for_status()
             except HTTPError as e:
-                raise gPodderDownloadHTTPError(url, resp.status_code, str(e))
+                if auth is not None:
+                    # Try again without authentification (bug 1296)
```
auouymous, line 281: `authentification` needs to be `authentication`, right?

src/gpodder/download.py
```diff
@@ -277,7 +277,13 @@ def retrieve_resume(self, url, filename, reporthook=None, data=None):
...
+                    self.channel.auth_username = None
+                    self.channel.auth_password = None
```
auouymous, line 283: This disables authentication in the channel and causes a failure if channel is updated without restarting gpodder.

auouymous (changes requested):

src/gpodder/download.py
auouymous: Oops, it causes permanent failure even after restarting gpodder.

JKAbrams: Thanks for the speedy review, as I said it was a rather naive fix but did not take into account how the download function is used in other parts of the program (as you discovered).
I guess the problem lies in overwriting the `self.channel.auth_username`, I assumed it was temporary but seems like I overwrite the storage. Guess what is required is a more temporary way of telling it to run without `auth` set.
Adding an additional optional argument to `retrieve_resume` named something like try_again_without_authentication would prevent this from happening, do you think  that kind of change would be a working/acceptable solution?

auouymous: I haven't looked at retrieve_resume yet, but if it directly uses the auth fields in channel then a parameter would be the best way. Perhaps name it disable_auth or disable_authentication, because the callsite is performing the "try again" part.

JKAbrams: Changed the method of retrying, fixed the typo.

auouymous: LGTM. I'm going to test it for a few days and then merge if no issues pop up.

auouymous:
Merged, thank you for the fix.

PR #1303, opened 2022-06-16
Add setting to disable find-as-you-type.

auouymous:
The new setting defaults to enabled to preserve the current behavior of opening search fields in channel and episode lists when the user begins typing. Disabling the feature ignores the keypresses and requires ctrl-f or ctrl-shift-f to open or focus the search field.


tpikonen:
How about also adding '/' as a shortcut for searching where the current focus is (either channels or episodes) when type-to-search is disabled? I'm probably going to add this kind of feature to the adaptive branch (vim-keys FTW) and it would be nice to be compatible.

PR #1304, opened 2022-06-17
gpo: Add Description and Link fields to info

tpikonen:
The `info` command in `gpo` currently shows the channel title and feed URL, but not the channel description or link. Add the description and link to the `info` output.


auouymous:
Thanks for the improvement.

PR #1305, opened 2022-06-17
Create context menus from GMenuModel

tpikonen:
A remix of / alternative to #1293. Add action definitions for context menu items and create context menus (just for channels, for now) from `menus.ui`.

### DIFF
--- a/share/gpodder/ui/gtk/menus.ui
+++ b/share/gpodder/ui/gtk/menus.ui
@@ -229,5 +229,46 @@
       </submenu>
     </submenu>
   </menu>
+  <menu id="channels-context">
+    <section>
+      <item>
+        <attribute name="action">win.updateChannel</attribute>
+        <attribute name="label" translatable="yes">Update channel</attribute>
+        <attribute name="icon">view-refresh</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.openChannelDownloadFolder</attribute>
+        <attribute name="label" translatable="yes">Open download folder</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name="action">win.markEpisodesAsOld</attribute>
+        <attribute name="label" translatable="yes">Mark episodes as old</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.channelAutoArchive</attribute>
+        <attribute name="label" translatable="yes">Archive all episodes</attribute>
+      </item>
+      <item>
+        <attribute name="action">win.refreshImage</attribute>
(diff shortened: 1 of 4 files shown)


elelay:
From a quick glance I don't see any drawback to this. Are there?

auouymous:
@elelay Did you look at the channel context menu? The icons are indented, and their text doesn't align with items that lack icons.

elelay:
> @elelay Did you look at the channel context menu? The icons are indented, and their text doesn't align with items that lack icons.

Indeed. On the other hand icons are not visible for me with the master branch.

auouymous:
Oh right, gtk3 defaults to `gtk-menu-images = 0`, setting it to `1` in gtk-3.0 settings.ini enables them. And `gtk-button-images = 1` enables button icons.

tpikonen:
Two releases have now been made after this PR and #1293 were proposed. Merging the context menu changes to the GMenuModel based menus used in the adaptive branch is starting to feel like useless busywork.

Could we move the master branch to GMenuModels, either by merging this or #1293? I can bring this PR up to date and finish the implementation if there's a consensus on merging this.

For reference, here are screenshots of how the channel context menu before and after:

Before:

context-menu-orig

After:

context-menu-new

elelay:
ok from me but I defer to @auouymous decision

auouymous:
I don't like the indented icons. Can this use the verb-icons like popovers? We could remove the indented icons and add verb-icons at the top for the most common actions.

The popovers lack any of the flaws that normal context menus have but @elelay didn't like them (https://github.com/gpodder/gpodder/pull/1293#issuecomment-1214435466) so they are out.

tpikonen:
Ok, I dove into GTK docs and sources and here's what I found:

 * The `verb-icon` attributes are used only in GtkPopovers. GtkMenus ignore it, so they cannot have horizontal button rows.
 * The non-indented icons are a feature of the deprecated GtkImageMenuItem, which is what gPodder currently uses. The GMenuModel based menus are composed of GtkModelMenuItems which can have both a check box (determined from the action type) and an icon (from the `icon` attribute), so this way all the icons will be aligned.

So, with GMenuModels the context menu will either be a GtkMenu with indented icons, or a GtkPopover with icons only in the horizontal button rows. I'm fine with either, but would also prefer the popovers.

elelay:
Thanks for looking into it.

@auouymous @tpikonen you are both convinced that popover is better so I change my position: let's go for popovers.

luzpaz:
Is there an ETA on this ?

auouymous:
@luzpaz #1293 will be merged once the injected extensions are fixed and a couple other minor fixes.

PR #1308, opened 2022-06-22
Wait for worker threads to finish before quitting.

### LINKED ISSUE #1306: blocks on exit when there are downloads in progress
There is an issue on all versions: Linux, windows (7 and 10).

gPodder freezes on exit when there were active downloads.

I thought it was related to the taskbar_progress extension (see #1211 ) but it's still happening when the extension is disabled.
It must be a timing issue of some sort because pausing in pdb sometimes prevents blocking.

I haven't had time to investigate if it was present in last release...
To reproduce, download an episode and close the window.
A popup warns you that there are active downloads; confirm via the Quit button.
Then the UI disappears but gpodder is still running. It can only be killed by task manager.

auouymous:
Closing gpodder while tasks are downloading or pausing causes a hang and must be killed from task manager or interrupted if launched from command line. This was introduced in ed5d18e1b04f5e32b18d90b3da151a57ae507571 but only applicable when using python 3.10. The idle_add() callback is no longer invoked to wake up the dequeue wait_for() condition, likely due to the main thread terminating early, which didn't happen in older python versions. Simply waiting for all worker threads to terminate fixes the issue.

Fixes #1306.

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -471,6 +471,9 @@ class DownloadQueueManager(object):
         self.tasks.queue_task(task)
         self.__spawn_threads()
 
+    def has_workers(self):
+        return len(self.worker_threads) > 0
+
 
 class DownloadTask(object):
     """An object representing the download task of an episode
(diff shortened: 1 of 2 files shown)


auouymous:
I tried to fix the rate limited pause transition delay a while back but added https://github.com/gpodder/gpodder/blob/master/src/gpodder/download.py#L640 instead.

PR #1320, opened 2022-07-05
util.open_website: Return True to block 'activate-link' default handler

tpikonen:
Clicking a web link on a GtkLabel (there is one at least in Preferences->youtube-dl) opens the web page twice, from util.open_website and from GtKLabels 'activate-link' default handler.

Return True from util.open_website, so that the default handler will not be run.


auouymous:
I don't know why my gtk doesn't call the default activate-link handler, but the docs say it does so we should return True.

Thank you.

PR #1322, opened 2022-07-05
Update Norwegian Nynorsk translation

huftis:


po/nn.po
```diff
@@ -2100,12 +2102,15 @@ msgid ""
 "<b>Download with youtube-dl</b> appears in the episode menu when this option "
 "is disabled, and can be used to manually download from supported sites."
 msgstr ""
+"Programmet youtube-dl gjev tilgang til fleire YouTube-format og DRM-innhald.  "
+"Merk at episodar frå kanalar på andre nett­stadar enn YouTube, men som er støtta av youtube-dl,"
+"<b>ikkje</b> vert lasta ned, med mindre du manuelt legg til eit <a href=\"https://gpodder.github.io/docs/youtube.html#formats\">tilpassa format</a> for kvar nettstad.  "
+"<b>Last ned med youtube-dl</b> vert vist i episode­menyen når denne utvidinga er slått av, slik at du kan lasta ned manuelt frå desse nettstadane."
```
auouymous, line 2108: Does the last line say the menu item will be shown when the youtube-dl extension is disabled? If so, it should say when the manage_downloads option the message sits below is disabled.

auouymous (changes requested):

po/nn.po
huftis: Yes, you’re right (in that the translation refers to the extension). I had trouble triggering the display of this string, so I couldn’t check how it appeared. And I still have problems triggering it. Any ideas? If I try to enable the youtube-dl extension in the Preferences dialog, it’s automatically disabled when I close an reopen the dialog.

auouymous: It should work fine installed by distro. Does gpodder's log mention anything related to youtube-dl after you try enabling it?

huftis: After installing youtube-dl using pip, everything is working fine. I had originally installed gpodder using `make install PREFIX=/usr`. When I tried using `setup.py` instead, I got a gpodder install which actually showed an error (dialog) about a missing Python `youtube-dl` module when I tried to enable the extension. (This did *not* happen with my original install.) Anyway, after I installed youtube-dl using pip, everything is OK.

I have updated the translation (including a small improvement in the missing Python module error message). I had some trouble phrasing the text to make it clear that it was referring to the checkbox above, but I think it should be OK now. Thanks for doing the extra QA on the translation! 😃

auouymous:
Thanks for the translation, and thanks for the youtube-dl capitalization issues in #518.

PR #1330, opened 2022-07-08
Add/remove extension preferences buttons when loaded/unloaded.

auouymous:
The user must currently close preferences and reopen to configure extensions that provide a preferences UI. This is a problem when the user does not know the extension has preferences, but the sudden appearance of the button when an extension is enabled might catch the user's attention.

Furthermore, removing the button when the extension is unloaded prevents any confusion as to why it is still there.

### DIFF
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -321,21 +321,37 @@ class gPodderPreferences(BuilderWidget):
 
         gpodder.user_extensions.on_ui_object_available('preferences-gtk', self)
 
+        self.inject_extensions_preferences(init=True)
+
+        self.prefs_stack.foreach(self._wrap_checkbox_labels)
+
+    def _wrap_checkbox_labels(self, w, *args):
+        if w.get_name().startswith("no_label_wrap"):
+            return
+        elif isinstance(w, Gtk.CheckButton):
+            label = w.get_child()
+            label.set_line_wrap(True)
+        elif isinstance(w, Gtk.Container):
+            w.foreach(self._wrap_checkbox_labels)
+
+    def inject_extensions_preferences(self, init=False):
+        if not init:
+            # remove preferences buttons for all extensions
+            assert(self.prefs_stack.get_visible_child_name() == 'extensions')
+            found_extensions = False
+            for child in self.prefs_stack.get_children():
+                if found_extensions:
+                    self.prefs_stack.remove(child)
+                elif child == self.prefs_stack.get_visible_child():
+                    found_extensions = True
+
+        # add preferences buttons for all extensions
         result = gpodder.user_extensions.on_preferences()
         if result:
             for label, callback in result:
                 self.prefs_stack.add_titled(callback(), label, label)
 
-        def _wrap_checkbox_labels(w, *args):
-            if w.get_name().startswith("no_label_wrap"):
-                return
-            elif isinstance(w, Gtk.CheckButton):
-                label = w.get_child()
-                label.set_line_wrap(True)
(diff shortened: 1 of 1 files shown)


src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -321,21 +321,37 @@ def new(self):
...
+
+    def inject_extensions_preferences(self, init=False):
+        if not init:
+            # remove preferences buttons for all extensions
+            assert(self.prefs_stack.get_visible_child_name() == 'extensions')
+            found_extensions = False
+            for child in self.prefs_stack.get_children():
+                if found_extensions:
+                    self.prefs_stack.remove(child)
+                elif child == self.prefs_stack.get_visible_child():
```
tpikonen, line 345: This relies on the 'Extensions' page being the last item in the Stack before the extension pages. Maybe prepend `"extension."` to the `name` of the extension stack children, and remove only those which have this prefix?

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -321,21 +321,37 @@ def new(self):
...
+                    found_extensions = True
+
+        # add preferences buttons for all extensions
         result = gpodder.user_extensions.on_preferences()
         if result:
             for label, callback in result:
                 self.prefs_stack.add_titled(callback(), label, label)
```
tpikonen, line 352: This line would then become `self.prefs_stack.add_titled(callback(), "extension." + label, label)`

src/gpodder/gtkui/desktop/preferences.py
auouymous: Do you know how to do this? This is a much better solution but I couldn't find a way to iterate over names, and get_child_by_name() would need the full extension name. I could split the removal code to another function called before the enabled/disabled extension has been added/removed to/from user_preferences, and then iterate over that to remove each extension.

tpikonen: Ok, it did not quite work as I thought, but setting the widget name to something recognizable works. See this patch:

```diff
diff --git a/src/gpodder/gtkui/desktop/preferences.py b/src/gpodder/gtkui/desktop/preferences.py
index 68550c13..e84ed898 100644
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -337,21 +337,19 @@ class gPodderPreferences(BuilderWidget):
     def inject_extensions_preferences(self, init=False):
         if not init:
             # remove preferences buttons for all extensions
-            assert(self.prefs_stack.get_visible_child_name() == 'extensions')
-            found_extensions = False
             for child in self.prefs_stack.get_children():
-                if found_extensions:
+                if child.get_name().startswith("extension."):
                     self.prefs_stack.remove(child)
-                elif child == self.prefs_stack.get_visible_child():
-                    found_extensions = True
 
         # add preferences buttons for all extensions
         result = gpodder.user_extensions.on_preferences()
         if result:
             for label, callback in result:
-                self.prefs_stack.add_titled(callback(), label, label)
-
-            self.prefs_stack.foreach(self._wrap_checkbox_labels)
+                page = callback()
+                name = "extension." + label
+                page.set_name(name)
+                page.foreach(self._wrap_checkbox_labels)
+                self.prefs_stack.add_titled(page, name, label)
 
     def _extensions_select_function(self, selection, model, path, path_currently_selected):
         return model.get_value(model.get_iter(path), self.C_SHOW_TOGGLE)
```

auouymous: Thank you, I have pushed your patch.

PR #1334, opened 2022-07-12
libgpod_ctypes: Use ctypes.c_time_t on Python 3.12

thp:
On top of #1289, the PR that adds `c_time_t` has been merged to CPython and should become available in 3.12.


auouymous:
Thanks, that will prevent issues on 32bit when 3.12 arrives.

PR #1336, opened 2022-07-12
preferences: Show extension context menu on row-activation

### LINKED ISSUE #1329: Visible link to extension help
Currently it's behind a right-click with no indication one should right-click in the first place.
We should find something better

tpikonen:
Change the 'Extensions' page context menu to a GtkPopover and make it pop-up when the `row-activation` signal of the Treeview is emitted (double click, return-key press), on long-press gesture and on right mouse click.

Also widen the toggles on the rows to make them more mobile friendly.

Possibly fixes #1329. A button popping up the context menu would be perhaps better, but there is no CellRenderer for that.

### DIFF
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -388,6 +388,15 @@ class gPodderPreferences(BuilderWidget):
         extension_column.set_expand(True)
         self.treeviewExtensions.append_column(extension_column)
 
+        def activate_to_context_menu(treeview, path, column):
+            if column == extension_column:
+                self.on_treeview_extension_show_context_menu(treeview)
+            else:
+                return
+
+        self.treeviewExtensions.connect('row-activated',
+            activate_to_context_menu)
+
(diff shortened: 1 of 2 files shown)


elelay:
> There is no 'GtkCellRendererButton' class, or similar, see [here](https://docs.gtk.org/gtk3/class.CellRenderer.html#descendants). There is `GtkCellRendererPixbuf`, but I don't see a way to make it clickable.

Maybe switch to `Gtk.ListBox` ? It's made for a list with additional interactions

tpikonen:
> Maybe switch to Gtk.ListBox ? It's made for a list with additional interactions

That seems to be the only way to get a clickable button. The rows will then contain something like [HdyActionRow](https://gnome.pages.gitlab.gnome.org/libhandy/doc/1-latest/class.ActionRow.html) widgets. If libhandy cannot be included as a dependency in master, then they will have to be re-implemented, like in `gpodderchannel.ui`.

tpikonen:
Rebased to master and removed the payment items also from the new popover menus.

This still needs a review. @elelay, would you have time for this?

PR #1343, opened 2022-07-29
Draft: Chapters UI

elelay:
2 takes on a chapters UI:

1. in textual shownotes, list chapters with start timestamp clickable to jump to (but lacks the actual jump-to)
2. with enqueue_in_mediaplayer extension, add a Chapters submenu with jump-to functionality 
    problem with that: only VLC seems to work, and semi-crashes when offset is too big.

Test feed: http://cre.fm/feed/m4a

This is only a little draft but feedback is welcome :-)
I want to replace the textual list in shownotes with a Gtk.Listbox for proper formatting and because I don't know what it'll look like in html shownotes. Navigating between shownotes and chapters is to be defined. Maybe a link/button beside the title to switch between widgets in a stack?


auouymous:
I don't have dbus or an MPRIS player to test the enqueue functionality but the text shownotes look good, and aligning the timestamps would look better. Youtube channels often stick timestamps in the description, so the current chapters implementation blends in and would be fine until something better was added in the future.

Would the shownotes timestamps call into the enqueue extension and fail with a dialog if not installed, or would the list of resumers be moved into gpodder?

elelay:
I've added a playbar to jump to chapters. It also has a button triggering the display of chapters in a dialog:
proposed UI

The dialog is functional, albeit with fake images and links in order to see how it looks with those. Clicking on timestamps will cause the currently playing episode to jump to the timestamp, using MPRIS2.

How to integrate the dialog contents with shownotes? I've kept the textual view in textual shownotes for the moment. Maybe adding the chapters widget after the shownotes so we see them when scrolling down?

elelay:
Here is the UI with MPRIS2 enabled and VLC playing the episode: notice the ticks showing chapter starts and current chapter's title shown under the playbar.

screenshot-2024-08-31-133354

auouymous:
> Maybe adding the chapters widget after the shownotes so we see them when scrolling down?

Is the widget that button in playbar or the contents of the dialog?

Does the playbar button open a dialog if nothing is playing? If so, I assume it opens for current episode. But when something is playing, does it only show the playing episode in dialog, or does it show the selected episode. Do you have to first stop the playing episode before jumping to chapters in another?

I often open multiple episodes in separate instances of mplayer or mpv. I assume if I had MPRIS, gpodder would show my progress in the tooltip of each episode. How does the playbar handle multiple instances?

Ideally the playbar would be tied to the selected episode to avoid the above issues. But then it probably needs a button that selects first playing episode if the current selection is not playing, or the next playing episode if the selected episode is playing. And maybe a previous button as well.

elelay:
Thanks for looking into this.

> > Maybe adding the chapters widget after the shownotes so we see them when scrolling down?
> 
> Is the widget that button in playbar or the contents of the dialog?

It is the contents of the dialog: vertical list of image, chapter title, link

> 
> Does the playbar button open a dialog if nothing is playing? If so, I assume it opens for current episode. But when something is playing, does it only show the playing episode in dialog, or does it show the selected episode. Do you have to first stop the playing episode before jumping to chapters in another?
> 

The button always shows the selected episode's shownotes. The playbar always shows currently playing episode.

> I often open multiple episodes in separate instances of mplayer or mpv. I assume if I had MPRIS, gpodder would show my progress in the tooltip of each episode. How does the playbar handle multiple instances?

Hopefully you don't listen to all of them simultaneously ;-)

Still the playbar will switch to the last play event: 
1. start playing episode 1 => episode 1 in playbar
2. start playing episode 2 in another player => episode 2 in playbar
3. pause episode 1 => episode 1 in playbar
4. pause or jump to time in episode 2 in player => episode 2 in playbar

It still lacks regularly querying the player: I want to save progress every 10s. That way 10s after step "3." above it would show episode 2 again, without fiddling with player controls...


> Ideally the playbar would be tied to the selected episode to avoid the above issues. But then it probably needs a button that selects first playing episode if the current selection is not playing, or the next playing episode if the selected episode is playing. And maybe a previous button as well.

That is many buttons ;-)

Maybe the button to the right of the playbar would be the "Go to shownotes of playing episode" instead of showing episode chapters in a dialog and the chapters should be together with the shownotes?

The button to the right of the playbar could be a dropdown/popover with all currently open episode playbacks to address the "multiple episodes open" scenario? We would have to figure out a way to meaningfully shorten the episode title to fit in a popover. Clicking on the episode would make the playbar display it and go to shownotes.

elelay:
rebased on top of #1792

PR #1347, opened 2022-08-01
Don't try to download media files with youtube-dl extension

### LINKED ISSUE #1263: Spreaker feeds not working - [EDIT - youtube-DL extension is the culprit]

tpikonen:
gPodder calls `registry.custom_downloader.resolve` for every episode/URL when a download is attempted. Some episode URLs which youtube-dl / yt-dlp claims to support are direct media links, and at least yt-dlp is buggy so that it does not in fact download anything when given such an URL. At least spreaker.com (bug #1263) and audioboom.com fail this way.

Change the youtube-dl extension so that it does not support episodes where the MIME type is `audio/*` or `video/*`. These episodes are then downloaded with the default downloader.

Fixes #1263.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -426,6 +426,8 @@ class gPodderYoutubeDL(download.CustomDownloader):
         """
         if not self.force and not self.my_config.manage_downloads:
             return None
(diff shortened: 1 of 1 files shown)


auouymous:
Thanks for the fix.

PR #1353, opened 2022-08-03
Fix hang when resuming thousands of episodes.

### LINKED ISSUE #1311: Getting error: expected string or bytes-like object
image
Also the problem is that it reports interminable errors, so I have to click on all the problem items messages one by one, maybe there should be a "yes for all".

auouymous:
When the number of episodes causes update_downloads_list() to exceed 1.5 seconds, the idle_add callback will never be called and cause gpodder to hang.

Fixes #1311.

### DIFF
--- a/src/gpodder/gtkui/download.py
+++ b/src/gpodder/gtkui/download.py
@@ -193,7 +193,10 @@ class DownloadStatusModel(Gtk.ListStore):
     # as only the main thread is allowed to manipulate the list store.
     def get_next(self):
         dqr = DequeueRequest()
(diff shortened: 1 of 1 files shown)


elelay:
Thanks for the fix!
Shouldn't it be `Glib.timeout_add(GLib.PRIORITY_DEFAULT, 0, lambda: ...)` because gobject.timeout_add seems deprecated?

auouymous:
@elelay Ya, I've had a note to replace the deprecated GObject.timeout_add and GObject.idle_add calls for a while now, and will get to it after this is fixed, which it sadly isn't yet.

PR #1355, opened 2022-08-03
Keep 'find partial downloads' dialog open longer.

auouymous:
It can take a while to populate the list when there are a lot of downloads to resume. This leaves the list empty and UI, including the `Resume All` button, unresponsive. Keeping the dialog open while Gtk populates the list prevents this from happening.


elelay:
LGTM

elelay:
Merged, thanks

PR #1356, opened 2022-08-04
Reset window position if completely off-screen.

### LINKED ISSUE #1352: 3.11.0 Soft crashing when checking for new podcasts
Every time the program tries to check for new podcasts it won't crash but I won't be able to click on anything. When I mouse over the X and minimize buttons they will highlight like they are active but won't do anything when click on them so I need to use task manager to end the program. I've tried uninstalling and reinstalling. I'm on Windows 10.

auouymous:
This requires one pixel of any side of the window to be on-screen, or it resets the position to default. It might be more useful if it required ten or more pixels so the window would be more visible.

Fixes #1352.

### DIFF
--- a/src/gpodder/gtkui/config.py
+++ b/src/gpodder/gtkui/config.py
@@ -22,6 +22,8 @@
 #  gpodder.gtkui.config -- Config object with GTK+ support (2009-08-24)
 #
 
+import logging
+
 import gi  # isort:skip
 gi.require_version('Gdk', '3.0')  # isort:skip
 gi.require_version('Gtk', '3.0')  # isort:skip
@@ -29,6 +31,8 @@ from gi.repository import Gdk, Gtk, Pango
 
 import gpodder
 from gpodder import config, util
+
+logger = logging.getLogger(__name__)
 
 _ = gpodder.gettext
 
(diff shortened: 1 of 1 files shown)


auouymous:
Or maybe at least 1/4 to 1/2 of the width and height to be visible.

auouymous:
I changed it to reset if more than 50% of the window is off-screen, so enough is visible to see it. I double anyone has 50% of the window off-screen because it is unlikely that even narrow windows would be usable with only 50% visible.

src/gpodder/gtkui/config.py
```diff
@@ -156,6 +160,25 @@ def connect_gtk_window(self, window, config_prefix, show_window=False):
         if gpodder.ui.win32:
             window.set_gravity(Gdk.Gravity.STATIC)
 
+        if cfg.x != -1 and cfg.y != -1 and cfg.width != -1 and cfg.height != -1:
```
elelay, line 163:
```suggestion
        if -1 not in (cfg.x, cfg.y, cfg.width, cfg.height):
```

elelay (approved):
LGTM.
minor syntax change to make it more concise

src/gpodder/gtkui/config.py
auouymous: Done, thanks.

elelay:
Merged, thanks

PR #1357, opened 2022-08-04
[shebang] Change /usr/bin/python -> /usr/bin/env python3

thp:
This should fix "unusual interpreter" Lintian warnings here:
https://udd.debian.org/lintian/?packages=gpodder

### DIFF
--- a/share/gpodder/examples/gpodder_mark_played.py
+++ b/share/gpodder/examples/gpodder_mark_played.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # Example script that can be used as post-play extension in media players
 #
 # Set the configuration options "audio_played_dbus" and "video_played_dbus"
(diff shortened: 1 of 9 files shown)


thp:
(not all of these scripts need or should have a shebang line to begin with, but this is the least intrusive change)

auouymous:
We might as well remove the shebang now if not needed.

Need shebang:
- share/gpodder/examples/gpodder_mark_played.py
- src/gpodder/minidb.py

Don't need shebang:
- share/gpodder/extensions/rm_ogg_cover.py
- share/gpodder/extensions/rockbox_convert2mp4.py
- share/gpodder/extensions/tagging.py
- src/gpodder/dbusproxy.py
- src/gpodder/gtkui/widgets.py
- src/gpodder/my.py
- src/gpodder/plugins/soundcloud.py

auouymous:
Thanks for the fixes.

tmancill (commented):
Thanks!

PR #1363, opened 2022-08-09
Embed subtitles to videos downloaded with the youtube-dl extension

### LINKED ISSUE #379: [Wish] Subtitles/Transcript for YouTube videos
I would like to download the transcript/subtitles from YouTube videos with my subscription.

Thanks.

tpikonen:
Fixes #379.

There are some problems with the youtube.py and youtube-dl extension interaction, plus this breaks downloads if ffmpeg is not installed, so setting as draft for now.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -47,6 +47,8 @@ DefaultConfig = {
     # If for some reason youtube-dl download doesn't work for you, you can fallback to gpodder code.
     # Set to False to fall back to default gpodder code (less available formats).
     'manage_downloads': True,
+    # Embed all available subtitles to downloaded videos. Needs ffmpeg.
+    'embed_subtitles': False,
 }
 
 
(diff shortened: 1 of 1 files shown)


tpikonen:
Rebased and adapted to current master.

tpikonen:
So, now that #1382 is merged, this reduced to just a couple of lines. Subtitle embedding works fine when ffmpeg is installed, and even previews work with non-fragment downloads in youtube.

When ffmpeg is not installed, this probably breaks most youtube-dl downloads. The solution is to either detect if ffmpeg is installed, or to put this behind a config var. Any ideas on which one is better?

auouymous:
The extension should add a setting that defaults to off to avoid putting subtitles in downloads. When enabled, it should use util.find_command('ffmpeg') and display an error message before disabling the setting if not found.

tpikonen:
Rebased to master and added a checkbox to enable subtitle embedding in the youtube-dl preferences page. If ffmpeg is not found, prevent enabling and show an infobox to inform the user about the requirement.

auouymous:
The checkbox and separator should sit below the note that is specifically for the manage-downloads checkbox.

tpikonen:
Ah, thanks for noticing. Fixed now.

auouymous:
Thanks for the improvement.

Issue #1378, opened 2022-08-23
Add WinGet Releaser github action to this repositiory


aman207:
Hi,
I'd like to suggest that the WinGet Releaser github action be added to this repo. It can be found on the marketplace here https://github.com/marketplace/actions/winget-releaser.

This action would automatically add new releases of gPodder to the WinGet repository. WinGet is a tool developed by Microsoft to install & upgrade Windows packages via the CLI. gPodder is already in said repository, however it would be beneficial if updates were automatically added instead of manually adding them when a new released is issued.

Let me know your thoughts. Thanks

auouymous:
It would be a nice feature for Windows users if you want to submit a PR.

PR #1380, opened 2022-08-28
Add winget releaser github action

aman207:
Discussion in #1378 

You will need to create a github personal access token and add it to the secrets within the repository. More info here (https://bittu.eu.org/docs/wr-intro) & here (https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow)

### DIFF
--- a/.github/workflows/winget-releaser.yml
+++ b/.github/workflows/winget-releaser.yml
@@ -6,7 +6,7 @@ jobs:
   publish:
     runs-on: windows-latest # action can only be run on windows
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@0727cc2cdf0f3ecafd492608d2aff67b13adff55
+      - uses: vedantmgoyal2009/winget-releaser@1492f6c8620789a03865436b2eb95c6a944f8405
         with:
           identifier: gpodder.gpodder
           version-regex: '[0-9.]+'
--- a/.github/workflows/winget-releaser.yml
+++ b/.github/workflows/winget-releaser.yml
@@ -10,6 +10,6 @@ jobs:
         with:
           identifier: gpodder.gpodder
           version-regex: '[0-9.]+'
-          installers-regex: 'installer\.exe$' # only installer.exe file
+          installers-regex: 'installer\.exe$' # only the installer.exe file
           delete-previous-version: 'false'
           token: ${{ secrets.WINGET_TOKEN }}
--- a/.github/workflows/winget-releaser.yml
+++ b/.github/workflows/winget-releaser.yml
@@ -6,7 +6,7 @@ jobs:
   publish:
     runs-on: windows-latest # action can only be run on windows
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@latest
+      - uses: vedantmgoyal2009/winget-releaser@0727cc2cdf0f3ecafd492608d2aff67b13adff55
         with:
           identifier: gpodder.gpodder
           version-regex: '[0-9.]+'
--- a/.github/workflows/winget-releaser.yml
+++ b/.github/workflows/winget-releaser.yml
@@ -1,0 +1,15 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    runs-on: windows-latest # action can only be run on windows
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@latest
+        with:
+          identifier: gpodder.gpodder
+          version-regex: '[0-9.]+'
+          installers-regex: 'installer\.exe$' # only installer.exe file
+          delete-previous-version: 'false'
+          token: ${{ secrets.WINGET_TOKEN }}
--- a/.github/workflows/winget-releaser.yml
+++ b/.github/workflows/winget-releaser.yml
@@ -11,5 +11,4 @@ jobs:
           identifier: gpodder.gpodder
           version-regex: '[0-9.]+'
           installers-regex: 'installer\.exe$' # only the installer.exe file
-          delete-previous-version: 'false'
           token: ${{ secrets.WINGET_TOKEN }}


.github/workflows/winget-releaser.yml
```diff
@@ -0,0 +1,15 @@
...
+  publish:
+    runs-on: windows-latest # action can only be run on windows
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@latest
+        with:
+          identifier: gpodder.gpodder
+          version-regex: '[0-9.]+'
+          installers-regex: 'installer\.exe$' # only the installer.exe file
+          delete-previous-version: 'false'
+          token: ${{ secrets.WINGET_TOKEN }}
```
vedantmgoyal2009, line 15:
```suggestion
          installers-regex: 'installer\.exe$' # only the installer.exe file
          token: ${{ secrets.WINGET_TOKEN }}
```

vedantmgoyal2009 (changes requested):

vedantmgoyal2009:
cc @auouymous

elelay:
Call me paranoid but I'm not confident entrusting a private token to a 3rd party app.

public_repo | Limits access to  public repositories. That includes read/write access to code, commit  statuses, repository projects, collaborators, and deployment statuses  for public repositories and organizations. Also required for starring  public repositories.
-- | --

auouymous:
I haven't had time to look at this yet, but if it really does need write access then I'd say no. We already run github_release.py with a public_repo token that could easily be modified to push a winget release without exposing the repo to any third parties.

vedantmgoyal2009:
The personal access token is used to create a pull request on https://github.com/microsoft/winget-pkgs to update the manifests and not for any other purposes.

elelay:
Yes it is intended to be used to open a PR, but it is consumed by an unspecified version (`@latest`) of the app, which then calls an [unspecified version of a powershell](https://github.com/vedantmgoyal2009/winget-releaser/blob/main/index.js#L65).

elelay (changes requested):
Due to security concerns with giving a personal token I'm against this addition.

elelay:
> > Yes it is intended to be used to open a PR, but it is consumed by an unspecified version (`@latest`) of the app, which then calls an [unspecified version of a powershell](https://github.com/vedantmgoyal2009/winget-releaser/blob/main/index.js?rgh-link-date=2022-09-14T17%3A46%3A08Z#L65).
> 
> It is the PowerShell script for updating manifests using data from previous manifests. The source of the file can be found in the root of the action's repository (https://github.com/vedantmgoyal2009/winget-releaser/blob/main/YamlCreate.ps1).
> 
> The `latest` in the URL is to use the script from `latest` tag, since the `main` branch has untested changes, which may break the action.

To be clear, based on a cursory glance I didn't find anything wrong with your app (I'm not able to judge the powershell dependencies). My issue is with future changes. I was going to request thay you reference a precise application version and a precise tag in your app instead of `latest` but a malicious actor could just as well force-push new tags to your repository. Immutable github apps would be necessary to reassure me.

aman207:
The release is actually called latest as you can see here: https://github.com/vedantmgoyal2009/winget-releaser/releases/tag/latest

According to the github documentation, anything after @ is referencing a tag or branch on the target action (https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management). So it seems like its not actually referencing the latest version of the code but rather a tag named 'latest'

aman207:
Regarding the concerns with a personal access token, could we create another github account with no ties to this project? There just needs to be _an_ account that sends a pull request to winget-pkgs

elelay:
> We could also reference the SHA of a specific commit (https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-a-commits-sha-for-release-management). Then its impossible for an attacker to takeover the tag or branch and publish malicious code.

This might be a good idea. And reference that in https://github.com/vedantmgoyal2009/winget-releaser/blob/latest/index.js#L72 because the app currently downloads anything with the tag latest. But now you get to commit twice and re-push your "latest" tag at any change in the powershell, not the most convenient...

aman207:
Can't speak to the edit of index.js, that'll have to be something @vedantmgoyal2009 speaks to. I added a commit with the most recent SHA on the 'latest' tag.

vedantmgoyal2009:
> This might be a good idea. And reference that in [https://github.com/vedantmgoyal2009/winget-releaser/blob/latest/index.js#L72](https://github.com/vedantmgoyal2009/winget-releaser/blob/latest/index.js?rgh-link-date=2022-09-24T21%3A57%3A49Z#L72) because the app currently downloads anything with the tag latest

That can be solved via `${{ github.action_ref }}`

Reference: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/workflows/winget-releaser.yml
```diff
@@ -0,0 +1,14 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    runs-on: windows-latest # action can only be run on windows
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@0727cc2cdf0f3ecafd492608d2aff67b13adff55
```
vedantmgoyal2009, line 9:
```suggestion
      - uses: vedantmgoyal2009/winget-releaser@1492f6
```

vedantmgoyal2009: https://github.com/vedantmgoyal2009/winget-releaser/blob/1492f6c8620789a03865436b2eb95c6a944f8405/index.js#L68

cc @elelay

vedantmgoyal2009:
https://github.com/vedantmgoyal2009/winget-releaser/blob/1492f6c8620789a03865436b2eb95c6a944f8405/index.js#L68

cc @elelay @aman207

PR #1381, opened 2022-08-30
Revert "Don't try to download media files with youtube-dl extension"

### LINKED ISSUE #1364: Does not honor 'youtube-dl.manage-downloads' setting with youtube channels
I have `extensions.youtube-dl.manage-downloads` config var set to True, but downloads from youtube still use the (throttled) code from `youtube.py`. Downloads fail completely with `gpodder.youtube.YouTubeError: No preferred formats found` if custom formats are enabled. This is with current master.

auouymous:
Reverts gpodder/gpodder#1347

Unfixes #1263.
Fixes #1364.
May fix unsupported DRM issue in #1311.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -428,8 +428,6 @@ class gPodderYoutubeDL(download.CustomDownloader):
         """
         if not self.force and not self.my_config.manage_downloads:
             return None
-        if episode.mime_type.startswith(('audio/', 'video/')):
-            return None
         if self.is_supported_url(episode.url):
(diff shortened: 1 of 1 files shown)


tpikonen:
Yes, this should be reverted. The spreaker feeds seem to be working now and I submitted a fix to the yt-dlp audioboom extension, which was the other failing example in #1263.

A more generic fix to #1263 could be to use `util.filename_from_url()` to detect direct media links in the youtube_dl extension.

auouymous:
The spreaker feed is failing to find formats with this PR applied, so I've reopened #1263.

PR #1382, opened 2022-08-31
Fix yt-dlp output file name

tpikonen:
Get the output file name from the yt-dlp result dict and rename the file to what is expected (`tempname`), if it is different.

Add the extension to the output template. This is needed to unbreak post-processing (like subtitle embedding). See yt-dlp issue 1844.

Set the 'paths' param in ydl to get all downloaded files to the correct location.

In addition, the first commit makes `util.sanitize_filename_ext()` to return a file name without an ending perdiod, if ext is empty.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -85,6 +85,15 @@ class YoutubeCustomDownload(download.CustomDownload):
         self._reporthook = None
         self._prev_dl_bytes = 0
         self._episode = episode
+        self._partial_filename = None
+
+    @property
+    def partial_filename(self):
+        return self._partial_filename
+
+    @partial_filename.setter
+    def partial_filename(self, val):
+        self._partial_filename = val
 
     def retrieve_resume(self, tempname, reporthook=None):
         """
@@ -94,7 +103,14 @@ class YoutubeCustomDownload(download.CustomDownload):
         # outtmpl: use given tempname by DownloadTask
         # (escape % because outtmpl used as a string template by youtube-dl)
         outtmpl = tempname.replace('%', '%%')
-        res = self._ytdl.fetch_video(self._url, outtmpl, self._my_hook)
+        info, opts = self._ytdl.fetch_info(self._url, outtmpl, self._my_hook)
+        if program_name == 'yt-dlp':
+            self.partial_filename = os.path.join(
+                opts['paths']['home'], opts['outtmpl']['default']) % info
+        elif program_name == 'youtube-dl':
+            self.partial_filename = opts['outtmpl'] % info
+
+        res = self._ytdl.fetch_video(info, opts)
         if program_name == 'yt-dlp':
             # yt-dlp downloads to whatever file name it wants, so rename
             filepath = res.get('requested_downloads', [{}])[0].get('filepath')
@@ -312,7 +328,7 @@ class gPodderYoutubeDL(download.CustomDownloader):
             opts['format'] += '/' + fallback
         logger.debug('format=%s', opts['format'])
 
-    def fetch_video(self, url, tempname, reporthook):
+    def fetch_info(self, url, tempname, reporthook):
         opts = {
             'paths': {'home': os.path.dirname(tempname)},
             # Postprocessing in yt-dlp breaks without ext
@@ -325,7 +341,12 @@ class gPodderYoutubeDL(download.CustomDownloader):
         opts.update(self._ydl_opts)
         self.add_format(self.gpodder_config, opts)
         with youtube_dl.YoutubeDL(opts) as ydl:
-            return ydl.extract_info(url, download=True)
+            info = ydl.extract_info(url, download=False)
+            return info, opts
+
+    def fetch_video(self, info, opts):
+        with youtube_dl.YoutubeDL(opts) as ydl:
+            return ydl.process_video_result(info, download=True)
 
     def refresh_entries(self, ie_result):
         # only interested in video metadata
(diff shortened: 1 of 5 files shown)


auouymous (changes requested):

auouymous:
It still spams `WARNING: --paths is ignored since an absolute path is given in output template` to console after every download. Do you see any issues with setting outtmpl to `os.path.basename(tempname) + '.%(ext)s'`?

auouymous:
We also need to decide if youtube-dl support should be dropped or if this patch needs to be modified to use the old renaming code for youtube-dl, and new code for yt-dlp. youtube-dl doesn't support --paths, or complain about it, but outtmpl would need the absolute path for it to work. Then there is the renaming code that uses the unsupported filepath in youtube-dl. I'd say drop it due to no releases this year, but it does appear active and might get a release eventually.

tpikonen:
Do you get the '--paths' warnings only with youtube-dl (not yt-dlp)? Admittedly I did not test with youtube-dl.

The support for original youtube-dl could be dropped, IMO.

auouymous:
> Do you get the '--paths' warnings only with youtube-dl (not yt-dlp)?

Only yt-dlp throws the warning, for every download. youtube-dl doesn't know about it and has no code to warn.


> The support for original youtube-dl could be dropped, IMO.

There is a chance that some distro doesn't have yt-dlp yet and would cause problems for users. We could drop it in January if they don't get a release out this year, but it might be good to continue supporting it until then.

There is a new `program_name` variable that could be used to switch code.

tpikonen:
Strangely I don't see the warning here, although when looking at yt-dlp code I should. There should be no problem in using `os.path.basename(tempname) + '.%(ext)s'` as `outtmp`.

I'll look into switching the rename code on `program_name`.

tpikonen:
Now has separate renaming for yt-dlp and youtube-dl, uses `os.path.basename(tempname)` and only adds '.%(ext)s' for yt-dlp.

Rebased to master, needs squashing later.

tpikonen:
I only tested this with yt-dlp, so please test with youtube-dl.

elelay:
Not tested, but code looks good. Thanks!

share/gpodder/extensions/youtube-dl.py
```diff
@@ -315,7 +315,9 @@ def add_format(self, gpodder_config, opts, fallback=None):
     def fetch_video(self, url, tempname, reporthook):
         opts = {
             'paths': {'home': os.path.dirname(tempname)},
-            'outtmpl': os.path.basename(tempname) + '.%(ext)s',  # Postprocessing breaks without ext
+            # Postprocessing in yt-dlp breaks without ext
+            'outtmpl': os.path.basename(tempname) + (
+                '.%(ext)s' if program_name == 'yt-dlp' else ''),
```
auouymous, line 320: `outtmpl` should be `(os.path.basename(tempname) if program_name == 'yt-dlp' else tempname) + '.%(ext)s'`. The `paths` is fine to always send because youtube-dl ignores it, and that means youtube-dl needs the full tempname path, not basename.

I don't know if youtube-dl post-processing breaks but it does support %(ext)s. I can't test it right now, do you see any reason to not append the extension for youtube-dl?

auouymous (changes requested):

share/gpodder/extensions/youtube-dl.py
tpikonen: I tested this with recent versions of both yt-dlp and youtube-dl. Both download fine with `.%(ext)s` added, but with youtube-dl, the preview functionality in gPodder (playing the temp-file being downloaded) is broken when the ext is added.

The preview works in youtube-dl only with single-file formats. With yt-dlp, the ext needs to be added, otherwise merging multi-file formats and other post-processing will fail. Is there an easy way to disable the preview for files downloaded with custom dowloader (i.e. the youtube-dl extension)?

auouymous: Preview is broken with youtube-dl and yt-dlp for me, because it passes "{name}.{ext}.partial" to the video player for both. https://github.com/gpodder/gpodder/blob/master/src/gpodder/model.py#L627 produces this filename and maybe custom downloaders could get a new method to resolve partial filenames. It could append the extension if known or return None if the partial file shouldn't be previewed. The preview menu could call get_playback_url and disable Preview if None.

If retrieve_resume could get the extension before downloading, it could store it in the downloader and we could have DownloadTask.run() make its `downloader` variable available in the task so `episode.download_task.custom_downloader.extension` is accessible by this new partial filename resolver.

If possible, retrieve_resume should detect DASH formats and set a flag so resolving the partial filename always returns None. Or we simply disable preview for all youtube-dl/yt-dlp episodes by always returning None.

tpikonen: The preview for the youtube-dl extension downloads is now implemented more or less as @auouymous describes above. Preview should now work for normal downloads and yt-dlp and youtube-dl downloads which do not use a fragment downloader (DASH etc.). Needs at least some testing and squashing before merging.

auouymous (changes requested):

src/gpodder/download.py
```diff
@@ -54,9 +55,24 @@
...
     """ abstract class for custom downloads. DownloadTask call retrieve_resume() on it """
 
+    @property
+    @abstractmethod
+    def partial_filename(self):
+        """
+        Full path to the temporary file actually being downloaded (downloaders
+        may not support setting a tempname).
+        """
+        ...
```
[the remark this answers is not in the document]
tpikonen: Not at least here. It's used in the `abc` module [docs](https://docs.python.org/3/library/abc.html#abc.abstractmethod), so I suppose this is standard notation.

auouymous (approved):

auouymous:
I'll merge it in a week if nothing breaks during testing.

tpikonen:
> I'll merge it in a week if nothing breaks during testing.

Great, I'll squash this before that.

PR #1395, opened 2022-09-12
Migrate from bundled minidb to minidb 2 (Fixes #1391)

### LINKED ISSUE #1391: Remove bundled "minidb", port to new minidb 2 and depend on it
(coming from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=891327)

The "minidb" package has since its inception more than 10 years ago turned into a project of its own:
https://github.com/thp/minidb

gPodder currently still ships an old copy of this here:
https://github.com/gpodder/gpodder/blob/master/src/gpodder/minidb.py

Version 2.0.0 of minidb changed a few things, so the code using it would need to be updated.

IMHO:

- [ ] Remove `src/gpodder/minidb.py`
- [ ] Install "minidb" like podcastparser/mygpoclient is installed right now (`tools/localdepends.py`, etc..)
- [ ] Update `src/gpodder/my.py` for the new API

Right now, minidb seems to be only used by the mygpo integration: `src/gpodder/my.py`

The model classes in `my.py` need to be rewritten to subclass from `minidb.Model` instead of using `__slots__`.

gPodder 4 Core (QML UI, ...) uses minidb 2 as package already:
https://github.com/gpodder/gpodder-core

thp:
WIP, please do not merge yet.


thp:
This is now ready for review :partying_face:

auouymous:
I don't see any problems in the patch.

Does this require a new gpodder.net database file?

```
1663493812.066773 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/home/test/git-gpodder/src/gpodder/gtkui/app.py", line 190, in do_activate
    self.window = gPodder(self, self.bus_name, core.Core(UIConfig, model_class=Model), self.options)
  File "/home/test/git-gpodder/src/gpodder/gtkui/main.py", line 92, in __init__
    BuilderWidget.__init__(self, None,
  File "/home/test/git-gpodder/src/gpodder/gtkui/interface/common.py", line 52, in __init__
    GtkBuilderWidget.__init__(self, gpodder.ui_folders, gpodder.textdomain, parent, **kwargs)
  File "/home/test/git-gpodder/src/gpodder/gtkui/base.py", line 71, in __init__
    self.new()
  File "/home/test/git-gpodder/src/gpodder/gtkui/main.py", line 205, in new
    self.mygpo_client = my.MygPoClient(self.config)
  File "/home/test/git-gpodder/src/gpodder/my.py", line 182, in __init__
    self._store.register(modelclass)
  File "/usr/lib/python3.10/site-packages/minidb.py", line 228, in register
    self._ensure_schema(table, slots)
  File "/usr/lib/python3.10/site-packages/minidb.py", line 203, in _ensure_schema
    raise TypeError('Column {} is {}, but expected {}'.format(name, next(dbtype for n, dbtype in
TypeError: Column action_type is TEXT, but expected INTEGER
```

Gentoo only has minidb 2.0.6 and a quick search suggests arch also only has 2.0.6. Have you checked other distros for 2.0.7? And yes, the above crash is with the required 2.0.7.

auouymous:
And should the new test show up in the "lint and test" workflow?

Off topic, but the two msgfmt fatal errors not failing the workflow doesn't seem right.

thp:
> Does this require a new gpodder.net database file?
> 
> ```
>   File "/usr/lib/python3.10/site-packages/minidb.py", line 203, in _ensure_schema
>     raise TypeError('Column {} is {}, but expected {}'.format(name, next(dbtype for n, dbtype in
> TypeError: Column action_type is TEXT, but expected INTEGER
> ```

Good point. I need to look into that, this needs fixing before merge.

> Gentoo only has minidb 2.0.6 and a quick search suggests arch also only has 2.0.6. Have you checked other distros for 2.0.7? And yes, the above crash is with the required 2.0.7.

2.0.7 was only released a few days ago (motivated by this very PR) -- we can probably postpone merging this until 2.0.7 has trickled into more distros. The above issue will need fixing separately as you said.

> And should the new test show up in the "lint and test" workflow?

Yes, it [does show up](https://github.com/gpodder/gpodder/actions/runs/3070110698/jobs/4959501626#step:6:27) - it's just a single all-in-one testcase for now:

```
tests/test_my.py .                                                       [100%]
```

> Off topic, but the two msgfmt fatal errors not failing the workflow doesn't seem right.

Filed as #1399 so we don't forget.

thp:
Setting to "draft", as the migration error (if the old database file has been created with minidb 1) still has to be handled:

    TypeError: Column action_type is TEXT, but expected INTEGER

thp:
I worked around the migration issue by just renaming the local database file from `gpodder.net` -> `mygposync.db`. This isn't ideal, but we might anyway have to rework this at some point.

auouymous:
So all actions are lost if users don't sync before upgrading?

thp:
> So all actions are lost if users don't sync before upgrading?

Yes, any actions not synced. I can add migration code to it.

auouymous:
When ready, this will need to mention minidb on https://gpodder.github.io/docs/run-from-git.html, and add it to `.github/workflows/linttest.yml`.

PR #1398, opened 2022-09-16
Fix the format_time() function to display times greater than 24 hours.

auouymous:
Using datetime converts the seconds value into date and time components. A value of 86400 generates a 00:00 time on the 2nd of January 1970.

The new code properly increments the hour field to 24 and beyond instead of wrapping around to zero. The left-most zero for hour or minute field is now dropped to make episode times easier to read.

### DIFF
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1386,7 +1386,7 @@ def bluetooth_send_file(filename):
         return False
 
 
-def format_time(value):
+def format_time(seconds):
     """Format a seconds value to a string
 
     >>> format_time(0)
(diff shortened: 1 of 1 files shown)


src/gpodder/util.py
```diff
@@ -1386,23 +1387,33 @@ def bluetooth_send_file(filename):
...
 
     >>> format_time(0)
-    '00:00'
+    '0:00'
     >>> format_time(20)
-    '00:20'
+    '0:20'
     >>> format_time(3600)
-    '01:00:00'
+    '1:00:00'
```
tpikonen, line 1398: Time formatted like this looks horrible. Please leave the output to the [ISO standard](https://en.wikipedia.org/wiki/ISO_8601#Times) time format, as it was.

tpikonen:
Where do you need time strings where the hour is > 23?

src/gpodder/util.py
auouymous: The format_time() function is only used for duration, and not actual timestamps. Any episode with a duration longer than a day will wrap and you might download a 25 hour livestream because it only shows as 01:00:00 or 1:00:00 in length, instead of 25:00:00. I can add the zero padding if you still want it.

tpikonen: Ok, showing durations with hours > 23 should be fine, since there are probably not that many cases where this number is above 100. I think the `%02d` format is essential for all H, M, S components though.

auouymous: Fixed.

src/gpodder/util.py
```diff
@@ -1397,12 +1398,22 @@ def format_time(value):
...
-    dt = datetime.datetime.utcfromtimestamp(value)
-    if dt.hour == 0:
-        return dt.strftime('%M:%S')
+    >>> format_time(86401)
+    '24:00:01'
+    """
+    hours = 0
+    minutes = 0
+    if seconds >= 3600:
+        hours = math.floor(seconds / 3600)
```
tpikonen, line 1407: You can leave out the `math` import if you use the `//` integer division operator here and below.

auouymous: Fixed, thank you!

tpikonen:
LGTM.

PR #1400, opened 2022-09-20
Write to log when sync directory is not writable.

auouymous:
The error message told user to check logs but nothing had been written.


elelay:
merged, thanks!

Issue #1371, opened 2022-08-14
Sync to Device should display an error message


tameware:
Under MacOS 12.5, gPodder requires Full Disk Access in order to sync to a mounted USB device. In version 3.11.0 (and likely earlier versions) "Sync to Device" fails silently if this permission is missing. The user should be directed to Full Disk Access settings in System Preferences -> Security & Privacy.

The command-line version shows:

```
(gpodder:42025): GLib-GIO-CRITICAL **: 18:49:44.332: g_dbus_proxy_get_name_owner: assertion 'G_IS_DBUS_PROXY (proxy)' failed
1660495784.375275 [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/Applications/gPodder 3.11.0.app/Contents/Resources/lib/python3.9/site-packages/gpodder/gtkui/main.py", line 3883, in on_sync_to_device_activate
    self.sync_ui.on_synchronize_episodes(self.channels, episodes, force_played)
  File "/Applications/gPodder 3.11.0.app/Contents/Resources/lib/python3.9/site-packages/gpodder/syncui.py", line 108, in on_synchronize_episodes
    if not device.open():
  File "/Applications/gPodder 3.11.0.app/Contents/Resources/lib/python3.9/site-packages/gpodder/sync.py", line 468, in open
    self.tracks_list = self.get_all_tracks()
  File "/Applications/gPodder 3.11.0.app/Contents/Resources/lib/python3.9/site-packages/gpodder/sync.py", line 572, in get_all_tracks
    for child_info in path_file.enumerate_children(attributes, Gio.FileQueryInfoFlags.NONE, None):
gi.repository.GLib.GError: g-io-error-quark: Error opening directory '/Volumes/FINISDUO/.Spotlight-V100': Operation not permitted (14)
```

auouymous:
This could just as easily happen in linux as well, because firefox on linux silently uploads empty files when it doesn't have permission to read them. Any message gpodder shows the user for this must be generic across platforms.

auouymous:
```
        # open is ok if the target is a directory, and it can be written to
        # for smb, query_info doesn't return FILE_ATTRIBUTE_ACCESS_CAN_WRITE,
        # -- if that's the case, just assume that it's writable
```

[...]

PR #1401, opened 2022-09-20
Catch unknown sync exceptions when opening device and show error.

auouymous:
Some devices lack writable info and gpodder assumes they are writable. This can lead to an exception being thrown when not writable.

See #1371.


elelay:
Merged, thanks

PR #1402, opened 2022-09-20
Add menu item to open logs folder.

auouymous:
This isn't usable when gpodder's UI is crashed or hung, but it might make it easier for users to find logs in other cases.


elelay:
Merged, thanks :-)
Needs to be localized + tested on Windows before release

PR #1403, opened 2022-09-21
Quoting file URIs moved by drag and drop

bahorn:
When dragging and dropping a podcast into another program, e.g VLC, some filenames would cause issues (i.e ones with `#`).

Seems this was handled with all the other cases involving file URIs, so I patched it the exact same way the other cases were handled.


thp (approved):

auouymous:
Thanks for the fix.

PR #1404, opened 2022-09-23
Use symbolic toolbar icons.

auouymous:
The preferences-desktop icon is not available for 64x64 and larger icon sizes, and gtk displays a missing icon instead of falling back to smaller icons. Switching to preferences-other-symbolic would fix the issue, but the other toolbar icons should probably be changed to symbolic to match.

### DIFF
--- a/share/gpodder/ui/gtk/gpodder.ui
+++ b/share/gpodder/ui/gtk/gpodder.ui
@@ -36,7 +36,7 @@
                 <property name="can-focus">False</property>
                 <property name="is-important">True</property>
                 <property name="label" translatable="yes">Play</property>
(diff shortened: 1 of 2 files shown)


elelay:
symbolic

elelay:
LGTM.
Before the patch I already symbolic icons + the non-symbolic preferences...

auouymous:
So you have some setting or configuration that forces symbolic icons if available. I don't use the toolbar but I do wish my double sized icons were as small as yours.

I forgot to mention that I changed the download icon from `go-down` to `document-save`, which is also used by the episode context menu.

elelay:
thanks,
yes I noticed the document-save icon and think it's fine like that.

PR #1421, opened 2022-10-11
Updated French translation

rene-coty:
I updated the fr.po file

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -10,8 +10,8 @@ msgstr ""
 "Project-Id-Version: gPodder\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2022-09-25 20:23-0600\n"
-"PO-Revision-Date: 2021-06-10 10:33+0200\n"
-"Last-Translator: Eric Le Lay <elelay@macports.org>\n"
+"PO-Revision-Date: 2022-10-11 17:59+0200\n"
+"Last-Translator: Irénée Thirion <irenee.thirion@e.email>\n"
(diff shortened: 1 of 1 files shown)


po/fr.po
```diff
@@ -2389,53 +2338,51 @@ msgstr "Nettoyage"
...
 #: share/gpodder/ui/gtk/gpodderpreferences.ui.h:32
 msgid "Only sync unplayed episodes"
 msgstr "Ne synchroniser que les épisodes non lus"
 
 #: share/gpodder/ui/gtk/gpodderpreferences.ui.h:33
-#, fuzzy
-#| msgid "Remove episodes deleted on device from gPodder"
 msgid "Remove episodes deleted in gPodder from device"
-msgstr "Retirer de gPodder les épisodes supprimés sur le périphérique"
+msgstr "Retirer les épisodes supprimés de gPodder du périphérique"
```
elelay, line 2369: Retirer du périphérique les épisodes supprimés dans gPodder

elelay (changes requested):
Looks good to me except one sentence that I would rephrase (see comment).

Also I'm not sure why you removed the space following colon, which I would consider necessary in French. Please explain :-)

rene-coty:
> Thank you for the translation. Looks good to me except one sentence that I would rephrase (see comment).

Okay I'll change it :+1:

rene-coty:
Normally there is no need of a space following colon in a French translation (if there is no text after it) ; it is only required to have an unbreakable space before a colon.

elelay:
Merged, thanks

PR #1429, opened 2022-10-27
Queue all episodes before download to allow resuming in gpo.

### LINKED ISSUE #1425: gpodder/gpo when checking for updates new updates are not added to download cue
hi,
when checking for updates for podcasts in gpo any new epesodes are not added to the download cue.
in gpodder-gtk i've set it so that when new epesodes are found, gpodder will auto add to download cue ready to be downloaded.
this does not happen for the gpo interface.
could this be looked at?
Majid

auouymous:
Fixes #1425.

### DIFF
--- a/bin/gpo
+++ b/bin/gpo
@@ -635,7 +635,10 @@ class gPodderCli(object):
 
     def _download_episode(self, episode):
         with self._action('Downloading %s' % episode.title):
-            task = download.DownloadTask(episode, self._config)
+            if episode.download_task is None:
+                task = download.DownloadTask(episode, self._config)
+            else:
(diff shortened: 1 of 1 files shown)


mhussaincov93:
hi,
so just updated gpodder from git,
and tryed your fix.
launching gpo and checking for updates,
I did cansel after a few channels updated,
I then typed rezume
no epesodes were found to rezume.
typeing partial also found no files.
am i doing something wrong?

auouymous:
This does nothing for an abort during channel update, gpodder-gtk also does not resume an aborted channel update. This PR allows `gpo download` to be resumed with `gpo resume`.

PR #1374, opened 2022-08-18
channel editor: Add buttons next to cover, feed URL, etc.

tpikonen:
Remove the cover image context menu and add buttons next to various items in the channel editor:

 * Cover image: Pop-up a menu to change or refresh the cover image
 * Feed URL: Copy the URL to clipboard
 * Download folder: Open the download folder

This makes the functionality easier to discover and more touch friendly.

Changing the cover image to a local file was not working, the second commit fixes that.

PR #1431, opened 2022-11-03
Add flake8 config and fix some code style issues

tpikonen:
I think it would be nice to have a stricter machine validated code style for gPodder (i.e. I started using autolinting on my editor and everything is red). This PR adds a flake8 config section to setup.cfg with ignores for some errors and warnings, and fixes (mostly automatically made) for some others.

There are still 157 flake8 errors in gPodder after the fixes and ignores, but it's a start. This is also probably really boring to review, but here goes.

### DIFF
--- a/bin/gpodder
+++ b/bin/gpodder
@@ -139,9 +139,9 @@ def main():
     from gpodder import log
     log.setup(options.verbose, options.quiet)
 
-    if (not (gpodder.ui.win32 or gpodder.ui.osx) and
-            os.environ.get('DISPLAY', '') == '' and
-            os.environ.get('WAYLAND_DISPLAY', '') == ''):
+    if (not (gpodder.ui.win32 or gpodder.ui.osx)
+            and os.environ.get('DISPLAY', '') == ''
+            and os.environ.get('WAYLAND_DISPLAY', '') == ''):
         logger.error('Cannot start gPodder: $DISPLAY or $WAYLAND_DISPLAY is not set.')
         sys.exit(1)
 
--- a/bin/gpodder
+++ b/bin/gpodder
@@ -39,7 +39,6 @@ logger = logging.getLogger(__name__)
 
 try:
     import dbus
-    from dbus.mainloop.glib import DBusGMainLoop
     have_dbus = True
 except ImportError:
     print("""
--- a/src/gpodder/gtkui/desktopfile.py
+++ b/src/gpodder/gtkui/desktopfile.py
@@ -169,7 +169,6 @@ class UserAppsReader(object):
 
         self.__has_read = True
         if gpodder.ui.win32:
-            import winreg
             for caption, types, hkey in WIN32_APP_REG_KEYS:
                 try:
                     cmdline = win32_read_registry_key(hkey)
--- a/tools/fake-dbus-module/dbus/__init__.py
+++ b/tools/fake-dbus-module/dbus/__init__.py
@@ -1,10 +1,9 @@
-import dbus.exceptions
+import dbus.exceptions  # noqa: F401
 
 
 class SessionBus(object):
     def __init__(self, *args, **kwargs):
         self.fake = True
-        pass
 
     def add_signal_receiver(self, *args, **kwargs):
         pass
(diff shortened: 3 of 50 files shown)


auouymous:
LGTM. I'm going to test it for a week to make sure it doesn't cause any major issues.

It'll be nice to have flake8 run with `make lint` after we fix the remaining errors.

bin/gpodder
```diff
@@ -39,7 +39,6 @@ logger = logging.getLogger(__name__)
 
 try:
     import dbus
-    from dbus.mainloop.glib import DBusGMainLoop
```
elelay, line 42: that one may need to stay for full dbus detection, don't you think?

src/gpodder/gtkui/desktopfile.py
```diff
@@ -169,7 +169,6 @@ def read(self):
 
         self.__has_read = True
         if gpodder.ui.win32:
-            import winreg
```
elelay, line 172: if winreg is not there it would raise the exception. Otherwise it would be logged in the try-catch line 174

elelay: (ok winreg is in the stdlib so it should be there!)

tools/fake-dbus-module/dbus/__init__.py
```diff
@@ -1,10 +1,8 @@
-import dbus.exceptions
```
elelay, line 1: it should stay so users import dbus.exceptions from fake-dbus-module as in real dbus module, right?

elelay (changes requested):
not sure about some removed imports

bin/gpodder
auouymous: I am using the fake dbus module instead of python-dbus and `have_dbus` is True with or without the DBusGMainLoop import.

tools/fake-dbus-module/dbus/__init__.py
auouymous: Wouldn't that import only apply inside that file, and it doesn't use any exceptions?

bin/gpodder
tpikonen: I'm somewhat out of my depth here, not having really used dbus-python, but is there a situation where `dbus` could be imported but `dbus.mainloop.glib` not? If this line is really needed, then it can be marked with `noqa: F401` comment.

tools/fake-dbus-module/dbus/__init__.py
tpikonen: This one is probably needed, since `dbus.exceptions.DBusException` is used in app.py without any `have_dbus`-style guards.

bin/gpodder
elelay: ok, I added DBusGMainLoop import in e4a16f152 but as far as I can tell [from the code](https://github.com/freedesktop/dbus-python/blob/dbus-python-1.2.18/dbus_glib_bindings/module.c#L180) it should be instantiated to do anything useful.

elelay (approved):
good to go :+1: 
thanks for this work!

tpikonen:
Squashed the fixup.

auouymous:
The following hunk was modified in #1382. Would it be easier to remove the hunk from this PR or rebase it in #1382?

```patch
@@ -621,8 +621,8 @@ def get_playback_url(self, config=None, allow_partial=False):
         """
         url = self.local_filename(create=False)
 
-        if (allow_partial and url is not None and
-                os.path.exists(url + '.partial')):
+        if (allow_partial and url is not None
+                and os.path.exists(url + '.partial')):
             return url + '.partial'
 
         if url is None or not os.path.exists(url):
```

auouymous:
And the following hunk was modified in #1374.

```patch
@@ -17,7 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-from gi.repository import Gdk, GdkPixbuf, Gtk
+from gi.repository import Gdk, Gtk
 
 import gpodder
 from gpodder import util
```

tpikonen:
> The following hunk was modified in https://github.com/gpodder/gpodder/pull/1382. Would it be easier to remove the hunk from this PR or rebase it in https://github.com/gpodder/gpodder/pull/1382?

I can rebase this to master and fix these after #1382 is merged.

tpikonen:
Rebased to master and fixed the conflicts.

auouymous:
Thank you for adding this.

PR #1440, opened 2022-11-22
Fix support for old versions of yt-dlp.

### LINKED ISSUE #1439: Current git does not work with yt-dlp 2022.04.08
https://github.com/gpodder/gpodder/commit/642bf7e738036e6e7a68aa8b8c752a3c9d619f52 in #1382 causes the following error when downloading episodes with yt-dlp 2022.04.08:

```
  File "/home/test/git-gpodder-tmp/src/gpodder/download.py", line 900, in run
    headers, real_url = downloader.retrieve_resume(self.tempname, self.status_updated)
  File "/home/test/git-gpodder-tmp/share/gpodder/extensions/youtube-dl.py", line 109, in retrieve_resume
    opts['paths']['home'], opts['outtmpl']['default']) % info
TypeError: string indices must be integers
```

I haven't yet determined which version of yt-dlp is required to make #1382 work. This issue either needs to be fixed for older versions of yt-dlp, or gpodder needs to split `want_ytdl_version` to support separate versions for youtube-dl and yt-dlp, which I already had plans to do. @tpikonen Do you know an easy fix for this or should we require a newer yt-dlp?

auouymous:
PR #1382 broke support for yt-dlp versions prior to 2022.6.22.1.

Fixes #1439.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -106,8 +106,8 @@ class YoutubeCustomDownload(download.CustomDownload):
         outtmpl = tempname.replace('%', '%%')
         info, opts = self._ytdl.fetch_info(self._url, outtmpl, self._my_hook)
         if program_name == 'yt-dlp':
-            self.partial_filename = os.path.join(
-                opts['paths']['home'], opts['outtmpl']['default']) % info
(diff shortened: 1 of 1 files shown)


tpikonen (approved):
The code is fine and works ok on post 2022.06.22.1 yt-dlp.

tpikonen:
Thanks for fixing this. Unfortunately it seems that yt-dlp does not provide a stable Python API.

PR #1441, opened 2022-11-28
delete tags before modifying

Skitals:
When "modify_tags" is enabled some episodes are getting carryover unrelated metadata prepended to the comments tag.

Screenshot 2022-11-28 at 12 14 01 AM

This displays different in different players. In iTunes the bug isn't visible and it displays "People trying to coax each other [...]". However in other clients, such as Emby, it displays "preroll_1;postroll_1" and the show notes are not displayed at all.

This patch clears existing tags from the mp3 before writing the new ones, which fixes this bug.

### DIFF
--- a/share/gpodder/extensions/tagging.py
+++ b/share/gpodder/extensions/tagging.py
@@ -94,7 +94,7 @@ class AudioFile(object):
             audio.add_tags()
 
         if modify_tags:
-            if remove_before_modify and audio.tags is not None:
+            if remove_before_modify:
                 audio.delete()
 
             if self.album is not None:
--- a/share/gpodder/extensions/tagging.py
+++ b/share/gpodder/extensions/tagging.py
@@ -95,7 +95,7 @@ class AudioFile(object):
         if modify_tags:
             if audio.tags is not None:
                 audio.delete()
-            
+
             if self.album is not None:
                 audio.tags['album'] = self.album
 
--- a/share/gpodder/extensions/tagging.py
+++ b/share/gpodder/extensions/tagging.py
@@ -93,6 +93,9 @@ class AudioFile(object):
             audio.add_tags()
 
         if modify_tags:
+            if audio.tags is not None:
+                audio.delete()
+            
             if self.album is not None:
                 audio.tags['album'] = self.album
 
--- a/share/gpodder/extensions/tagging.py
+++ b/share/gpodder/extensions/tagging.py
@@ -62,7 +62,8 @@ DefaultConfig = {
     'auto_embed_coverart': False,
     'set_artist_to_album': False,
     'set_version': 4,
-    'modify_tags': True
+    'modify_tags': True,
+    'remove_before_modify': False
 }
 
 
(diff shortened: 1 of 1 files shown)


auouymous:
What is the URL for this feed?

I don't use the tagging extension, but wouldn't removing tags delete those not set by gpodder, and also those which are empty in gpodder?

Skitals:
> What is the URL for this feed?
> 
> I don't use the tagging extension, but wouldn't removing tags delete those not set by gpodder, and also those which are empty in gpodder?

The feed in the example is The Daily from The New York Times: https://feeds.simplecast.com/54nAGcIl

It also happens in some episodes of This American Life: https://www.thisamericanlife.org/podcast/rss.xml

Yes, with this change when "modify_tags" is enabled all tags are stripped except ones set by gpodder.

auouymous:
Mutagen 1.45.1 and 1.46.0 delete all existing tags and don't set gpodder tags with or without this patch, even though the modify_tags code path is executed and the tags are all set to non-empty strings.

Skitals:
> Mutagen 1.45.1 and 1.46.0 delete all existing tags and don't set gpodder tags with or without this patch, even though the modify_tags code path is executed and the tags are all set to non-empty strings.

I am on mutagen 1.41.1 and that is not the case on gpodder 3.10.21 or 3.11.0. It works as expected with the exception of the excess metadata bug I am trying to fix. These are my extension settings:

```
  "extensions": {
    "enabled": [
      "update_feeds_on_startup",
      "rename_download",
      "tagging"
    ],
    "rename_download": {
      "add_podcast_title": false,
      "add_sortdate": true
    },
    "tagging": {
      "always_remove_tags": false,
      "auto_embed_coverart": true,
      "genre_tag": "Podcast",
      "modify_tags": true,
      "set_artist_to_album": true,
      "set_version": 4,
      "strip_album_from_title": true
    }
  },
```

Skitals:
new commit makes this an option that is defaulted off. If "remove_before_modify" is enabled existing tags are deleted before writing basic tags.

auouymous (changes requested):

auouymous (approved):

auouymous:
LGTM, I'll squash and merge in a week if no one has a problem with the change.

Are you able to upgrade mutagen to 1.45.1 or newer? I'm curious to know if it is only broken for me on Gentoo, or if it is broken for everyone with a newer version, and no one uses the extension or realizes it is broken.

auouymous:
Ignore that, I thought you said 1.41.1 which wouldn't run for me on Python 3.10.

Skitals:
> Ignore that, I thought you said 1.41.1 which wouldn't run for me on Python 3.10.

FWIW I'm running xthursdayx/gpodder-docker. You can see what is installed [in the dockerfile](https://github.com/xthursdayx/gpodder-docker/blob/main/light/Dockerfile). I reinstalled the build-deps and did a git pull on gpodder to update to latest. Tagging works great with this one exception, which could have gone unnoticed for a long time since some players don't display the extra comment data.

auouymous:
Thank you for the improvement.

PR #1461, opened 2023-01-26
Allow channel to be selected from both episode menus.

### LINKED ISSUE #1170: Go to channel from episode
It would be nice if the channel for a given episode could be selected from the episode list. When there are a lot of channels that scroll, or the channels are in another language, it can be difficult to find the channel to filter episode list. And a channel search takes time.

The episode context menu could contain an entry that selects the channel.

Requested by naota on IRC.

auouymous:
Closes #1170.


elelay:
Merged, thanks

PR #1463, opened 2023-01-29
additional config to set playlist file extension

steviehs:
e.g. Sony NW-MS* players need: m3u8 instead of m3u.
Right now, no entry in UI configuration, but should be fine anyways.


auouymous:
Thanks for the improvement.

PR #1464, opened 2023-01-29
Accept coverart as SVG

tpikonen:
Some feeds have an SVG-file as their cover (the `atom:icon` element). Currently gPodder only accepts raster file formats (.png .jpeg, etc.) as cover files. This adds SVG to the allowed formats.

This causes an implicit dependency on librsvg via `GdkPixbuf.Pifbuf.new_from_file()`. This is probably not a problem on Linux, but could be on other platforms.


auouymous:
Thank you. The mac and windows builds already include librsvg.

PR #1465, opened 2023-01-29
correct absolute path config settings for playlist

steviehs:
self._config.device_sync.playlists.use_absolute_path when checking if absolute path in playlists was desired
This lead to always have the absolute path (which does lead using non existing config variables not to an error?)


auouymous:
Thank you for the fix.

PR #1480, opened 2023-02-19
appdata: Update to version 3.11.1, fixes

tpikonen:
Add the latest release to the appdata.xml, fix the XML syntax and add the release dates of 3.9.x versions.

The flathub guidelines now recommend that the appdata file is maintained with the upstream sources. It would be nice to have the release notes included in the appdata before the release, so it does not need to be patched for the flatpak.


auouymous:
Thank you for the fix.

I'll add this to the release process.

PR #1484, opened 2023-02-21
Drop old python versions

auouymous:
Podcastparser has dropped support for EOL Python 3.5 and 3.6 versions. Does anyone object to doing the same in gPodder?

I will update website after this is merged.


elelay:
go for it :+1:

PR #1488, opened 2023-03-06
Catch errors for directories inside sync device.

auouymous:
A directory inside the sync device might not be accessible and aborts the entire sync. The inaccessible directory might belong to the device and not meant to be read by Gpodder.

Mac places an inaccessible `.Trashes` folder on storage devices. The GUI will not tell the user there was an error in the logs, which is great for this situation, but this might mask an actual problem. It might be better to skip hidden directories in get_all_tracks().


auouymous:
@elelay Do you know why PRs don't produce a Mac build?

PR #1492, opened 2023-03-14
Catch errors for directories inside sync device.

auouymous:
A directory inside the sync device might not be accessible and aborts the entire sync. The inaccessible directory might belong to the device and not meant to be read by Gpodder.

See #1488.

### DIFF
--- a/src/gpodder/sync.py
+++ b/src/gpodder/sync.py
@@ -578,10 +578,13 @@ class MP3PlayerDevice(Device):
             if self._config.one_folder_per_podcast:
                 if path_info.get_file_type() == Gio.FileType.DIRECTORY:
                     path_file = root_path.get_child(path_info.get_name())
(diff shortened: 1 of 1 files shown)


auouymous:
@elelay Do you see any problems merging this? The errors would happen on each sync and a dialog would be quite annoying. Should it send a notification or would that also be annoying? This example would send two notifications per sync. No one would care about these as they aren't real errors, I'm just not sure if there would be any issues this would hide.

```
ERROR: get all tracks for
  file:///Volumes/FINISDUO/.Spotlight-V100 failed: Error opening directory
    '/Volumes/FINISDUO/.Spotlight-V100': Operation not permitted
ERROR: get all tracks for
  file:///Volumes/FINISDUO/.Trashes failed: Error opening directory
    '/Volumes/FINISDUO/.Trashes': Operation not permitted
```

elelay:
sounds good to me :+1:

Issue #1050, opened 2021-05-24
Video to Audio extension


auouymous:
Requested by @mhussaincov93

The extension would run `ffmpeg -i "$filename" "${filename%.*}.mp3"` and then set episode.download_filename to ${filename%.*}.mp3 and set episode.mime_type to audio/mp3. It could have an option to use ogg instead of mp3.

tpikonen:
It looks to me that the audio_converter.py extension could be easily modified to do this. Maybe just changing the accepted MIME types would be enough?

tpikonen:
Well, I had to try. Adding 'video/mp4' to MIME_TYPES makes the audio converter extension accept a downloaded youtube video, and conversion works fine from the context menu. Here's a patch for reference:

```
diff --git a/share/gpodder/extensions/audio_converter.py b/share/gpodder/extensions/audio_converter.py
index 6e48b3b9..1939b68c 100644
--- a/share/gpodder/extensions/audio_converter.py
+++ b/share/gpodder/extensions/audio_converter.py
@@ -33,7 +33,7 @@ DefaultConfig = {
 
 
 class gPodderExtension:
-    MIME_TYPES = ('audio/x-m4a', 'audio/mp4', 'audio/mp4a-latm', 'audio/mpeg', 'audio/ogg', 'audio/opus')
+    MIME_TYPES = ('video/mp4', 'audio/x-m4a', 'audio/mp4', 'audio/mp4a-latm', 'audio/mpeg', 'audio/ogg', 'audio/opus')
     EXT = ('.m4a', '.ogg', '.opus', '.mp3')
     CMD = {'avconv': {'.mp3': ['-n', '-i', '%(old_file)s', '-q:a', '2', '-id3v2_version', '3', '-write_id3v1', '1', '%(new_file)s'],
                       '.ogg': ['-n', '-i', '%(old_file)s', '-q:a', '2', '%(new_file)s'],
```

auouymous:
So the extension just needs more video mime types, and also needs to force a refresh of the episode icon, which did change when played.

PR #1517, opened 2023-06-19
Update audio_converter.py

wilyarti:
Add MIME type of MP4 to allow MP4 to MP3 conversion as per issue  #1050

### DIFF
--- a/share/gpodder/extensions/audio_converter.py
+++ b/share/gpodder/extensions/audio_converter.py
@@ -33,7 +33,7 @@ DefaultConfig = {
 
 
 class gPodderExtension:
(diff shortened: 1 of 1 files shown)


auouymous:
There are a couple problems with this approach.

It will convert all mp4 videos to an audio file. I imagine most users of the extension are using it to convert to mp3 because their sync device doesn't support ogg, or they simply don't want mp3 files and convert them to ogg. Anyone converting for device support might not be downloading videos, but those who don't want mp3 might have videos and might not want them converted to audio. So there should be an option to enable the video mime type.

There is more than one video type. Perhaps when it doesn't match an audio type and the video option is enabled, it uses a separate check for `video/` in the mime type so it will support all types.

wilyarti:
For my personal use case I use YouTube channels as "podcasts" and download the videos. However my Surfans F20 doesn't play video so I'm having to convert everything to MP3.

But yeah I suppose you could put a few extra flags in there or some extra logic.

PR #1520, opened 2023-06-30
Adapt youtube-dl plugin to latest yt-dlp

tpikonen:
This fixes two problems found when using the youtube-dl plugin with the latest yt-dlp (2023.06.22).

Use `'color': 'no_color'` in yt-dlp params instead of `'no_color': True`. This prevents warning log output.

The internal member `_VALID_URL_RE` is now an iterable. Adapt the `is_supported_url` function and the regex cache code.

This likely breaks compatibility with original youtube-dl.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -296,9 +296,14 @@ class gPodderYoutubeDL(download.CustomDownloader):
         os.makedirs(cachedir, exist_ok=True)
         self._ydl_opts = {
             'cachedir': cachedir,
-            'no_color': True,  # prevent escape codes in desktop notifications on errors
             'noprogress': True,  # prevent progress bar from appearing in console
         }
+        # prevent escape codes in desktop notifications on errors
+        if program_name == 'yt-dlp':
+            self._ydl_opts['color'] = 'no_color'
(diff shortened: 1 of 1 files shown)


tpikonen:
Actually it was not too hard to make this compatible with the old version of yt-dlp and youtube-dl. Not tested though.

computerfreak105:
Thanks!

 

 

Von: Teemu Ikonen ***@***.***> 
Gesendet: Freitag, 30. Juni 2023 20:19
An: gpodder/gpodder ***@***.***>
Cc: Subscribed ***@***.***>
Betreff: Re: [gpodder/gpodder] Adapt youtube-dl plugin to latest yt-dlp (PR #1520)

 

Actually it was not too hard to make this compatible with the old version of yt-dlp and youtube-dl. Not tested though.

auouymous:
Thank you. fccea8a6d99c53cabc201253cddd84284b2ea139 bumps the required yt-dlp version so those with an older yt-dlp will be told to upgrade when they have trouble with the no_color change.

Youtube-dl can't be tested until they put out a release for the 2023.02.17 youtube.com breakage. It is under active developement, just no releases. At some point we can drop support, but for now we will try to remain compatible and assume all changes work.

PR #1522, opened 2023-07-14
Fix various typos

luzpaz:
Found via `codespell -q 3 -S *.po -L bloc,extracter,parms`


auouymous:
Thank you. That might be useful to run with the linter.

PR #1542, opened 2023-08-19
Fix #1538 GUI issues - Problem with accented letters in translation

### LINKED ISSUE #1538: GUI issues - Problem with accented letters in translation
@auouymous 

Please check screenshots.

image

image

image

In some part of GUI the accented letters like à ì are now showed right.

elelay:
Calling `libintl.bind_textdomain_codeset` from the dynamically loaded libintl library
is sufficient to fix accented letters issues.


auouymous:
Ah, it was only the one call that needed to be removed. Good to have you back, early.

PR #1543, opened 2023-08-20
Draft: Win32 launcher cleanup warnings

elelay:


tools/win_installer/misc/create-launcher.py
```diff
@@ -114,14 +114,48 @@ def get_launcher_code(entry_point):
...
 
-    Py_NoUserSiteDirectory = 1;
-    Py_IgnoreEnvironmentFlag = 1;
-    Py_DontWriteBytecodeFlag = 1;
-    Py_Initialize();
-    PySys_SetArgvEx(__argc, __wargv, 0);
-    result = PyRun_SimpleString("%s");
-    Py_Finalize();
-    return result;
+
```
auouymous, line 117: Is the double newline intentional?

tools/win_installer/misc/create-launcher.py
```diff
@@ -114,14 +114,48 @@ def get_launcher_code(entry_point):
...
+    PyStatus status;
+
+    PyConfig config;
+    PyConfig_InitPythonConfig(&config);
+
+    config.parse_argv = 1;
+    config.isolated = 1;
+    config.use_environment = 0;
+    config.write_bytecode = 0;
+    config.run_command =
```
auouymous, line 128: Is this C code? Does it need a value and semicolon?

auouymous (changes requested):

tools/win_installer/misc/create-launcher.py
elelay: my mistake. it's not valid c code. I found out I needed to used  `PyConfig_SetBytesString` instead of directly setting `config.run_command` but forgot to remove this line. I'm trying to get away with setting up a local build environment but this mean really stupid and untested code gets pushed. Sorry for the noise ;-)

auouymous (approved):

PR #1544, opened 2023-08-20
Add setting to toggle sync filesize comparisons.

### LINKED ISSUE #1416: question when syncing to a device and the opperation is interupted via the user, when rezumeing sed opperation will gpodder /gpo rezume from the epesode that was interupted?
sorry for the long title
but when gpodder /gpo rezumes sync opperations will it continue from the last epesode that it was syncing before the interuption?
or will it resync the set epesode that was interupted?
thanks,
Majid

auouymous:
Fixes #1416.

This defaults to comparing episode file sizes. I haven't yet figured out how to alert the user when it re-syncs all files, as the check happens inside the sync task code, and not in a loop that could easily track existing syncs.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -1019,6 +1019,22 @@
                                 <property name="position">10</property>
                               </packing>
                             </child>
+                            <child>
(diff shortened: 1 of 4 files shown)


auouymous:
Alerting the user should probably happen if **any** existing files are overwritten. The dialog could show the number of overwritten files and let the user know they were either partial uploads or their device modifies files and the setting should be disabled.

Issue #1448, opened 2022-12-23
GDPR prevents subscribing to youtube URLs with @ names


computerfreak105:
[2022-12-23.log](https://github.com/gpodder/gpodder/files/10297038/2022-12-23.log)
Hello!
If I'm not mistaken it's possible to subscribe to youtube channels in GPodder GUI by entering youtube channels in several ways. It doesn't necessarily have to be a channel ID like
https://www.youtube.com/channel/UCIzJqMXr6rT7PMgidXBFmtw

So, since youtube has introduced handles almost every channel has got new urls like
https://www.youtube.com/@paperworld2676
or
https://www.youtube.com/@user-og8ss9cj2t

Until now I used this online Tool here to find the channel ID of new youtube channels:

https://commentpicker.com/youtube-channel-id.php	

But I thought it would be nice if we could subscribe directly without having to use this tool.
So I tried this with GPO on CLI, because I wanted to see how this would look when I export the subscription as OPML-file.
GPO crashes when entering urls this way. (see attached log).

I tried to make a subscription with the GPodder GUI, but it's not accessible enough on windows for blind people and I think it didn't work - at least there was no subscription in the OPML file I exported afterwards.

So this is a report because of the crashing and a question, if it's possible at all to subscribe with the new urls youtube offers.
Thanks.

auouymous:
I think this worked when the feature was announced, are you using 3.11 or an older version? Did you subscribe with GPO from the command line or from within its command line interface? The former might not work on Windows because I think @ is a special character. If that is the case, try wrapping the URL in quotes or subscribe from within gpo.

I'll take a look in a few days after I fix my computer.

computerfreak105:
I'm using the portable version build artifact version 1.0.932.
And I start GPO from the windows explorer by opening the shortcut GPO from the portable version's folder.

Usually I prepare an OPML file and put all my channels I want to subscribe in there. Afterwards I'm importing the OPML file into GPO by writing import filename.opml.

"A line for a subscription looks like:

        <outline title="Cactus ASMR, Eng" text="No description available" xmlUrl="https://www.youtube.com/feeds/videos.xml?channel_id=UCsMNy2GLFpnonnJJf2cnAQg" type="rss"/>

The reason I do that is because I can decide the title of channels myself. This becomes handy for me because youtube channels may alter their title or they may contain foreigen language and unicode characters which I don't like having in a folder name.

Thanks for having a look.

auouymous:
[...]

PR #1549, opened 2023-08-25
youtube: Set consent cookie in get_channel_id_url()

### LINKED ISSUE #1290: GDPR prevents subscribing to youtube URLs with channel names
Trying to subscribe to youtube channel URLs of the type `https://www.youtube.com/c/<channel name>` results in

```
Exception: Could not retrieve youtube channel id.
```

Apparently there is an extra layer of cookie consent (at least in Europe) which prevents the actual channel page from being read. Subscribing to these channels with an URL with the channel id (`https://www.youtube.com/channel/<channel id>`) works fine.

tpikonen:
This surprisingly simple patch fixes the GDPR consent issues on parsing Youtube URLs. The cookie was taken from yt-dlp code.

Please check if this works on regions other than mine (EU).

Fixes #1290 and #1448.

### DIFF
--- a/src/gpodder/youtube.py
+++ b/src/gpodder/youtube.py
@@ -434,7 +434,7 @@ def get_channel_id_url(url, feed_data=None):
     if 'youtube.com' in url:
         try:
             if feed_data is None:
-                r = util.urlopen(url)
+                r = util.urlopen(url, cookies={'SOCS': 'CAI'})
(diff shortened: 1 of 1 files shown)


auouymous:
Subscribing and downloading works in the US.

auouymous:
Thanks for the fix.

PR #1551, opened 2023-09-02
Show remaining duration for each download in progress tab.

auouymous:
There is a brief moment at the start of download were task speed is zero, and it displays "00:00". Should it instead show "?" or "??:??", the later looking more like a an unknown duration? If so, it would need to be translatable since some languages don't use "?" as question mark, such as ";" in Greek.

### DIFF
--- a/src/gpodder/gtkui/download.py
+++ b/src/gpodder/gtkui/download.py
@@ -95,10 +95,12 @@ class DownloadStatusModel(Gtk.ListStore):
                     task.STATUS_MESSAGE[task.status],
                     task.error_message)
         elif task.status == task.DOWNLOADING:
-            status_message = '%s (%.0f%%, %s/s)' % (
-                    task.STATUS_MESSAGE[task.status],
-                    task.progress * 100,
-                    util.format_filesize(task.speed))
+            status_message = _('%(status)s (%(progress).0f%%, %(rate)s/s, %(remaining)s)') % {
+                    'status': task.STATUS_MESSAGE[task.status],
+                    'progress': task.progress * 100,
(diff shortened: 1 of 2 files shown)


tpikonen:
Maybe use a dash ('-') for unknown speed and duration, instead of 0?

elelay:
you don't have to find a single notation for all languages: use `_(...)` to let people translate...
Something like that, (I think it needs the `if else` so that i18n tools pick the string to localize)?

 ````
if task.speed > 0:
    status_text  = _('%(status)s (%(progress).0f%%, %(rate)b/s, %(remaining)s)')
else:
    status_text = _('%(status)s (%(progress).0f%%, %(rate)s/s, ?s)')
status_message = status_text % {
                    'status': task.STATUS_MESSAGE[task.status],
                    'progress': task.progress * 100,
                    'rate': util.format_filesize(task.speed))
                    'remaining': util.format_time(round((task.total_size * (1 - task.progress)) / task.speed)) if task.speed > 0 else 0)
}
````

auouymous:
> Maybe use a dash ('-') for unknown speed and duration, instead of 0?

Oh ya, I forgot about dash.


> status_text  = _('%(status)s (%(progress).0f%%, %(rate)b/s, %(remaining)s)')

Did you mean `%(rate)s/s` with an `s` instead of `b`?


> status_text = _('%(status)s (%(progress).0f%%, %(rate)s/s, ?s)')

Is the `?s` meant to just be `?`?

And do you think we should make these translatable or just use dash? It would allow the `/s` to be translated, except that format_filesize() doesn't use translatable strings for the size units. Should we also make those translatable?

PR #1553, opened 2023-09-07
Add plural form of translation string.

neodyne:
Noticed a translation string without a plural form while working on bringing the Slovak translation of gPodder up to date.
I guess `Renamed 1 downloaded episodes` is passable in English, but it stands out in other languages, especially those with multiple plural forms (like Slovak):
```
Renamed 1 episode    -> Premenovaná 1 epizóda
Renamed 2-4 episodes -> Premenované 2 epizódy
Renamed 5+ episodes  -> Premenovaných 5 epizód
```
`make messages` will need to be rerun as well, to update `messages.pot` and individual translations.

### DIFF
--- a/share/gpodder/extensions/rename_download.py
+++ b/share/gpodder/extensions/rename_download.py
@@ -14,6 +14,7 @@ from gpodder.model import PodcastEpisode
 logger = logging.getLogger(__name__)
 
 _ = gpodder.gettext
+N_ = gpodder.ngettext
 
 __title__ = _('Rename episodes after download')
 __description__ = _('Rename episodes to "<Episode Title>.<ext>" on download')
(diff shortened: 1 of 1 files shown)


elelay:
merged, thanks!

elelay:
By the way, do you have a strong opinion on #1551? If so I would like to ear from you :-)

neodyne:
> By the way, do you have a strong opinion on #1551? If so I would like to [h]ear from you :-)

@elelay Not necessarily a strong opinion, but I agree with "--:--", it gets the point across and doesn't have translation issues. Filesize / transfer speed should be translated (so `Ko/s` in French?, `КБ/с` in Russian etc.).

By the way, the Slovak translation in #1555 should be ready to be merged; I also submitted #1556, if you'd take a look.

PR #1580, opened 2023-11-03
Fix appveyor build

auouymous:


auouymous:
The switch to setuptools in #1571 is causing appveyor to fail https://ci.appveyor.com/project/elelay/gpodder/builds/48434410 while installing charset-normalizer.

auouymous:
@elelay Do you know what is causing the `build_info.py` failure when using setuptools instead of distutils?

Issue #1511, opened 2023-05-23
Wrong youtube stream address passed to Potplayer


Tienssy:
Hello,

I wanted to bring to your attention an issue I'm experiencing with playing the "Stream" feature on gpodder, specifically with video streams from YouTube. Currently, when I select the option to open the stream from a loaded video episode on gpodder's list, using PotPlayer as my video and audio player (_"C:\Program Files\DAUM\PotPlayer\PotPlayerMini64.exe"_), it takes an unusually long time to load and does not function correctly afterward. The playback is quite buggy. However, when I manually open the YouTube video address directly in PotPlayer, everything works smoothly as expected.

I would like to request a modification to the format of the YouTube address that is generated and passed to PotPlayer when selecting the "Stream" option. The current address format, such as "**rr1--sn-2b7sn7y.googlevideo.com/videoplayback...**", does not work with PotPlayer. Instead, a correctly formatted YouTube address like "**https://www.youtube.com/watch?v=aK3lhinHd_8**" is needed for proper playback in PotPlayer.

Thank you for your attention to this matter.

Tienssy:
So it won't be possible to repair it? It's a shame not to have functional streaming when it would be enough for the "Stream" function to pass the youtube URL in the correct format to the PotPlayer.

PR #1583, opened 2023-11-11
Add PotPlayer Support For Windows

TheNightRider12:
This pull request aims to add support for the Pot Player media playing app. Specifically for Windows. As an alternate to apps such as VLC, Windows Media Player Legacy, etc,. I recently found this app, as of tonight. So I thought why not include it and make it ready to go for anyone else that uses Pot Player.


elelay:
Thanks for the PR

According to [wikipedia](https://en.wikipedia.org/wiki/PotPlayer) it used to bundle ad/spyware but not anymore and a cursory glance of the terms of use seems ok to me.

Is it a popular media player?

TheNightRider12:
I found it while I was searching for alternatives to VLC due to VLC being weird with a few podcasts I listen to. It seems to be pretty popular to me. I was reading various VLC alternative reddit posts and such and a quite a bit of people said they have used Pot Player for a long time. I have been using it the past few days when listening to Podcasts and I haven't had any issues or concerns with it.

auouymous:
> Is it a popular media player?

It has a Wikipedia page, and at least one other user likes it: #1511.

@TheNightRider12 Thank you for adding this.

PR #1590, opened 2023-11-24
subscription_stats.py Extension

ndias:
Show publishing statistics for subscriptions, original from Brand Huntsman <http://qzx.com/mail/ , add column for  last feed update and last episode update dates.
Ticket with this requests 
https://github.com/gpodder/gpodder/pull/978


anewuser:
I get this error when I try to use your version of this script:

```
[gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "C:\gpodder\data\lib\python3.11\site-packages\gpodder\gtkui\main.py", line 404, in <lambda>
    return lambda action, param: callback()
                                 ^^^^^^^^^^
  File "C:\gpodder\data\share\gpodder\extensions\subscription_stats.py", line 134, in open_dialog
    last_feed = datetime.strptime(channel_last, '%a, %d %b %Y %H:%M:%S %Z').strftime('%Y/%m/%d')
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.11/_strptime.py", line 568, in _strptime_datetime
  File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.11/_strptime.py", line 349, in _strptime
ValueError: time data 'Tue, 27 Jun 2023 23:14:29 GMT' does not match format '%a, %d %b %Y %H:%M:%S %Z'

```

I've been listening to the full archives of some podcasts, so I haven't done a full check for updates in ages. If this issue happens because gPodder has changed its timestamp format, you don't need to stress over it.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+        return [(_("Subscription Statistics"), self.open_dialog)]
+
+    def add_page(self, notebook, type, channels):
+        scrolled = Gtk.ScrolledWindow()
+        scrolled.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
+
+        store = Gtk.ListStore(str, str, str, str)
+        for average, name, last, feed_last, paused in channels:
+            store.append([
+                ('%.1f' % round(average, 1)) if average > 0 else '?',
```
tpikonen, line 46: Dangling whitespace.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+                    if total == 0:
+                        edate = row[0]
+                    total += (prev - row[0])
+                    nr_episodes += 1
+                    prev = row[0]
+                cur.close()
+
+            average = (total / nr_episodes) / 86400 if nr_episodes > 0 else 0
+            last = strftime('%Y/%m/%d', localtime(edate))
+            if channel_last:
```
tpikonen, line 133: Dangling whitespace.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def on_ui_object_available(self, name, ui_object):
+        if name == 'gpodder-gtk':
+            self.gpodder = ui_object
+
+    def on_create_menu(self):
+        # extras menu
+        return [(_("Subscription Statistics"), self.open_dialog)]
+
+    def add_page(self, notebook, type, channels):
```
tpikonen, line 39: 'type' is a python builtin, maybe 'category' would be better?

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+
+        channellastcell = Gtk.CellRendererText()
+        channellastcell.set_property('xalign', 0)
+        channellastcell.set_property('alignment', Pango.Alignment.LEFT)
+        channellastcolumn = Gtk.TreeViewColumn(_('Last Feed Updated'))
+        channellastcolumn.pack_start(channellastcell, True)
+        channellastcolumn.add_attribute(channellastcell, 'text', 3)
+        tree.append_column(channellastcolumn)
+
+        notebook.append_page(scrolled, Gtk.Label('%d %s %s' % (len(channels), type, type)))
```
tpikonen, line 85: This prints the type twice ('n daily daily').

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+                    if row is None:
+                        break
+                    if total == 0:
+                        edate = row[0]
+                    total += (prev - row[0])
+                    nr_episodes += 1
+                    prev = row[0]
+                cur.close()
+
+            average = (total / nr_episodes) / 86400 if nr_episodes > 0 else 0
```
tpikonen, line 131: Expanding this product to `24 * 60 * 60` might be clearer.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+
+        label = Gtk.Label(_('Average days between the last 25 episodes.'))
+        label.set_line_wrap(True)
+        box.add(label)
+
+        hbox = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, spacing=0)
+        button = Gtk.Button(stock=Gtk.STOCK_CLOSE)
+        button.connect('clicked', lambda w: dlg.destroy())
+        hbox.pack_end(button, False, False, 0)
+        box.add(hbox)
```
tpikonen, line 182: It would be easier to just use `dlg.add_button(_('_Close'), Gtk.ResponseType.OK)` here. That way you also don't need the deprecated GtkStock.

tpikonen (commented):
Nice, except for the few nitpicks.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+                        break
+                    if total == 0:
+                        edate = row[0]
+                    total += (prev - row[0])
+                    nr_episodes += 1
+                    prev = row[0]
+                cur.close()
+
+            average = (total / nr_episodes) / 86400 if nr_episodes > 0 else 0
+            last = strftime('%Y/%m/%d', localtime(edate))
```
tpikonen, line 132: It would be better to use the ISO date format ('%Y-%m-%d') here and on the other date,

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+        lastcolumn.add_attribute(lastcell, 'text', 2)
+        tree.append_column(lastcolumn)
+
+        channellastcell = Gtk.CellRendererText()
+        channellastcell.set_property('xalign', 0)
+        channellastcell.set_property('alignment', Pango.Alignment.LEFT)
+        channellastcolumn = Gtk.TreeViewColumn(_('Last Feed Updated'))
+        channellastcolumn.pack_start(channellastcell, True)
+        channellastcolumn.add_attribute(channellastcell, 'text', 3)
+        tree.append_column(channellastcolumn)
```
tpikonen, line 83: Does this column add much information compared to the date of the last episode? It's also empty for feeds fetched with the youtube-dl extension.

ndias:
It seems this is due to different ways python versions are  treat the different locales!
I'm changing the extension.

> I get this error when I try to use your version of this script:
> 
> ```
> [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
>   File "C:\gpodder\data\lib\python3.11\site-packages\gpodder\gtkui\main.py", line 404, in <lambda>
>     return lambda action, param: callback()
>                                  ^^^^^^^^^^
>   File "C:\gpodder\data\share\gpodder\extensions\subscription_stats.py", line 134, in open_dialog
>     last_feed = datetime.strptime(channel_last, '%a, %d %b %Y %H:%M:%S %Z').strftime('%Y/%m/%d')
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.11/_strptime.py", line 568, in _strptime_datetime
>   File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.11/_strptime.py", line 349, in _strptime
> ValueError: time data 'Tue, 27 Jun 2023 23:14:29 GMT' does not match format '%a, %d %b %Y %H:%M:%S %Z'
> ```
> 
> I've been listening to the full archives of some podcasts, so I haven't done a full check for updates in ages. If this issue happens because gPodder has changed its timestamp format, you don't need to stress over it.

share/gpodder/extensions/subscription_stats.py
ndias: easy to repair

share/gpodder/extensions/subscription_stats.py
ndias: ok, trivial change

share/gpodder/extensions/subscription_stats.py
ndias: Ok.

share/gpodder/extensions/subscription_stats.py
ndias: The reason I use the %Y/%m/%d format was because in the mail window of gpodder in the released date we have this format ... just to be consistent ... but have no strong feelings about the format to use

share/gpodder/extensions/subscription_stats.py
ndias: Maybe you are right, I just put this column because there are some feeds that have recent dates compared with the last episode date, but like you wrote some feeds don't have this date that's the reason they are empty, after some thoughts I will remove this column.

share/gpodder/extensions/subscription_stats.py
ndias: yes, a bug from my part ...

anewuser:
It's working now, but can you please make the columns sortable and resizable?

auouymous:
Since you want to finish this for me, here is what kept me from releasing it, other than adding the 'last episode' column.

'%s subscriptions (%d paused)' should be %d not %s.

The 25 count should be clamped to `episode.limit` in the label, it doesn't matter for the database query. That way users with a low limit don't think it is some how averaging episodes they don't have.

Should it have a documentation link to a page in https://github.com/gpodder/gpodder.github.io/tree/master/docs/extensions?

share/gpodder/extensions/subscription_stats.py
ndias: > The main window uses `%x`, which changes with locale.

Ok, nice, I didn't check the code only check in the GUI and that was an error from my part ... I've change to use %x that make sense

ndias:
> Since you want to finish this for me, here is what kept me from releasing it, other than adding the 'last episode' column.
> '%s subscriptions (%d paused)' should be %d not %s.

Done.

> The 25 count should be clamped to `episode.limit` in the label, it doesn't matter for the database query. That way users with a low limit don't think it is some how averaging episodes they don't have.

Sorry, can you explain better this? 

> Should it have a documentation link to a page in https://github.com/gpodder/gpodder.github.io/tree/master/docs/extensions?

Done.

ndias:
> It's working now, but can you please make the columns sortable and resizable?

Done, you can test the new version

tpikonen:
It would be nice to be able to use the statistics dialog in also in narrow screens (i.e. phones). Here's a suggestion on how to make it shrinkable and almost borderless. Since the podcast name is set to ellipsize at the end of the string, the pause indicator is moved to the beginning and shortened to a symbol. The close button and labels are now without margins, so they should be padded separately.

```diff
diff --git a/share/gpodder/extensions/subscription_stats.py b/share/gpodder/extensions/subscription_stats.py
index b1d52e31..66ab927e 100644
--- a/share/gpodder/extensions/subscription_stats.py
+++ b/share/gpodder/extensions/subscription_stats.py
@@ -44,9 +44,9 @@ class gPodderExtension:
         store = Gtk.ListStore(str, str, str)
         for average, name, last, paused in channels:
             store.append([
                 ('%.1f' % round(average, 1)) if average > 0 else '?',
-                name if not paused else (_('%s (paused)') % name), last,
+                ('‖ ' if paused else '') + name, last,
             ])
 
         tree = Gtk.TreeView(model=store)
         scrolled.add(tree)
@@ -62,12 +62,14 @@ class gPodderExtension:
 
         channelcell = Gtk.CellRendererText()
         channelcell.set_property('xalign', 0)
         channelcell.set_property('alignment', Pango.Alignment.LEFT)
+        channelcell.set_property('ellipsize', Pango.EllipsizeMode.END)
         channelcolumn = Gtk.TreeViewColumn(_('Podcast'))
         channelcolumn.set_sort_column_id(1)
         channelcolumn.pack_start(channelcell, True)
         channelcolumn.add_attribute(channelcell, 'text', 1)
+        channelcolumn.set_expand(True)
         tree.append_column(channelcolumn)
 
         lastcell = Gtk.CellRendererText()
         lastcell.set_property('xalign', 0)
@@ -149,15 +151,16 @@ class gPodderExtension:
         dlg.set_size_request(400, 400)
         dlg.set_resizable(True)
 
         box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=10)
-        box.set_border_width(10)
+        box.set_border_width(0)
 
         label = Gtk.Label(_('%d subscriptions (%d paused)') % (len(channels), nr_paused))
         box.add(label)
 
         notebook = Gtk.Notebook()
         notebook.set_vexpand(True)
+        notebook.set_scrollable(True)
         self.add_page(notebook, _('daily'), daily)
         self.add_page(notebook, _('weekly'), weekly)
         self.add_page(notebook, _('monthly'), monthly)
         self.add_page(notebook, _('yearly'), yearly)

```

auouymous:
> > The 25 count should be clamped to `episode.limit` in the label, it doesn't matter for the database query. That way users with a low limit don't think it is some how averaging episodes they don't have.
> 
> Sorry, can you explain better this?

```patch
-label = Gtk.Label(_('Average days between the last 25 episodes.'))
+label = Gtk.Label(_('Average days between the last %d episodes.') % (episode.limit if episode.limit < 25 else 25))
```

Something like that.

ndias:
> The Days column is now sorted as string, which does not work as it should with strings which are numbers (the sorting is ok on start though). One solution would be to set a custom sort function with `.set_sort_func()`.

Done, I think the sort now works, I use another column with the epoc date and the sort is made with that column

ndias:
> > > The 25 count should be clamped to `episode.limit` in the label, it doesn't matter for the database query. That way users with a low limit don't think it is some how averaging episodes they don't have.
> > 
> > 
> > Sorry, can you explain better this?
> 
> ```diff
> -label = Gtk.Label(_('Average days between the last 25 episodes.'))
> +label = Gtk.Label(_('Average days between the last %d episodes.') % (episode.limit if episode.limit < 25 else 25))
> ```
> 
> Something like that.

Ok, thanks add this change

ndias:
> It would be nice to be able to use the statistics dialog in also in narrow screens (i.e. phones). Here's a suggestion on how to make it shrinkable and almost borderless. Since the podcast name is set to ellipsize at the end of the string, the pause indicator is moved to the beginning and shortened to a symbol. The close button and labels are now without margins, so they should be padded separately.

Add this changes, thanks

ndias:
> ```
> [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
>   File "C:\gpodder\data\lib\python3.11\site-packages\gpodder\gtkui\main.py", line 404, in <lambda>
>     return lambda action, param: callback()
>                                  ^^^^^^^^^^
>   File "C:\gpodder\data\share\gpodder\extensions\subscription_stats.py", line 174, in open_dialog
>     label = Gtk.Label(_('Average days between the last %d episodes.') % (episode.limit if episode.limit < 25 else 25))
>                                                                                           ^^^^^^^
> NameError: name 'episode' is not defined
> ```



> ```
> [gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
>   File "C:\gpodder\data\lib\python3.11\site-packages\gpodder\gtkui\main.py", line 404, in <lambda>
>     return lambda action, param: callback()
>                                  ^^^^^^^^^^
>   File "C:\gpodder\data\share\gpodder\extensions\subscription_stats.py", line 174, in open_dialog
>     label = Gtk.Label(_('Average days between the last %d episodes.') % (episode.limit if episode.limit < 25 else 25))
>                                                                                           ^^^^^^^
> NameError: name 'episode' is not defined
> ```

Sorry, the problem is solved.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        lastcolumn.set_sort_column_id(3)
+        lastcolumn.pack_start(lastcell, True)
+        lastcolumn.add_attribute(lastcell, 'text', 2)
+        tree.append_column(lastcolumn)
+
+        edatecell = Gtk.CellRendererText()
+        edatecolumn = Gtk.TreeViewColumn()
+        edatecolumn.add_attribute(edatecell, 'text', 2)
+        edatecolumn.set_visible(False)
+        tree.append_column(edatecolumn)
```
tpikonen, line 88: This CellRenderer and column are not necessary, adding them to the TreeView without being set visible leads to Gtk errors. Sorting works fine also without this.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+            ])
+
+        tree = Gtk.TreeView(model=store)
+        scrolled.add(tree)
+
+        dayscell = Gtk.CellRendererText()
+        dayscell.set_property('xalign', 1)
+        dayscell.set_property('alignment', Pango.Alignment.RIGHT)
+        dayscolumn = Gtk.TreeViewColumn(_('Days'))
+        dayscolumn.set_sort_column_id(0)
```
tpikonen, line 59: The Days column sorting does not work correctly, you need to fix it like in the `last` column by adding a column with the `average` as float to the ListStore and using it as a sort column.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # open dialog
+        dlg = Gtk.Dialog(_('Subscription Statistics'), self.gpodder.main_window)
+        dlg.set_size_request(400, 400)
+        dlg.set_resizable(True)
+
+        box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=10)
+        box.set_border_width(0)
+
+        label = Gtk.Label(_('%d subscriptions (%d paused)') % (len(channels), nr_paused))
```
tpikonen, line 163: This label should have a larger top margin.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        notebook.set_vexpand(True)
+        notebook.set_scrollable(True)
+        self.add_page(notebook, _('daily'), daily)
+        self.add_page(notebook, _('weekly'), weekly)
+        self.add_page(notebook, _('monthly'), monthly)
+        self.add_page(notebook, _('yearly'), yearly)
+        box.add(notebook)
+
+        conf = config.Config(gpodder.config_file)
+        label = Gtk.Label(_('Average days between the last %d episodes.') % (conf.limit.episodes if conf.limit.episodes < 25 else 25))
```
tpikonen, line 176: This label should have larger bottom margin.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        label.set_line_wrap(True)
+        box.add(label)
+
+        dlg.add_button(_('_Close'), Gtk.ResponseType.OK)
```
tpikonen, line 180: The close button should also have larger margins, you can get the button widget here as return value and set the margin property there.

tpikonen (changes requested):

share/gpodder/extensions/subscription_stats.py
ndias: You are right, removed

share/gpodder/extensions/subscription_stats.py
ndias: Although I can't see the problem reported I add the column and the sort is made there

share/gpodder/extensions/subscription_stats.py
ndias: Done.

share/gpodder/extensions/subscription_stats.py
ndias: Do you mean top margin? if yes done.

share/gpodder/extensions/subscription_stats.py
ndias: Done, the button has larger margins

share/gpodder/extensions/subscription_stats.py
```diff
@@ -66,27 +67,21 @@ def add_page(self, notebook, category, channels):
         channelcell.set_property('alignment', Pango.Alignment.LEFT)
         channelcell.set_property('ellipsize', Pango.EllipsizeMode.END)
         channelcolumn = Gtk.TreeViewColumn(_('Podcast'))
```
tpikonen, line 69: Maybe also use 'Channel' here as column header?

share/gpodder/extensions/subscription_stats.py
```diff
@@ -66,27 +67,21 @@ def add_page(self, notebook, category, channels):
...
         lastcell = Gtk.CellRendererText()
         lastcell.set_property('xalign', 0)
         lastcell.set_property('alignment', Pango.Alignment.LEFT)
         lastcolumn = Gtk.TreeViewColumn(_('Last Updated'))
-        lastcolumn.set_sort_column_id(3)
+        lastcolumn.set_sort_column_id(4)
         lastcolumn.pack_start(lastcell, True)
-        lastcolumn.add_attribute(lastcell, 'text', 2)
+        lastcolumn.add_attribute(lastcell, 'text', 3)
         tree.append_column(lastcolumn)
```
tpikonen, line 83: The 'Last Updated' column is too narrow when the dialog is first shown and does not show the full date. Resizing the dialog makes it bounce to the correct width.

Adding `lastcolumn.set_min_width(130)` fixes this somehow, but it's a bit of a hack. If you can find out how to make this column to have the correct width from the start, that would be great.

tpikonen (commented):
Thanks for getting back to this. The dialog layout and functionality look mostly OK (except for the date column width) and the functionality is fine.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Show publishing statistics for subscriptions.
 # Released under the same license terms as gPodder itself.
-# version 0.6 - 2023/12/07 - Nuno Dias <Nuno.Dias+gpodder@gmail.com>
+# version 0.7 - 2024/01/15 - Nuno Dias <Nuno.Dias+gpodder@gmail.com>
```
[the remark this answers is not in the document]
ndias: deleted

share/gpodder/extensions/subscription_stats.py
ndias: Changed to Channel.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -1,7 +1,7 @@
...
 # Add Last Episode updates, sort columns and other minor changes.
 
 import datetime
```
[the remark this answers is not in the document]
ndias: deleted import

share/gpodder/extensions/subscription_stats.py
ndias: I have no solution for this problem,  I've allowed the column to expand and this solve the problem when the dialog is show, but if the dialog is resized the column is resized also ...

share/gpodder/extensions/subscription_stats.py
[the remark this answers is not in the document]
ndias: make lint and flake8 run and problems resolved.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        # sort by averages
+        daily.sort(key=lambda e: e[0])
+        weekly.sort(key=lambda e: e[0])
+        monthly.sort(key=lambda e: e[0])
+        yearly.sort(key=lambda e: e[0])
+
+        # open dialog
+        dlg = Gtk.Dialog(_('Subscription Statistics'),
+                         self.gpodder.main_window)
+        dlg.set_size_request(400, 400)
```
tpikonen, line 150: `set_size_request()` sets a fixed minimum size, so shrinking the dialog does not work. It's better to use `set_default_size()`. Setting `dlg.set_default_size(420, 600)` looked good at least on my screen with all the notebook tabs visible.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        box.set_border_width(0)
+
+        label = Gtk.Label(_('%d subscriptions (%d paused)') % (len(channels),
+                          nr_paused))
+        label.set_padding(0, 10)
+        box.add(label)
+
+        notebook = Gtk.Notebook()
+        notebook.set_vexpand(True)
+        notebook.set_scrollable(True)
```
tpikonen, line 163: Adding `notebook.set_show_border(False)` here removes the small but useless border.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,185 @@
...
+        channelcell.set_property('ellipsize', Pango.EllipsizeMode.END)
+        channelcolumn = Gtk.TreeViewColumn(_('Channel'))
+        channelcolumn.set_sort_column_id(2)
+        channelcolumn.pack_start(channelcell, True)
+        channelcolumn.add_attribute(channelcell, 'text', 2)
+        channelcolumn.set_expand(True)
+        tree.append_column(channelcolumn)
+
+        lastcell = Gtk.CellRendererText()
+        lastcolumn = Gtk.TreeViewColumn(_('Last Updated'))
```
tpikonen, line 70: The header in this column is now usually wider than the values. Using just 'Updated' as column header would take less space.

tpikonen (changes requested):
Thanks. I still found some things to tweak, but it's almost there.

share/gpodder/extensions/subscription_stats.py
tpikonen: Putting the 'Channel' column last and removing `lastcolumn.set_expand(True)` seems to work fine. The channel column then receives all the extra space. Its strings are not ellipsized when the dialog is first rendered, but I think we can live with that.

ndias: That is a good idea, I've made some more tests and I think maybe the culprit is the  ellipsize, if I comment that line the Updated column is ok, but in the other hand the dialogue  doesn't resize to the initial size!
No solution for this so, I follow your advice and changed the order of the columns

share/gpodder/extensions/subscription_stats.py
ndias: The 400,400 was suggested from someone because of phones, but 420,600 maybe ok also for phones, so I changed this.

share/gpodder/extensions/subscription_stats.py
ndias: Ok, chnaged.

share/gpodder/extensions/subscription_stats.py
ndias: Done.

share/gpodder/extensions/subscription_stats.py
auouymous: This is probably something we should deal with soon. The menus, notebook tab, and "All episodes - from all podcasts" use "podcast", while both context menus use "channel". Several extensions use "podcast" in their description, and the youtube-dl extension uses "channel" on its preferences page.

tpikonen (approved):

share/gpodder/extensions/subscription_stats.py
tpikonen: Grepping `messages.pot` gives about 50 results for '[Pp]odcast'. I can convert these to 'channel' where applicable, if that's OK?

auouymous: @tpikonen I have no problem changing to channel as I am guilty of using it in places. However, channel really only applies to youtube channels, with most feeds being called podcasts. And the title on gpodder.org is "gPodder: Media aggregator and podcast client". We could start a discussion to rename it to channel, but podcast is probably better.

share/gpodder/extensions/subscription_stats.py
```diff
@@ -41,11 +41,12 @@ def add_page(self, notebook, category, channels):
...
 
-        store = Gtk.ListStore(str, str, str)
-        for average, name, last, paused in channels:
+        store = Gtk.ListStore(str, str, str, int)
+        for average, name, edate, paused in channels:
+            last = strftime('%x', localtime(edate))
             store.append([
                 ('%.1f' % round(average, 1)) if average > 0 else '?',
-                name if not paused else (_('%s (paused)') % name), last,
+                ('‖ ' if paused else '') + name, last, edate,
```
auouymous: The `‖` is small and hard to see, does anyone have an objection to using two heavy vertical bars `❚❚`?

share/gpodder/extensions/subscription_stats.py
```diff
@@ -40,46 +36,42 @@ def on_create_menu(self):
...
-                average_flot, ('‖ ' if paused else '') + name, last, edate,
+                average, ('‖ ' if paused else '') + name, last, edate,
             ])
 
         tree = Gtk.TreeView(model=store)
         scrolled.add(tree)
 
         dayscell = Gtk.CellRendererText()
-        dayscell.set_property('xalign', 1)
-        dayscell.set_property('alignment', Pango.Alignment.RIGHT)
```
auouymous: I can't find the conversation on why the days column was changed from right aligned to centered, but it is hard to compare numbers when centered. It works when all of the numbers have the same number of digits, but the weekly and yearly categories can have varying widths. Right aligning also pulls the numbers closer to the podcast name, making it easier to associate them.

share/gpodder/extensions/subscription_stats.py
auouymous: Would moving the updated column far left also work? The wide date pulls the days column further from the podcast name, requiring more tracing across to associate them.

auouymous (changes requested):

share/gpodder/extensions/subscription_stats.py
```diff
@@ -0,0 +1,186 @@
...
+        if name == 'gpodder-gtk':
+            self.gpodder = ui_object
+
+    def on_create_menu(self):
+        # extras menu
+        return [(_("Subscription Statistics"), self.open_dialog)]
+
+    def add_page(self, notebook, type, channels):
+        scrolled = Gtk.ScrolledWindow()
+        scrolled.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
```
auouymous, line 39: The window width can be resized pretty small now, so the NEVER should be changed to AUTOMATIC, or remove the entire line since both default to automatic. However, something is causing the horizontal scrollbar to appear when the dialog is opened, and resizing narrower or wider hides it. Anyone know why?

Should we also set a minimum width for the channel name column? The name can be unusably short when resized to minimum width, more so for a paused channel.

auouymous (changes requested):

share/gpodder/extensions/subscription_stats.py
ndias: The alignment centered was tests from my part to see if resolved the issues with the Updated column (when was the last column), and was left in the last version, nevertheless I've changed also the position of the columns so next to the Days column are now the Updated column.
So it's enough to change again to dayscell.set_property('xalign', 1)
I also remove the dayscell.set_property('alignment', Pango.Alignment.RIGHT) because I think is not necessary, but I could be wrong :)

auouymous: You are correct, the alignment property is not needed here. It is only required if the cell can have more than one line.

share/gpodder/extensions/subscription_stats.py
auouymous: `channelcell.set_property('width-chars', 20)` makes the column at least 20 characters wide before ellipsizing it. This only matters if we decide to enable the horizontal scrollbar, but it would be better to just not ellipsize it, and then the width-char property isn't needed.

I don't care if we keep it as is (and not need to worry about the broken scrollbar), or enable the scrollbar and get rid of the ellipsis so the full name displays. Anyone with a narrow window might also have a narrow gpodder window and channel names in the podcast and episode lists would be ellipsized.

@tpikonen Do you have a preference?

ndias: I've made some more tests and I think the best solution is to get rid of the ellipsis, that was introduced when the Channel was not the last column, so if the final design in the columns is "Updated | Days | Channel" we don't have anymore the problems that led to the introduction of the ellipsis.

auouymous:
I'm not sure how to fix the horizontal scrollbar issue, but it should be okay to merge it as is. Everyone agreed on changing `‖` to `❚❚` so that could be done. And the `Channel` column name should be changed to `Podcast` until we discuss renaming it through out the project. The use of 'channel' in the source is fine.

ndias:
> I'm not sure how to fix the horizontal scrollbar issue, but it should be okay to merge it as is. Everyone agreed on changing `‖` to `❚❚` so that could be done. And the `Channel` column name should be changed to `Podcast` until we discuss renaming it through out the project. The use of 'channel' in the source is fine.

Ok, I upload a new version with the requested changes.

auouymous:
@ndias Thank you for finishing this for me!

@tpikonen Thank you for reviewing it.

auouymous:
Open the `Extras` menu.

PR #1592, opened 2023-12-15
Add socks to settings

masoudd:
- [x] Add proxy input to settings
- [x] Podcast download uses the proxy settings
- [x] Feed update and Add podcast should use the proxy settings
- [x] Add username and password support for proxy
- [x] Show the proxy environment variables in the proxy settings tab if they are set. (They would take effect only when "Use proxy" is disabled)
- [x] youtube-dl should use the proxy settings

Current UI:
proxy5

Please let me know if anything needs to change.

### DIFF
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -338,10 +338,6 @@ class gPodderPreferences(BuilderWidget):
                                               self.checkbutton_delete_deleted_episodes)
         self._config.connect_gtk_togglebutton('device_sync.compare_episode_filesize',
                                               self.checkbutton_compare_episode_filesize)
-        self._config.connect_gtk_togglebutton('network.use_proxy',
-                                              self.checkbutton_use_proxy)
-        self._config.connect_gtk_togglebutton('network.proxy_use_username_password',
-                                              self.checkbutton_proxy_use_username_password)
 
         # Have to do this before calling set_active on checkbutton_enable
         self._enable_mygpo = self._config.mygpo.enabled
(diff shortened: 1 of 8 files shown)


src/gpodder/util.py
```diff
@@ -1259,7 +1260,9 @@ def urlopen(url, headers=None, data=None, timeout=None, **kwargs):
     s.mount('http://', a)
     s.mount('https://', a)
     headers.update({'User-agent': gpodder.user_agent})
-    return s.get(url, headers=headers, data=data, timeout=timeout, **kwargs)
+    proxies = config._proxies
+    logger.debug(f"urlopen: url: {url}, proxies:{proxies}")
```
auouymous, line 1264: Missing a space in `proxies:{proxies}`.

src/gpodder/gtkui/main.py
```diff
@@ -93,6 +93,10 @@ def __init__(self, app, bus_name, gpodder_core, options):
         self.refresh_episode_dates()
 
         self.on_episode_list_selection_changed_id = None
+        observer = gpodder.config.get_network_proxy_observer(self.config)
```
auouymous, line 96: Need a blank line to separate the observer code from the episode list line.

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -317,6 +336,8 @@ def new(self):
                                               self.checkbutton_delete_deleted_episodes)
         self._config.connect_gtk_togglebutton('device_sync.compare_episode_filesize',
                                               self.checkbutton_compare_episode_filesize)
+        self._config.connect_gtk_togglebutton('network.use_proxy',
```
auouymous, line 341: All of the proxy code should be grouped together, instead of spread out over the function.

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -805,10 +808,20 @@ def on_btn_playlist_folder_clicked(self, widget):
...
+        widgets = (self.grid_network_proxy_details,
+                   self.vbox_network_proxy_username_password)
         if widget.get_active():  # Enable the proxy input details
-            self.grid_network_proxy_details.set_sensitive(True)
+            for w in widgets:
+                w.set_sensitive(True)
         else:  # Disable
-            self.grid_network_proxy_details.set_sensitive(False)
+            for w in widgets:
+                w.set_sensitive(False)
```
auouymous, line 818: This could be done with a single for loop and `w.set_sensitive(widget.get_active())`.

auouymous (changes requested):
gpo crashes with this patch installed.


> - [ ] youtube-dl should use the proxy settings (it's a separate plugin with a settings menu of its own)

The proxy prefs are in gpodder, any extension capable of using the proxy prefs should use it.

https://stackoverflow.com/questions/46389519/how-to-embed-socks5-proxy-in-youtube-dls-python-code

auouymous (changes requested):
gpo doesn't initialize the `_proxies` variable. Adding the following to the end of gPodderCli __init__ seems to work.

```
        observer = gpodder.config.get_network_proxy_observer(self._config)
        self._config.add_observer(observer)
        # Trigger the global gpodder.config._proxies observer contraption to initialize it.
        observer("network.", None, None)
```

auouymous:
Thank you for implementing this.

Issue #1194, opened 2021-11-11
Add an option for the dark theme


xmha97:
Hi, now gPodder supports the dark theme.
![image](https://user-images.githubusercontent.com/23485114/141367845-dec0606b-0e65-4821-86a6-ffe69aa1604e.png)
```
gtk-application-prefer-dark-theme = true
```
Please add an option for the dark theme on settings.
![image](https://user-images.githubusercontent.com/23485114/141368174-74899b39-f07d-43b0-a12e-a8de77e4297e.png)

elelay:
I don't see a way for the gPodder app to do it easily

auouymous:
Adding the following to `bin/gpodder` changes the theme from inside gpodder. We could have a config to choose between system theme, `Adwaita`, `Adwaita-dark`, or an option to enter a custom theme name. It might also be possible to scan known theme directories to provide more options.

```patch
                 logger.info('Cannot connect to remote object.', exc_info=True)
 
     if gpodder.ui.gtk:
+        if "GTK_THEME" not in os.environ:
+            os.environ["GTK_THEME"] = "Adwaita-dark"
+
         from gpodder.gtkui import app
         gpodder.ui_folders.insert(0, os.path.join(ui_folder, 'gtk'))
         app.main(options)
```

But I don't think `Settings.json` is loaded before the `from gpodder.gtkui import app` line, and setting GTK_THEME in `do_activate` in `gtkui/app.py`, where Core is initialized, is too late to work.

PR #1595, opened 2023-12-28
Add color scheme (dark and light mode) support

tpikonen:
Add 'ui.gtk.color_scheme' config variable with 'system', 'light' and 'dark' string values. Config observer in main.py changes the color to match the config.

System color scheme is read from freedesktop.org Settings portal via Dbus.

The setting can be changed with a combo box in the preferences window via new function connect_gtk_combo_box_text().

This fixes issue #1194, but only on Linux and with the Settings portal running.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
 <!--*- mode: xml -*-->
 <interface>
   <requires lib="gtk+" version="3.16"/>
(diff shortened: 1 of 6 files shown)


auouymous:
It throws an exception for me
`AttributeError: 'SessionBus' object has no attribute 'call_async'`

And linux users without that portal, mac users and windows users will all see a non-functional setting in preferences, right?

tpikonen:
> It throws an exception for me `AttributeError: 'SessionBus' object has no attribute 'call_async'`

Are you perhaps using the fake dbus module? I'll add a dummy `call_async` method there, so this should work.

> And linux users without that portal, mac users and windows users will all see a non-functional setting in preferences, right?

The 'System' setting will not work in this case, but forcing the light and dark modes should work. This could be either left as is, or a non-compatible platform (fake dbus, windows, mac) could be detected and the system setting removed in that case.

auouymous:
Yes, I am using fake dbus. The setting should be removed for anyone who can't use it, otherwise we will get a lot of reports that it doesn't work.

tpikonen:
This should now have better chances of working without Dbus or Settings portal. There are 3 kinds of platforms this should work in:

 1. With Dbus and Settings portal running
 2. With Dbus, but without Settings portal
 3. Without Dbus

In the last two cases the 'system' color scheme is interpreted as 'light' and the config key is also changed to 'light' in preferences window initialization. Unfortunately I can easily test only the case where Settings are available.

auouymous:
So there is no way to detect `gtk-application-prefer-dark-theme` in settings.ini to set the default? It ignored my preference and started up in light mode, however, changing the new setting in gpodder did work and did persist across gpodder restarts.

I haven't yet reviewed the patch, but I would like to verify it works on Windows and Mac. If only PRs produced Mac builds, because the Windows build is currently not working.

tpikonen:
> So there is no way to detect gtk-application-prefer-dark-theme in settings.ini to set the default?

It's probably not possible to detect that this setting is set in the GTK config file (or not).

It can be read at startup and set as default, but it would require an extra 'unset' state in the config key, so that it would only be made once. But this should only be made when a portal is not found, since the default in that case should be the 'system' setting.

auouymous:
> it would require an extra 'unset' state in the config key

Assuming it is even possible to detect the value, the config key could default to `None`, and that value triggers the detection.

tpikonen:
> Assuming it is even possible to detect the value, the config key could default to None, and that value triggers the detection.

Good idea. Now this detects a non-default dark mode GtkSetting and uses it as default. The whole thing got a bit too complex for my taste, but at least it's backwards compatible.

auouymous:
Works for me. The only problem is that it starts up briefly in the system color before it switches to the preference in gpodder. However, there is most likely no way around this, and anyone who can't easily configure 'settings.ini' probably won't mind a short white flash to get dark mode.

I'll review the patch tomorrow.

auouymous:
Thanks for implementing this.

auouymous:
>     * System theme detection: Is it possible to add a third entry like "OS setting" (or something like that) to the theme selection list in addition to dark/light?

Are you using Mac or Windows? The Linux build should have a system option, but Mac and Windows builds do not since gtk's settings is inside the build and probably difficult to edit. We also can't easily access a system setting on Mac and Windows without platform code for each one.


>     * Contrast refinement (see screenshot): The contrast of the purely white text is quite high in comparison to the dark background. Not easy on the eyes, I think. I would suggest that the text could me more kind of a mid grey color. Like midway between plain white and the grey color of the column title.

That is only possible by changing gtk's theme, on Linux. Mac and Windows users are stuck with the Adwaita theme included in the build.

PR #1596, opened 2024-01-14
Allow manual update checks via menu item.

auouymous:
Notify user that gPodder is managed by their distribution when the `no-update-check` file is present.

### DIFF
--- a/src/gpodder/gtkui/main.py
+++ b/src/gpodder/gtkui/main.py
@@ -3707,11 +3707,6 @@ class gPodder(BuilderWidget, dbus.service.Object):
     def on_homepage_activate(self, widget, *args):
         util.open_website(gpodder.__url__)
 
-    def check_for_distro_updates(self):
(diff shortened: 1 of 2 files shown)


src/gpodder/gtkui/main.py
```diff
@@ -3740,12 +3735,24 @@ def check_for_updates(self, silent):
...
                 _('Newest version: %s') % version,
                 _('Release date: %s') % released,
-                '',
-                _('Download the latest version from gpodder.org?'),
             ])
+            if os.path.exists(gpodder.no_update_check_file):
+                message = '\n'.join([
+                    message,
+                    '',
+                    _('Please check your distribution for gPodder updates.'),
```
thp, line 3743: Is "your distribution" clear to all users? Or should it be something like "your system's package manager"?

tpikonen:
Is there now (with this PR) a way to completely disable the update check?

PR #1601, opened 2024-01-26
Modernize userappsdirs

tpikonen:
Adds tilde expansion to userappsdirs (= the list of directories where .desktop files for players are searched).

Adds the local application dir and local and system flatpak app dirs to userappsdirs.

Removes '/usr/share/applications/kde/' which does not seem to be used in modern (post 4.0) KDE (this directory does not exist in Debian, not sure about other distros).


auouymous (approved):

auouymous:
Merged, thank you.

PR #1613, opened 2024-03-14
fix Windows build

auouymous:


auouymous:
Something other than the change to setuptools has caused the Windows build to break, maybe an msys2 update. 62e89ad7b4059e5f032995eaaf2b7a767e8c16a0 was the third commit related to the Windows build, and is required to find `build_info.py` with or without the other two setuptools patches.

computerfreak105:
Hi auouymous!

 

Just to be sure this is remembered: On Feb 6th Masoud Naservand wrote

 

[The directory share/gpodder does not end up in the installers. At least for the portable version. All you need to do to make it work is to just copy the share/gpodder directory from the git repo to the data/share/ in the portable install dir. I'll investigate further for a fix]

 

I do that and can run the latest windows portable anyway. Maybe all you’d had to do is to make sure the Gpodder directory is copied again into the share folder?

 

And thank you so much, Masoud, for sharing your findings with us!

PR #980, opened 2021-03-16
OPDS extension

elelay:
> The Open Publication Distribution System (OPDS) catalog format is a syndication format for electronic publications based on Atom and HTTP (wikipedia)

This extension makes it possible to subscribe to Calibre or other ebook distribution channels from gPodder.
When books are directly available in the opds feed, they are downloaded by gPodder, like a podcast episode attachment.
Some book sellers also support OPDS catalogs, but then the book is not directly available for download.
The extension generates a link in the description in this case (gPodder being more like a news reader in this case).

PR #1615, opened 2024-03-15
add podverse search (first page only)

elelay:


share/gpodder/extensions/podverse.py
```diff
@@ -0,0 +1,72 @@
...
+__description__ = _('Search podverse podcast index')
+__authors__ = 'Eric Le Lay <elelay.fr:contact>'
+__doc__ = 'https://gpodder.github.io/docs/extensions/podverse.html'
+
+
+class PodverseDirectoryProvider(Provider):
+    def __init__(self):
+        self.name = _('Podverse search')
+        self.kind = Provider.PROVIDER_SEARCH
+        self.icon = None  # TODO: there is no way for a provider to register a custom icon
```
auouymous, line 31: Why can't the icon be placed in `share/gpodder/images/` with the others? I understand it wouldn't work for user installed extensions, and the code should be fixed to also search `~/.local/share/gpodder/images/` or `~/gPodder/Extensions/`.

share/gpodder/extensions/podverse.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def on_search(self, query):
+        # see https://api.podverse.fm/api/v1/swagger#operations-podcast-getPodcasts
+        json_url = f"https://api.podverse.fm/api/v1/podcast?page=1&searchTitle={quote_plus(query)}&sort=top-past-week"
+
+        json_data = urlopen(json_url, headers={"accept": "application/json"}).json()
```
auouymous, line 37: One of my searches retried three times, threw an exception to console and displayed a blank results pane with no indication in the UI it timed out. Should we catch the exception and open a dialog?

auouymous (changes requested):

share/gpodder/extensions/podverse.py
elelay: done

share/gpodder/extensions/podverse.py
elelay: I agree, but shouldn't it be in the gtkui/desktop/podcastdirectory code?
Really the podcast directory dialog should be reworked to remove the modal progress and display exceptions in place of content when any, what do you think?

auouymous: Ah yes, we don't want an error dialog active with the progress dialog.

elelay: Here is a POC for a revamped podcast directory dialog:
- get rid of the left/right pane: select a provider full-width, then go to search/results panel full-width
- display any error in a text area in the results

To be discussed:
- provider list seems a bit tiny or too compressed together (alternative UI using buttons or listbox is provided, see it with the glade program)
- I want to get rid of tag cloud support: it has been disabled forever
- should pressing escape bring go back to provider list (currently closes the dialog)? 
- better display of errors?
- progressbar only has 3 states: not started / in progress / finished at the moment.
  Should we get into progressive results? Then the api for DirectoryProvider should change a bit: maybe yielding results a page at a time? This would be only useful for Podverse search, for result pagination to be honest.
- I still got OPDS browser in the back of my mind, but I think a separate UI would be better (how to bolt navigation into the DirectoryProvider api and podcast directory interface?)
- Should we support search from CLI?
- move back/cancel/add buttons in headerbar, leaving select all/none on their own line at the bottom only when displaying results?

elelay: also: we should display images

auouymous: I don't know how to use glade but do see a window with a button list that looks good, although it has the bottom bar from the query window which doesn't make sense on the provider window.

I don't know what tag cloud is, I'm fine with removing something that is disabled.

Ya, pressing escape could return to the provider list, and pressing it there could then close the window.

I was able to create a JustAWarning exception in directory.py, have podverse raise it for the short query and search error, and then catch it and not log the stacktrace to console. You have a FIXME right below that to suppress the stacktrace in the window, it could detect the JustAWarning exception and only display the error message in window, and display stacktrace for other exceptions.

Three progress states is better than none. Progressive results depends on how much time you want to invest in doing it. The first commit suggests it only fetches the first page, continuing to download pages could result in a lot of requests the user might not look at. Maybe add a button at bottom of list to fetch next page, and append it to list, keeping the button at bottom until no more pages.

Are you referring to #980? I don't see any UI in it.

It would be nice if CLI could do everything the GTK client did, but time...

Buttons can be in headerbar, but remember to detect CSD and keep them in the window if not enabled.

I like the new search UI way more than the old, but the window needs to be larger by default, and maybe save/restore its size.

src/gpodder/gtkui/desktop/podcastdirectory.py
```diff
@@ -318,3 +319,228 @@ def on_btnOK_clicked(self, widget, *args):
...
+        # cell.set_property('ellipsize', Pango.EllipsizeMode.END)
+        column.pack_start(cell, True)
+        column.add_attribute(cell, 'text', DirectoryProvidersModel.C_TEXT)
+        column.add_attribute(cell, 'weight', DirectoryProvidersModel.C_WEIGHT)
+        self.tv_providers.append_column(column)
+
+        self.tv_providers.set_row_separator_func(self.providers_model.is_row_separator)
+
+        self.tv_providers.set_model(self.providers_model)
+
```
auouymous, line 384: Indented blank line.

share/gpodder/ui/gtk/gpodderbrowser.ui
```diff
@@ -0,0 +1,1014 @@
...
+    <property name="pixbuf">../../images/directory-toplist.png</property>
+  </object>
+  <object class="GtkImage" id="directory-youtube">
+    <property name="visible">True</property>
+    <property name="can-focus">False</property>
+    <property name="pixbuf">../../images/directory-youtube.png</property>
+  </object>
+  <object class="GtkTextBuffer" id="textbuffer1">
+    <property name="text" translatable="yes">Error: AABABBAEffl r;grr
+grglth,th</property>
```
auouymous, line 47: Does this mean something?

src/gpodder/gtkui/desktop/podcastdirectory.py
```diff
@@ -318,3 +319,228 @@ def on_btnOK_clicked(self, widget, *args):
...
+        self.podcasts_model.clear()
+
+        @util.run_in_background
+        def download_data():
+            podcasts = []
+            error = None
+            try:
+                podcasts = callback()
+            except Exception as e:
+                logger.warning('Got exception while loading podcasts1: %r', e, exc_info=True)
```
auouymous, line 478: Is the 1 in `podcasts1` accidental?

auouymous (changes requested):

share/gpodder/extensions/podverse.py
elelay: Sorry for the delayed answer:

here is the flow I implemented:
1. select a provider list
2. example selecting using podvers search podverse search
3. example selecting using gpodder.net top 50 top50

Alternate to (1.) using buttons (not implemented, just mocked in glade):
buttons

ok to remove tag cloud

Implemented JustAWarning exception, used by providers as you suggested.

I'll keep the simple 3 progress states at the moment, so no more than the first page will be loaded from podverse search results.

Regarding opds (#980), it doesn't have  a ui yet, but what I meant is that the updated UI I'm building right now is still not suited to it.

ok to implement podcast add later in gpo CLI.

Indeed the window is small. It changes size between states, which is not very smooth.

src/gpodder/gtkui/desktop/podcastdirectory.py
elelay: fixed

share/gpodder/ui/gtk/gpodderbrowser.ui
elelay: I put better placeholder text

src/gpodder/gtkui/desktop/podcastdirectory.py
elelay: fixed

src/gpodder/gtkui/desktop/podcastdirectory.py
```diff
@@ -450,7 +450,9 @@ def download_data():
             try:
                 podcasts = callback()
             except Exception as e:
-                logger.warning('Got exception while loading podcasts1: %r', e, exc_info=True)
+                logger.warning(
```
auouymous, line 453: Is there a reason the warnings should be logged to file and console?

auouymous (changes requested):

share/gpodder/extensions/podverse.py
auouymous: Either listing style is fine, gpodder uses buttons on the first window a new user sees, and it uses a list for channels, episodes and downloads.

Allowing the user to resize the window is nice but it should save that size or just use a larger initial size so resizing isn't needed in most cases.

There is a bug returning to the provider list. Click podverse and do a search, then go back to the providers. The buttons remain along the bottom and I can click select all and add (didn't test what would be added). But clicking podverse again shows an empty query.

src/gpodder/gtkui/desktop/podcastdirectory.py
elelay: no, you're right. I log only when Exception

share/gpodder/ui/gtk/gpodderbrowser.ui
```diff
@@ -2,49 +2,30 @@
...
-  </object>
   <object class="GtkImage" id="directory-youtube">
     <property name="visible">True</property>
     <property name="can-focus">False</property>
     <property name="pixbuf">../../images/directory-youtube.png</property>
   </object>
   <object class="GtkTextBuffer" id="textbuffer1">
-    <property name="text" translatable="yes">Error: AABABBAEffl r;grr
-grglth,th</property>
+    <property name="text" translatable="yes">Error: Unexpected response from soundcloud: {'code': 401, 'message': '', 'link': 'https://developers.soundcloud.com/docs/api/explorer/open-api', 'status': '401 - Unauthorized', 'errors': [], 'error': None}
```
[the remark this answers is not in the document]
elelay: I wanted a placeholder for editing in glade. If you prefer it can be removed. I've removed the translatable attribute

elelay:
Moved the new code back to gPodderPodcastDirectory and fixed the Escape key behaviour.

elelay:
I don't know how to make the window big enough that it won't resize...

src/gpodder/gtkui/desktop/podcastdirectory.py
```diff
@@ -239,10 +239,12 @@ def download_data():
             error = None
             try:
                 podcasts = callback()
+            except directory.JustAWarning:
```
auouymous, line 244: Needs `... as e:`.

share/gpodder/ui/gtk/gpodderpodcastdirectory.ui
```diff
@@ -1,258 +1,417 @@
...
+                      </object>
+                      <packing>
+                        <property name="expand">True</property>
+                        <property name="fill">True</property>
+                        <property name="position">3</property>
+                      </packing>
+                    </child>
                   </object>
+                  <packing>
+                    <property name="name">page4</property>
```
auouymous, line 244: What are the page1, page2, page3 and page4 translatable names for?

In `lb_search` there is a translatable "label", is that shown in the UI?

auouymous (changes requested):
LGTM, other than those issues. Could you rebase to pull in my commit that fixes linting?

src/gpodder/gtkui/desktop/podcastdirectory.py
elelay: you're right

share/gpodder/ui/gtk/gpodderpodcastdirectory.ui
elelay: the title would appear in a gtk stack switcher if we used one. I've put some informative text because it might help screen readers.

elelay:
Rebased and your remarks taken into account.
Thanks for catching my errors :-)

auouymous (approved):

auouymous:
Do you want to clean up the patches or just squash them?

elelay:
Good idea: I refactored the history in 2 commits: podcastdirectory work and podsearch.
Thanks

PR #1618, opened 2024-04-04
Add view menu podcast and episode list checkboxes to preferences

tpikonen:
The View menu is pretty crowded with preference-like check boxes which control the podcast and episode list appearance. Maybe they could be added, or even moved completely to the preferences UI?

I made this patch to the adaptive version, which had no way to change these, except the config var editor. The last two episode list check boxes from the View menu are missing, because they are not applicable for the adaptive UI.

Things to be decided:

 * Should these options be moved from the View menu to Preferences, or is duplication OK?
 * If these are moved, then the `Ctrl+D` keyboard shortcut for `Episode descriptions` probably needs to be removed.
 * Should the `Right align...` and `Require control click...` items also be included?

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -324,6 +324,115 @@
                           </packing>
                         </child>
                         <child>
+                          <object class="GtkBox" id="vbox_view">
+                            <property name="visible">True</property>
+                            <property name="can-focus">False</property>
+                            <property name="border-width">12</property>
+                            <property name="orientation">vertical</property>
+                            <property name="spacing">12</property>
+                            <child>
+                              <object class="GtkCheckButton" id="checkbutton_podcast_list_hide_empty">
+                                <property name="label" translatable="yes">Hide podcasts without episodes</property>
+                                <property name="visible">True</property>
+                                <property name="can-focus">True</property>
+                                <property name="receives-default">False</property>
+                                <property name="draw-indicator">True</property>
+                              </object>
+                              <packing>
+                                <property name="expand">False</property>
+                                <property name="fill">True</property>
+                                <property name="position">0</property>
+                              </packing>
+                            </child>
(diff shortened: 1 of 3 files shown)


auouymous:
Duplication will lead to future changes only being made to one area. I wouldn't have a problem moving most of the checkboxes and columns menu to preferences. Would it be possible to programmatically add the others to both?

> If these are moved, then the Ctrl+D keyboard shortcut for Episode descriptions probably needs to be removed.

I have ctrl+D bound to download in the episode menu, so I wouldn't have a problem removing it. I don't know if anyone else would though.

I have ctrl+- bound to `Hide podcasts without episodes` and use it along with the radio button binds all the time.

If we can programmatically share items between the menu and preferences, we could add back any items users rely on daily.

tpikonen:
> Would it be possible to programmatically add the others to both?

What would this mean in practice? I think menu items can be hidden on runtime, so technically this would be possible. Should this be controlled by a configuration variable or something else?

auouymous:
Injecting the widgets in prefs or menuitems in view menu using a single function instead of duplicating them across two glade files. Similar to the `insert_menuitem` functions in `gtkui/main.py`, but it would have a boolean parameter to select the output.

I'm not sure how well that would work since the items wouldn't all be grouped together. So perhaps just put a comment on each shared item in both glade files indicating they are shared between prefs and view menu.

tpikonen:
I seem to have forgotten this. Now rebased and I

> put a comment on each shared item in both glade files indicating they are shared between prefs and view menu.

auouymous:
It might be better if these were in a View tab of preferences, to avoid cluttering the general tab and it gives them a section name.

Are none of the other settings needed in adaptive? You are free to add any others there.

tpikonen:
I added a 'View' tab to the preferences with the checkboxes.

Adaptive version either hardcodes other view settings or has other controls to change them, so this is all that's needed.

auouymous:
Thank you.

Issue #1471, opened 2023-02-02
How to use extensions


amanofgodforever:
Hi, 

Forgive my complete ignorance on how to use github but I need some training. 

I have installed gpodder and want to normalise audio, but it's telling me 'command not found: normalize-ogg' and I don't know how to find that command or install it. 

I have also noticed that if I set it to rename downloaded episodes then that is fine in their respective folders, but when I select episodes within gpodder and send them to a local folder, the original (not renamed) name is there when they export, and I have to manually rename every episode (the point of setting the renaming extension being to avoid such a tedious task). 

Any advice would be greatly appreciated. 

Thanks.

auouymous:
You need to install http://normalize.nongnu.org/.

I agree that the filename on disk should be used when sending to local folder. However, some feeds use cryptic filenames for episodes, and using the episode title guarantees the resulting file has a readable name, even when not using the rename extension. The best approach might be a new config option to use the filename. Maybe even a text entry setting similar to `device_sync.custom_sync_name` = `{episode.pubdate_prop}_{episode.title}`, allowing customization of the generated filename.

amanofgodforever:
Thanks for the input. Not complaining, but do you know why that is not clear in the gpodder documentation? (about normalize)

How does one go about making that a suggested change to future gpodder versions? (or is there a way for me to change that within my own installation?)

auouymous:
> do you know why that is not clear in the gpodder documentation? (about normalize)

https://gpodder.github.io/docs/extensions/normalizeaudio.html now mentions that normalize-audio is not included.


> How does one go about making that a suggested change to future gpodder versions?

This open issue is the suggested change.


> is there a way for me to change that within my own installation?

You could change the `episode.sync_filename()` on https://github.com/gpodder/gpodder/blob/master/src/gpodder/gtkui/main.py#L1986 to `os.path.splitext(episode.download_filename)[0]`.

PR #1620, opened 2024-04-08
SendTo automatic rename filename

ndias:
This pull request is to allow automatic rename of the file Sent To Folders, issue https://github.com/gpodder/gpodder/issues/1471

### DIFF
--- a/src/gpodder/config.py
+++ b/src/gpodder/config.py
@@ -243,7 +243,7 @@ defaults = {
         'enabled': [],
     },
     'sendto': {
-        'file_format': 'episode.sortdate+" "+episode.title',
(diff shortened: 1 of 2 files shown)


auouymous:
This should simply pass the config to `sync_filename()` along with an enabled config, as is done with `custom_sync_name` and `custom_sync_name_enabled`.

ndias:
> This should simply pass the config to `sync_filename()` along with an enabled config, as is done with `custom_sync_name` and `custom_sync_name_enabled`.

Ok, changes made as requested, enabled option and use of sync_filename function

auouymous:
Thank you for implementing this.

PR #1658, opened 2024-08-26
Fix #1405 Group subscriptions in exported OPML files

### LINKED ISSUE #1405: Group subscriptions in exported OPML files
Exported subscriptions should be grouped according to the user's sections/categories. Example syntax from [Inoreader](https://www.inoreader.com/):

```
    <outline text="Music" title="Music">
      <outline text="Example 1" title="Example 1" type="rss" xmlUrl="https://example.com/feed" htmlUrl="https://example.com/"/>
      <outline text="Example 2" title="Example 2" type="rss" xmlUrl="https://example.net/feed" htmlUrl="https://example.net/"/>
    </outline>
    <outline text="Interviews" title="Interviews">
      <outline text="Example 3" title="Example 3" type="rss" xmlUrl="https://example.org/feed" htmlUrl="https://example.org/"/>
    </outline>
```

elelay:
Fix #1405 Group subscriptions in exported OPML files

also implement sections in OPML import.
Adapted from the same feature by Keeper-of-the-Keys in gpodder-core, thanks!

Now any DirectoryProvider can set the channel['section'] to have it added as is.

### DIFF
--- a/src/gpodder/opml.py
+++ b/src/gpodder/opml.py
@@ -75,7 +75,6 @@ class Importer(object):
             # Make sure we are dealing with a valid link type (ignore case)
             otl_type = outline.getAttribute('type')
             if otl_type is None or otl_type.lower() not in self.VALID_TYPES:
-                breakpoint()
                 otl_title = outline.getAttribute('title')
                 otl_text = outline.getAttribute('text')
                 # gPodder sections will have name == text, if OPML accepts it type=section
@@ -97,7 +96,7 @@ class Importer(object):
                         outline.getAttribute('text')
                         or outline.getAttribute('xmlUrl')
                         or outline.getAttribute('url'),
-                    'section': section or 'audio',
+                    'section': section,
                 }
 
                 if channel['description'] == channel['title']:
--- a/src/gpodder/opml.py
+++ b/src/gpodder/opml.py
@@ -70,10 +70,17 @@ class Importer(object):
         else:
             doc = xml.dom.minidom.parse(io.BytesIO(util.urlopen(url).content))
 
+        section = None
         for outline in doc.getElementsByTagName('outline'):
             # Make sure we are dealing with a valid link type (ignore case)
             otl_type = outline.getAttribute('type')
             if otl_type is None or otl_type.lower() not in self.VALID_TYPES:
+                breakpoint()
+                otl_title = outline.getAttribute('title')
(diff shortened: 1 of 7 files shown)


src/gpodder/opml.py
```diff
@@ -70,10 +70,17 @@ def __init__(self, url):
         else:
             doc = xml.dom.minidom.parse(io.BytesIO(util.urlopen(url).content))
 
+        section = None
         for outline in doc.getElementsByTagName('outline'):
             # Make sure we are dealing with a valid link type (ignore case)
             otl_type = outline.getAttribute('type')
             if otl_type is None or otl_type.lower() not in self.VALID_TYPES:
+                breakpoint()
```
auouymous, line 78: Enters debugger when importing the OPML. ;)

auouymous (changes requested):

src/gpodder/opml.py
```diff
@@ -90,6 +97,7 @@ def __init__(self, url):
                         outline.getAttribute('text')
                         or outline.getAttribute('xmlUrl')
                         or outline.getAttribute('url'),
+                    'section': section or 'audio',
```
auouymous, line 100: Should be "Audio" to match what gpodder already uses. It works great, however, importing an old OPML file places all channels in "audio". Without this patch, it checks the feed type and places them in "Audio" or "Video".

auouymous (changes requested):

auouymous:
Only feature missing is sections in the import window. Simplest would be to show the section name for each channel, and a more complex system would group the channels in sections with the ability to select entire sections. But that could be added later.


> Now any DirectoryProvider can set the channel['section'] to have it added as is.

Added as what?

src/gpodder/opml.py
auouymous: Getting rid of ` or 'audio'` reverts to the old behavior when importing old feeds.

elelay: thanks

src/gpodder/opml.py
elelay: woopsie

elelay:
>>    Now any DirectoryProvider can set the channel['section'] to have it added as is.

> Added as what?

Any DirectoryProvider can set the channel['section'] to have the podcast added into this section (not choosable by the user).

auouymous:
LGTM, are you ready to merge?

PR #1664, opened 2024-10-07
Fix #1661 Interrupt download blocked in Retry-After

### LINKED ISSUE #1661: Impossible to cancel audio download if server send "Retry-After" header
I am running gPodder 3.11.4 for Windows, but the issue is present in all recent versions (possible in all versions at all).

If is requested the download of an audio, but the server send an answer with **Error 503** containing the "**Retry-After**" header, it is impossible to cancel the download until the "Retry-After" expires.

For example, if the server send the "Retry-After: 1800", where 1800 is in seconds, gPodder keeps the download sleeping for 1800 seconds, and even if the download is "cancelled", it remains there sleeping for 1800 seconds and the right-click menu does not allows to cancel again anymore. 

gpodder bug

If gPodder is closed, two gPoder processes are kept running in the Windows. To cancel the download one has to "kill" that two processes.

gpodder bug 2

elelay:
Feed refresh (also when adding the feed to gPodder):
 - only through gtkui
 - only normal feeds (not youtube-dl)

Downloads:
 - better with gtkui, needs to be plugged into gpo
 - normal and youtube-dl work
 - Cancelling during a Retry-After sleep is instaneous for normal downloads
 - youtube-dl downloads are cancelled at the next status update


elelay:
this is pretty much wip. A new API is missing to cancel podcast refresh.

auouymous:
Should it just sleep for a while with no indication as to why it isn't downloading? The user might then cancel and retry, not realizing they need to wait hours before it will work.

Or would it be better to throw an error with a retry-after delay message? I assume you've looked into applying InterruptibleRetry to yt-dlp and had to manually support cancelling for it. And having yt-dlp not retry and manually implementing our own retry would probably not have access to the retry-after delay, so I don't know what to do there other than failing with an error after 3 retries.

elelay:
See #1661 #1675

PR #1666, opened 2024-10-09
run /bin/gpo without install Module PyGObject

MarkusHackspacher:
With this changes could you run python /bin/gpo without install Module PyGObject.

### DIFF
--- a/src/gpodder/deviceplaylist.py
+++ b/src/gpodder/deviceplaylist.py
@@ -25,9 +25,12 @@ from gpodder import util
 from gpodder.sync import (episode_filename_on_device,
                           episode_foldername_on_device)
 
-import gi  # isort:skip
-gi.require_version('Gio', '2.0')  # isort:skip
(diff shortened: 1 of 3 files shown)


auouymous:
`make_directory()` is required to create the initial `gPodder` directory, and a directory each time you subscribe to a channel. Are you running gpo on a system that reads the gPodder directory from another system used to manage the channels? `gpodder-migrate2tres` and `core.py` don't check its return value, which is a bug, and should be exiting if false. Other callers should at least log an error like `model.py` does.

MarkusHackspacher:
I had not understood Gio.File, I close this PR.  
https://lazka.github.io/pgi-docs/#Gio-2.0/classes/File.html#Gio.File
is a high level abstraction for manipulating files on a virtual file system. GFile``s are lightweight, immutable objects that do no I/O upon creation. It is necessary to understand that ``GFile objects do not represent files, merely an identifier for a file.

PR #1668, opened 2024-10-10
Fix an empty `mygpo.device.uid` to avoid mygpod failures.

### LINKED ISSUE #751: "Webservice access not enabled"
I'm trying to migrate from Podcast Addict to AntennaPod and gPodder.

I'm following this [instruction](http://www.mameau.com/gpodder-podcast-import-script/). I get an error in step 4, but that's [another issue](https://github.com/sairuk/gpodder_pc_import/issues/1) and not what I'm addressing here.

So, I've skipped step 4. I now see my subscriptions in gPodder and want to sync them to gpodder.net. As a server, I left "gpodder.net" and entered my username and password as well as a device name. I added the same device also on gpodder.net.

The subscriptions don't automatically sync to gpodder.net. So, I've clicked the button and tried to replace my subscriptions on the server with the local subscriptions. This raises the error:

`Webservice access not enabled.`

I use Windows 10 and gPodder Portable from PortableApps.com

### LINKED ISSUE #1663: gPodder fails to upload and sync subscriptions to gpodder.net
gPodder fails to upload and sync subscriptions to gpodder.net. I have entered my login credentials correctly but every time this error message appears. Do I need to install additional plugins or change some expert setting to make it work or is it really a bug?

grafik

auouymous:
Some platforms could return an empty hostname string. This resulted in mygpo always failing because it depends on the uid to be non-empty.

If hostname is empty, it now returns 'unknown'. It also appends a 32 digit identifier to the end of it to increase uniqueness of the uid.

Old installs are fixed by detecting an empty uid and generated a new value.

Fixes #751.
Fixes #1663.

@elelay Hostnames can be 64 to 255 characters in length, and this now appends 33 additional characters to that. It looks like uid on the server is limited to 50 characters, which doesn't even hold a valid hostname. Should I reduce the random identifier length?

### DIFF
--- a/src/gpodder/my.py
+++ b/src/gpodder/my.py
@@ -202,6 +202,11 @@ class MygPoClient(object):
         self._worker_thread = None
         atexit.register(self._at_exit)
 
+        # Old versions could set uid to an empty string which causes mygpo to always fail
+        if self._config.mygpo.device.uid == '':
+            self._config.mygpo.device.uid = util.get_hostname_uid()
+            logger.warning('mygpo.device.uid was empty, initializing to "%s"' % self._config.mygpo.device.uid)
+
     def create_device(self):
         """Upload the device changes to the server.
 
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -41,6 +41,7 @@ import mimetypes
 import os
 import os.path
 import platform
+import random
 import re
 import shlex
 import shutil
@@ -1777,17 +1778,33 @@ def relpath(p1, p2):
 
 def get_hostname():
     """Return the hostname of this computer.
-
-    This can be implemented in a different way on each
-    platform and should yield a unique-per-user device ID.
     """
-    nodename = platform.node()
 
+    nodename = platform.node()
     if nodename:
         return nodename
 
     # Fallback - but can this give us "localhost"?
-    return socket.gethostname()
+    nodename = socket.gethostname()
+    if nodename:
+        return nodename
+
+    return 'unknown'
+
+
+def get_hostname_uid():
+    """Return the hostname of this computer and a random identifier.
+
+    This can be implemented in a different way on each
+    platform and should yield a unique-per-user device ID.
+    """
+
+    # Generate a 32-bit random identifier
+    random.seed()
+    uid = '%X' % random.getrandbits(32)
+
+    hostname = get_hostname()
+    return f'{hostname}-{uid}'
 
 
 def detect_device_type():
--- a/src/gpodder/util.py
+++ b/src/gpodder/util.py
@@ -1803,7 +1803,9 @@ def get_hostname_uid():
     random.seed()
     uid = '%X' % random.getrandbits(32)
 
-    hostname = get_hostname()
+    # conform to mygpo device uid limit
+    # (https://github.com/gpodder/mygpo/blob/master/mygpo/users/models.py#L235)
+    hostname = get_hostname()[:64 - len(uid) - 1]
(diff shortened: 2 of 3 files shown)


elelay:
I don't know but it seems uid is indeed [max 64 chars](https://github.com/gpodder/mygpo/blob/master/mygpo/users/models.py#L235)
Please why append a uuid when hostname is known? Is it to support multiple installs per host?

auouymous:
The comment mentioned implementing a unique per-user ID. There is nothing stopping a user from having the same hostname on multiple computers, so the ID is technically always required for a unique uid.

After more thought, a 32bit ID "should" provide enough randomness to never produce the same value for a user with multiple machines, all with the same hostname. That would reduce it to 9 additional characters, reducing the chance of the uid overflowing the server's 50 char limit.

elelay:
sounds good!
where does the 50 char limit come from? It's [64 chars](https://github.com/gpodder/mygpo/blob/master/mygpo/users/models.py#L235), right?

auouymous:
I assume https://github.com/gpodder/mygpo/blob/80c41dc0c9a58dc0e85f6ef56662cdfd0d6e3b16/mygpo/web/forms.py#L105 is the same uid.

auouymous:
I also changed `-%08X` to `-%X` which will reduce the size below 9 extra digits on ~50% of installs.

src/gpodder/my.py
```diff
@@ -202,6 +202,11 @@ def __init__(self, config):
         self._worker_thread = None
         atexit.register(self._at_exit)
 
+        # Old versions could set uid to an empty string which causes mygpo to always fail
+        if self._config.mygpo.device.uid == '':
+            self._config.mygpo.device.uid = util.get_hostname_uid()
```
elelay, line 207:
```suggestion
            # conform to mygpo device uid limit
            # (https://github.com/gpodder/mygpo/blob/master/mygpo/users/models.py#L235)
            self._config.mygpo.device.uid = util.get_hostname_uid()[:64]
```

elelay (changes requested):
LGTM, with the truncation to avoid issues with mygpo

src/gpodder/my.py
auouymous: Should we instead trim the hostname inside `get_hostname_uid()` to `63-len(uid)`? That way we don't trim off the random uid and get long hostname that might be the same across all of a user's devices.

elelay: yes

src/gpodder/util.py
```diff
@@ -1777,17 +1778,33 @@ def relpath(p1, p2):
...
+
+    This can be implemented in a different way on each
+    platform and should yield a unique-per-user device ID.
+    """
+
+    # Generate a 32-bit random identifier
+    random.seed()
+    uid = '%X' % random.getrandbits(32)
+
+    hostname = get_hostname()
```
elelay, line 1806:
```suggestion
    # conform to mygpo device uid limit
    # (https://github.com/gpodder/mygpo/blob/master/mygpo/users/models.py#L235)
    hostname = get_hostname()[:64 - len(uid) - 1]
```

elelay (approved):

auouymous:
I just noticed that https://github.com/gpodder/mygpo/blob/80c41dc0c9a58dc0e85f6ef56662cdfd0d6e3b16/mygpo/web/forms.py#L105 in my comment above was never resolved, is it being limited to 50 there?

elelay:
yes, the web forms is more restrictive than the db. I don't know why

PR #1672, opened 2024-11-18
rockbox_coverart: convert, resize, and make baseline images

dconrad5:
Extend the rockbox_coverart extension to:
 - Convert jpeg files to baseline
 - Resize files to (user-defined) square size, default 500px
 - Optionally allow upsizing of files
 - Convert to (user-defined) image type, default jpeg
 - Allow falling back to original behavior if desired
 - Lock file so we don't crash when running many at once
 - Try a few different source file names, couldn't figure
   out how to get source cover art name from gpodder

Extension now requires python modules wand, filelock. Both
are available on pip.

Also make filename sanitizing more strict.


dconrad5:
Hmm. This still has issues with allowing multiple downloads at one time by doing the following:
  1. Set "limit downloads to" to more than 1
  2. Sync to device
  3. Quit gPodder
  4. Open gPodder back up
  5. Sync to device again
  6. gPodder will crash (segmentation fault)

dconrad5:
Ok, with this newest push I've converted from using image magick / Wand to using pillow (PIL), and we can now successfully check the details of the on-device files without crashing, even with unlimited parallel downloads (tested with a set of 10 podcasts).

dconrad5:
I believe this PR is ready for review, it's working well for me.

share/gpodder/extensions/rockbox_coverart.py
```diff
@@ -37,10 +45,101 @@ def on_episode_synced(self, device, episode):
...
+                    and device_filetype.lower() != "png":
+                device_filetype = "jpg"
+
+            # sanitize for filetype checking - "jpg" will not match "jpeg"
+            if device_filetype.upper() == "JPG":
+                device_match_filetype = "JPEG"
+            else:
+                device_match_filetype = device_filetype.upper()
+
+            # barring any way for gpodder core to tell us the source cover art name,
```
elelay, line 64: I'd rather add a method to get artwork location from the model

src/gpodder/deviceplaylist.py
```diff
@@ -104,7 +105,15 @@ def get_absolute_filename_for_playlist(self, episode):
         if foldername:
             filename = os.path.join(foldername, filename)
         if self._config.device_sync.playlists.use_absolute_path:
-            filename = os.path.join(util.relpath(self._config.device_sync.device_folder, self.mountpoint.get_uri()), filename)
+            # find mount point, ensuring we don't end up locked up in the loop
+            drive_start = self._config.device_sync.device_folder
+            while os.path.ismount(drive_start) is not True and drive_start != os.path.dirname(drive_start):
```
elelay, line 110:
```suggestion
            while not os.path.ismount(drive_start) and drive_start != os.path.dirname(drive_start):
```

src/gpodder/deviceplaylist.py
```diff
@@ -104,7 +105,15 @@ def get_absolute_filename_for_playlist(self, episode):
...
+                drive_start = os.path.dirname(drive_start)
+            filename = os.path.join(util.relpath(self._config.device_sync.device_folder, drive_start), filename)
+            # distinguish this as an absolute path relative to the device
+            filename = "/" + filename
```
elelay, line 114: is the rest of the code supposed to work with url as a path (eg `mtp:///`)?
Then prepending a slash will break it: `/mtp:///`

elelay (changes requested):
Thanks for contributing.
It's nice that you actually use device sync and adjusted it for a working workflow.
Please see my comments for details.

src/gpodder/deviceplaylist.py
dconrad5: The intention is to not have a prefix like mtp:// or file:// at all for the filename.

I actually had to adjust things because self.mountpoint.get_uri() was giving me "file://path/to/somewhere". I _think_ this will tend to give just a regular file path, but I'm not knowledgeable enough to know if it can be guaranteed or not with the way it is right now. It does work for me as is right now, at least.

share/gpodder/extensions/rockbox_coverart.py
dconrad5: Hmm, I'll try to figure out how to do that then.

src/gpodder/deviceplaylist.py
dconrad5: Perhaps I should check if the filename is a url or a regular file path and try to strip any prefix we find? I think that should be do-able.

share/gpodder/extensions/rockbox_coverart.py
dconrad5: Seems like the solution was under my nose the whole time, we can get the cover path and filename from CoverDownloader(), provided we know most of it, minus extension. Seems to work pretty well.

auouymous:
> Extension now requires python modules pillow (PIL), filelock. Both are available on pip.

The dependencies also need to be added to the Mac build.

src/gpodder/deviceplaylist.py
dconrad5: I believe this is sufficiently handled with the newest commit, [f475386](https://github.com/gpodder/gpodder/pull/1672/commits/f47538601754299b3bb8c308c75bac81f0491789). Purposefully strip file: and mtp: prefixes for m3u entry paths.

dconrad5:
> > Extension now requires python modules pillow (PIL), filelock. Both are available on pip.
> 
> The dependencies also need to be added to the Mac build.

I thought I had added them with [cfd555c](https://github.com/gpodder/gpodder/pull/1672/commits/cfd555c7c06f3747f6b39678c0e354ccb68b768c), unless there's more to do so? Or something I don't have access to?

dconrad5:
> > Extension now requires python modules pillow (PIL), filelock. Both are available on pip.
> 
> The dependencies also need to be added to the Mac build.

I believe I should have that covered now with [a0a2e50](https://github.com/gpodder/gpodder/pull/1672/commits/a0a2e504afe09e1bc1b9cb5b58401e8273543661)

Also added [a4f9f89](https://github.com/gpodder/gpodder/pull/1672/commits/a4f9f8907d345db3e9f689f312c09a5e4c0e2fa8), but I'm not actually sure if it's necessary.

dconrad5:
Rebased this change set on top of current master

dconrad5:
Added preferences tabs for rockbox_coverart.py and tagging.py. Please take a look, they are working well for me.

dconrad5:
I'm working on adding smarter cover art processing to the mutagen tagging extension. It's working really well for mp3, but I have not extended it to the other two file types the extension supports, ogg and mp4. However, if this pull request is getting close to the point it can be merged, I will hold off on merging that stuff into this pull request and make that into a separate pull request.

Is there a desire to get this pull request across the finish line? I would appreciate if someone would take another look and tell me what I need to do to accomplish that. Or if the answer is "we are not going to merge this".

dconrad5:
Ok, added whitespace fixes according to linter results. Hopefully linter comes back clean now.

dconrad5:
Sorted imports according to isort now.

dconrad5:
Ok, all checks passed, I'll stop touching this now.

elelay:
Thank you. I'll review when I have time for the project

share/gpodder/extensions/rockbox_coverart.py
```diff
@@ -37,10 +52,194 @@ def on_episode_synced(self, device, episode):
...
+                                        out = img.resize((int(self.config.convert_size), int(self.config.convert_size)))
+                                    else:
+                                        out = img.copy()
+                                    out.save(device_art)
+                            except OSError:
+                                logger.info("%s image error!", episode_art)
+                        else:
+                            logger.info("%s %s" % (device_art, "already exists"))
+
+                        device_lockfile.release()
```
elelay, line 132: I'm not super comfortable with `device_lockfile` not used [as a context manager or with a try catch](https://py-filelock.readthedocs.io/en/latest/api.html#filelock.BaseAsyncFileLock.acquire).
Let's look at the code once more after other comments are addressed.

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
 
         if modify_tags:
             if remove_before_modify:
+                logger.info("removing before writing")
```
elelay, line 110: are all these `logger.info` needed? I would use `logger.debug`

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
                 audio.delete()
 
-            if self.album is not None:
+            if write_album is True and self.album is not None:
```
elelay, line 113:
```suggestion
            if write_album and self.album is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
+                logger.info("writing album")
                 audio.tags['album'] = self.album
 
-            if self.title is not None:
+            if write_title is True and self.title is not None:
```
elelay, line 117:
```suggestion
            if write_title and self.title is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
+                logger.info("writing title")
                 audio.tags['title'] = self.title
 
-            if self.subtitle is not None:
+            if write_subtitle is True and self.subtitle is not None:
```
elelay, line 121:
```suggestion
            if write_subtitle and self.subtitle is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
+                logger.info("writing subtitle")
                 audio.tags['subtitle'] = self.subtitle
 
-            if self.subtitle is not None:
+            if write_comments is True and self.subtitle is not None:
```
elelay, line 125:
```suggestion
            if write_comments and self.subtitle is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
+                logger.info("writing comments")
                 audio.tags['comments'] = self.subtitle
 
-            if self.genre is not None:
+            if write_genre is True and self.genre is not None:
```
elelay, line 129:
```suggestion
            if write_genre and self.genre is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -95,27 +107,35 @@ def write_basic_tags(self, remove_before_modify, modify_tags, set_artist_to_albu
...
+                logger.info("writing genre")
                 audio.tags['genre'] = self.genre
 
-            if self.pubDate is not None:
+            if write_pubdate is True and self.pubDate is not None:
```
elelay, line 133:
```suggestion
            if write_pubdate and self.pubDate is not None:
```

share/gpodder/extensions/tagging.py
```diff
@@ -302,3 +328,194 @@ def get_cover(self, podcast):
         downloader = coverart.CoverDownloader()
         return downloader.get_cover(podcast.cover_file, podcast.cover_url,
             podcast.url, podcast.title, None, None, True)
+
+    def toggle_sensitivity_of_widgets(self):
+        if not self.container.config.always_remove_tags:
```
elelay, line 333: nitpick: I'd rather inverse the if clauses, to avoid `if not ... else:`

elelay (changes requested):
a few codestyle changes to make, but it looks good

elelay:
There is an edge case with device sync when using absolute path

    mkdir -p /tmp/gposync/Playlists

- choose `/tmp/gposync` as the root of the filesystem device from Preferences.
- keep the default 'Playlists' value for playlist folder
- the setting `device_sync.playlist.user_absolute_path` is `True`
- on my computer `/tmp` is a separate filesystem (a tmpfs), so `os.path.ismount('/tmp')` returns True.

I downloaded an episode of the Moth as an example, and synced to device. Files are at the right place, but the path in the playlist is wrong:
````
#EXTM3U
#EXTINF:0,The Moth Podcast: Resolutions
/gposync/The Moth/The Moth Podcast_ Resolutions.mp3
````

With master, content is:
````
gPodder> cat "/tmp/gposync/Playlists/The Moth.m3u"
#EXTM3U
#EXTINF:0,The Moth Podcast_ Resolutions
../The Moth/The Moth Podcast_ Resolutions.mp3
````

If using an absolute path, I think the path should be `/The Moth/The Moth Podcast_ Resolutions.mp3`, right?

**Comment by elelay** [comment]
To address the edgecase, I propose to change the fallback code in deviceplaylist.py constructor:

```
        try:
            self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()
        except GLib.Error as err:
            logger.error('find_enclosing_mount folder %s failed: %s', self.playlist_folder.get_uri(), err.message)
        # changed fallback code
        if not self.mountpoint:
            # look for the common path of device folder and playlists
            # (not using os.path.commonpath because we want a gio.File)
            # must work with:
            # - Playlist path a parent of the device folder (playlists at device root)
            # - Playlist path a child of the device folder (Playlists folder next to each podcast folder)
            # - Playlist path next to or in a subfolder next to the device folder
            mountpoint = device_folder
            # we constructed the path to playlist using util.relpath, so it must be / separated
            to_parent_path_components = self._config.device_sync.playlists.folder.split('/')
            while mountpoint and to_parent_path_components and to_parent_path_components[0] == '..':
                # path starts with .. : move up once
                mountpoint = mountpoint.get_parent()
                to_parent_path_components.pop(0)
            if mountpoint and (self.playlist_folder.has_prefix(mountpoint) or self.playlist_folder.equal(mountpoint)):
               self.mountpoint = mountpoint
            else:
                 raise Exception(
                     "Unable to find the device root path from %s and playlist %s"
                     % (self._config.device_sync.device_folder, self._config.device_sync.playlists.folder))
            logger.warning('could not find mount point for MP3 player - using %s as MP3 player root', self.mountpoint.get_uri())
```

Then `get_path_to_filename_for_playlist` is simplified into:
```
    def get_path_to_filename_for_playlist(self, episode):
        """
        get the filename including full path for the given episode for the playlist
        """
        filename = self.get_filename_for_playlist(episode)
        foldername = episode_foldername_on_device(self._config, episode)
        if foldername:
            filename = os.path.join(foldername, filename)
        if self._config.device_sync.playlists.use_absolute_path:
            device_folder = util.new_gio_file(self._config.device_sync.device_folder)
            file_ = device_folder.resolve_relative_path(filename)
            filename = "/" + util.relpath(file_.get_path(), self.mountpoint.get_path())
        else:
            filename = os.path.join(self.playlist_to_device_relpath, filename)
        return filename
```

what do you think?

File: share/gpodder/extensions/rename_download.py [hunk 13]
```diff
@@ -107,7 +107,9 @@ def make_filename(self, current_filename, title, sortdate, podcast_title):
             new_basename,
             ext,
             PodcastEpisode.MAX_FILENAME_LENGTH,
-            PodcastEpisode.MAX_FILENAME_WITH_EXT_LENGTH)
+            PodcastEpisode.MAX_FILENAME_WITH_EXT_LENGTH,
+            self.config.sanitize_filenames.remove_more_punctuation,
```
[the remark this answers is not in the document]
dconrad5: Ah, I meant to remove those entirely. I ended up reverting https://github.com/gpodder/gpodder/pull/1672/commits/4d78ccc201eada7f0564662be0ffc6aeccdb4321 and https://github.com/gpodder/gpodder/pull/1672/commits/4086ee0161c9f6e279f10e4263ca2f685b218553 entirely in PR #1715. Should I try to cherry-pick the revert commits into this branch? I think I should be able to do that.

src/gpodder/util.py
```diff
@@ -1632,7 +1632,8 @@ def sanitize_filename(filename, max_length):
         filename = filename.decode('utf-8')
 
     # see #361 - at least slash must be removed
-    filename = re.sub(r"[\"*/:<>?\\|]", "_", filename)
+    # add additional common punctuation
```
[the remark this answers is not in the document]
elelay: please cherry-pick them, yes :-)

elelay:
I've just tested your MR + the changes I wrote above (also visible as commit dcd3580).
It works.

In both cases I create and select a *gPodder* folder inside the tablet to store the episodes, with a Playlists subfolder.


# Linux test

## mtp
Tested android tablet in mtp mode using [go-mtpfs](https://wiki.archlinux.org/title/Media_Transfer_Protocol#go-mtpfs):
the storage appears via file:///path/to/chosen/mountpoint.

It is not deemed an "interesting" mountpoint by gio so the same fallback code is used in `deviceplaylist.py`.

## usb mass storage
Tested android tablet in mass storage mode:
it appears on the left of the file picker dialog and the storage uri is file:///run/media/elelay/0000-006F.

It is an "interesting" mountpoint for gio so `find_enclosing_mount().get_root()` actually works.
Absolute paths work in playlist:
```
cat /run/media/elelay/0000-006F/gPodder/Playlists/The\ Moth\ Podcast.m3u
#EXTM3U
#EXTINF:0,The Moth Podcast: Gifts and Gratitude
/gPodder/The Moth Podcast/The Moth Podcast_ Gifts and Gratitude.mp3
```

# Windows

## mtp
the mountpoint is not visible in gtk file browser so can't be selected.

## mass storage
the folder is saved as `file:///E:/gPodder` in the settings
I can select it from the left panel in the gtk file browser but  `find_enclosing_mount().get_root()` fails.
Indeed there is nothing in `Gio.GVolumeMonitor.get().get_connected_drives()` or `Gio.GVolumeMonitor.get().get_mounts()`

`mountpoint.get_path()` returns `E:\gPodder` with a backslash. The uri has forward slashes `file:///E:/gPodder`

- absolute path in the m3u: `/The Moth Podcast\filename.mp3` (a mix of forward and back slashes)
- relative path in the m3u: ` ..\The Moth Podcast\filename.mp3`  (all backslashes)


The android audio player finds both files wether with forward or back slashes so I would keep the patch as is.

dconrad5:
> I've just tested your MR + the changes I wrote above (also visible as commit [dcd3580](https://github.com/gpodder/gpodder/commit/dcd3580261680c9fe48a59c0e25c72236741e032)). It works.
> 
> In both cases I create and select a _gPodder_ folder inside the tablet to store the episodes, with a Playlists subfolder.
> 
> # Linux test
> ## mtp
> Tested android tablet in mtp mode using [go-mtpfs](https://wiki.archlinux.org/title/Media_Transfer_Protocol#go-mtpfs): the storage appears via file:///path/to/chosen/mountpoint.
> 
> It is not deemed an "interesting" mountpoint by gio so the same fallback code is used in `deviceplaylist.py`.
> 
> ## usb mass storage
> Tested android tablet in mass storage mode: it appears on the left of the file picker dialog and the storage uri is file:///run/media/elelay/0000-006F.
> 
> It is an "interesting" mountpoint for gio so `find_enclosing_mount().get_root()` actually works. Absolute paths work in playlist:
> 
> ```
> cat /run/media/elelay/0000-006F/gPodder/Playlists/The\ Moth\ Podcast.m3u
> #EXTM3U
> #EXTINF:0,The Moth Podcast: Gifts and Gratitude
> /gPodder/The Moth Podcast/The Moth Podcast_ Gifts and Gratitude.mp3
> ```
> 
> # Windows
> ## mtp
> the mountpoint is not visible in gtk file browser so can't be selected.
> 
> ## mass storage
> the folder is saved as `file:///E:/gPodder` in the settings I can select it from the left panel in the gtk file browser but `find_enclosing_mount().get_root()` fails. Indeed there is nothing in `Gio.GVolumeMonitor.get().get_connected_drives()` or `Gio.GVolumeMonitor.get().get_mounts()`
> 
> `mountpoint.get_path()` returns `E:\gPodder` with a backslash. The uri has forward slashes `file:///E:/gPodder`
> 
> * absolute path in the m3u: `/The Moth Podcast\filename.mp3` (a mix of forward and back slashes)
> * relative path in the m3u: ` ..\The Moth Podcast\filename.mp3`  (all backslashes)
> 
> The android audio player finds both files wether with forward or back slashes so I would keep the patch as is.

Excellent! I should be able to work through your comments tonight.

share/gpodder/extensions/rename_download.py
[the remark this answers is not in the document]
dconrad5: Ok, should now be reverted in this PR with [1ddc573](https://github.com/gpodder/gpodder/pull/1672/commits/1ddc573f8d50e2d646e53e6947ee8b2d45012c5f) and [2159402](https://github.com/gpodder/gpodder/pull/1672/commits/21594028f0f218a5cd8b736d8322170ace3dc7e8)

src/gpodder/util.py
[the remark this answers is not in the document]
dconrad5: Ok, should now be reverted in this PR with [1ddc573](https://github.com/gpodder/gpodder/pull/1672/commits/1ddc573f8d50e2d646e53e6947ee8b2d45012c5f) and [2159402](https://github.com/gpodder/gpodder/pull/1672/commits/21594028f0f218a5cd8b736d8322170ace3dc7e8)

share/gpodder/extensions/rockbox_coverart.py
```diff
@@ -37,10 +52,194 @@ def on_episode_synced(self, device, episode):
...
+                                        copyflag = True
+                                    elif img.format.upper() != device_match_filetype:
+                                        copyflag = True
+                                    try:
+                                        if img.info['progressive'] == 1:
+                                            copyflag = True
+                                    except:
+                                        pass  # expected result if baseline jpeg or png
+                            except OSError:
+                                logger.info("%s check image error!", device_art)
```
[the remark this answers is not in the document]
dconrad5: I want to say we probably want to set copyflag false on error here, either convert_size couldn't be converted to int, some file access occurred, or some bug with PIL. I think any of those will probably mean we'll have issues if we try to copy it over to the device. We can definitely add the error to the logger printout though.

share/gpodder/extensions/rockbox_coverart.py
```diff
@@ -37,10 +52,194 @@ def on_episode_synced(self, device, episode):
...
+                            copyflag = True
+
+                        if copyflag is True:
+                            logger.info("%s %s" % (device_art, "copying"))
+                            try:
+                                # should we file lock the source file?
+                                with Image.open(episode_art) as img:
+                                    if img.height > int(self.config.convert_size)\
+                                            or self.config.convert_allow_upscale_art is True:
+                                        out = img.resize((int(self.config.convert_size), int(self.config.convert_size)))
```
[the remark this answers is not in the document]
dconrad5: That's a good question. In my experience non-square art is rather rare, but maybe it would make sense to scale it to the larger side be equal to convert_size if there is one. There might be a built-in PIL option for that, I'll check.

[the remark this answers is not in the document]
dconrad5: Looks like it should be pretty easy to maintain ratio, so I'll get that in there.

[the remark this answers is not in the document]
dconrad5: Please see [2a18d85](https://github.com/gpodder/gpodder/pull/1672/commits/2a18d850b4420861c95e59a76e6c4cfdfd1df96f)

share/gpodder/extensions/tagging.py
dconrad5: I can definitely change all of them, yeah.

dconrad5: Please see commit [4b9508b](https://github.com/gpodder/gpodder/pull/1672/commits/4b9508bee0e6a666e475fe868ba976a714990581)

share/gpodder/extensions/tagging.py
dconrad5: Fair enough, will do.

dconrad5: Please see commit [1df7a95](https://github.com/gpodder/gpodder/pull/1672/commits/1df7a95d8dc578446df6b25d462d7346e9b0e1d8)

share/gpodder/extensions/rockbox_coverart.py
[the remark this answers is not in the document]
dconrad5: Added [4d42d47](https://github.com/gpodder/gpodder/pull/1672/commits/4d42d4718e88e264a670b1c665c63410fd92a7d2) to explicitly set copyflag false here.

dconrad5:
> There is an edge case with device sync when using absolute path
> 
> ```
> mkdir -p /tmp/gposync/Playlists
> ```
> 
>     * choose `/tmp/gposync` as the root of the filesystem device from Preferences.
> 
>     * keep the default 'Playlists' value for playlist folder
> 
>     * the setting `device_sync.playlist.user_absolute_path` is `True`
> 
>     * on my computer `/tmp` is a separate filesystem (a tmpfs), so `os.path.ismount('/tmp')` returns True.
> 
> 
> I downloaded an episode of the Moth as an example, and synced to device. Files are at the right place, but the path in the playlist is wrong:
> 
> ```
> #EXTM3U
> #EXTINF:0,The Moth Podcast: Resolutions
> /gposync/The Moth/The Moth Podcast_ Resolutions.mp3
> ```
> 
> With master, content is:
> 
> ```
> gPodder> cat "/tmp/gposync/Playlists/The Moth.m3u"
> #EXTM3U
> #EXTINF:0,The Moth Podcast_ Resolutions
> ../The Moth/The Moth Podcast_ Resolutions.mp3
> ```
> 
> If using an absolute path, I think the path should be `/The Moth/The Moth Podcast_ Resolutions.mp3`, right?

We don't have a path_to_podcasts_folder (or such) setting, to my knowledge. I guess I've always used the device_sync.device_folder as the actual podcasts folder setting. In the preferences window, it's labelled slightly differently, suggesting it should be the _mount point_ of the device. So maybe that could cause some confusion.

But my usage of it has always been that I want to point it to a podcasts folder on my device - I'll usually have separate `/music` and `/podcasts` folders on my device.

So tf /tmp is considered a removable device, and given the above assumptions about how it _should_ work, I would think your example absolute path of '/gposync/The Moth/The Moth Podcast_ Resolutions.mp3' is correct.

dconrad5:
> To address the edgecase, I propose to change the fallback code in deviceplaylist.py constructor:
> 
> ```
>         try:
>             self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()
>         except GLib.Error as err:
>             logger.error('find_enclosing_mount folder %s failed: %s', self.playlist_folder.get_uri(), err.message)
>         # changed fallback code
>         if not self.mountpoint:
>             # look for the common path of device folder and playlists
>             # (not using os.path.commonpath because we want a gio.File)
>             # must work with:
>             # - Playlist path a parent of the device folder (playlists at device root)
>             # - Playlist path a child of the device folder (Playlists folder next to each podcast folder)
>             # - Playlist path next to or in a subfolder next to the device folder
>             mountpoint = device_folder
>             # we constructed the path to playlist using util.relpath, so it must be / separated
>             to_parent_path_components = self._config.device_sync.playlists.folder.split('/')
>             while mountpoint and to_parent_path_components and to_parent_path_components[0] == '..':
>                 # path starts with .. : move up once
>                 mountpoint = mountpoint.get_parent()
>                 to_parent_path_components.pop(0)
>             if mountpoint and (self.playlist_folder.has_prefix(mountpoint) or self.playlist_folder.equal(mountpoint)):
>                self.mountpoint = mountpoint
>             else:
>                  raise Exception(
>                      "Unable to find the device root path from %s and playlist %s"
>                      % (self._config.device_sync.device_folder, self._config.device_sync.playlists.folder))
>             logger.warning('could not find mount point for MP3 player - using %s as MP3 player root', self.mountpoint.get_uri())
> ```
> 
> Then `get_path_to_filename_for_playlist` is simplified into:
> 
> ```
>     def get_path_to_filename_for_playlist(self, episode):
>         """
>         get the filename including full path for the given episode for the playlist
>         """
>         filename = self.get_filename_for_playlist(episode)
>         foldername = episode_foldername_on_device(self._config, episode)
>         if foldername:
>             filename = os.path.join(foldername, filename)
>         if self._config.device_sync.playlists.use_absolute_path:
>             device_folder = util.new_gio_file(self._config.device_sync.device_folder)
>             file_ = device_folder.resolve_relative_path(filename)
>             filename = "/" + util.relpath(file_.get_path(), self.mountpoint.get_path())
>         else:
>             filename = os.path.join(self.playlist_to_device_relpath, filename)
>         return filename
> ```
> 
> what do you think?

I'm running out of time tonight so I might not get back to it until Wednesday, but I'll take a look and see if I can parse it out. Sounds like it's largely working as intended for you though, which is good!

dconrad5:
> To address the edgecase, I propose to change the fallback code in deviceplaylist.py constructor:
> 
> ```
>         try:
>             self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()
>         except GLib.Error as err:
>             logger.error('find_enclosing_mount folder %s failed: %s', self.playlist_folder.get_uri(), err.message)
>         # changed fallback code
>         if not self.mountpoint:
>             # look for the common path of device folder and playlists
>             # (not using os.path.commonpath because we want a gio.File)
>             # must work with:
>             # - Playlist path a parent of the device folder (playlists at device root)
>             # - Playlist path a child of the device folder (Playlists folder next to each podcast folder)
>             # - Playlist path next to or in a subfolder next to the device folder
>             mountpoint = device_folder
>             # we constructed the path to playlist using util.relpath, so it must be / separated
>             to_parent_path_components = self._config.device_sync.playlists.folder.split('/')
>             while mountpoint and to_parent_path_components and to_parent_path_components[0] == '..':
>                 # path starts with .. : move up once
>                 mountpoint = mountpoint.get_parent()
>                 to_parent_path_components.pop(0)
>             if mountpoint and (self.playlist_folder.has_prefix(mountpoint) or self.playlist_folder.equal(mountpoint)):
>                self.mountpoint = mountpoint
>             else:
>                  raise Exception(
>                      "Unable to find the device root path from %s and playlist %s"
>                      % (self._config.device_sync.device_folder, self._config.device_sync.playlists.folder))
>             logger.warning('could not find mount point for MP3 player - using %s as MP3 player root', self.mountpoint.get_uri())
> ```
> 
> Then `get_path_to_filename_for_playlist` is simplified into:
> 
> ```
>     def get_path_to_filename_for_playlist(self, episode):
>         """
>         get the filename including full path for the given episode for the playlist
>         """
>         filename = self.get_filename_for_playlist(episode)
>         foldername = episode_foldername_on_device(self._config, episode)
>         if foldername:
>             filename = os.path.join(foldername, filename)
>         if self._config.device_sync.playlists.use_absolute_path:
>             device_folder = util.new_gio_file(self._config.device_sync.device_folder)
>             file_ = device_folder.resolve_relative_path(filename)
>             filename = "/" + util.relpath(file_.get_path(), self.mountpoint.get_path())
>         else:
>             filename = os.path.join(self.playlist_to_device_relpath, filename)
>         return filename
> ```
> 
> what do you think?

Just tried it, and this does not work correctly for me. I pointed it to a folder on my device ("MAXMUZAK"):
`device_sync.device_folder = /Volumes/MAXMUZAK/test%20files/pods`

With the playlist folder located at:
`device_sync.playlists.folder = ../playlists`

And, of course, `device_sync.playlists.use_absolute_path = true`.

I would expect an absolute path to look like the following:
`/test%20files/pods/My Brother, My Brother And Me/MBMBaM 748_ Gooped and Gagged, Dad.mp3`

But instead I get:
`/pods/My Brother, My Brother And Me/MBMBaM 748_ Gooped and Gagged, Dad.mp3`.

It would appear that `self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()` does not work correctly for me. I get the following error message and then go to the fallback code you posted:
```
1739406744.544160 [gpodder.deviceplaylist] ERROR: find_enclosing_mount folder file:///Volumes/MAXMUZAK/test%20files/playlists failed: Containing mount for file /Volumes/MAXMUZAK/test files/playlists not found
1739406744.544207 [gpodder.deviceplaylist] WARNING: could not find mount point for MP3 player - using file:///Volumes/MAXMUZAK/test%20files as MP3 player root
```

So my question is, given the above, are we expecting find_enclosing_mount().get_root() to succeed, or are we expecting to use the fallback code? Because if we are expecting find_enclosing_mount().get_root() to succeed, it does not. If we are expecting to need to use the fallback code, it does not appear to work as intended for me.

The fundamental assumption I am using is that the "device_sync.device_folder" setting does not _need_ to be the root of the device, despite how it is labelled in the preferences window. Is this the intended use case? Or should we revamp things so that it is clearly meant to be the root of the device, and then we can add some setting which sets the path from the root of the device to the desired podcasts folder?

elelay:
> > If using an absolute path, I think the path should be `/The Moth/The Moth Podcast_ Resolutions.mp3`, right?
> 
> We don't have a path_to_podcasts_folder (or such) setting, to my knowledge. I guess I've always used the device_sync.device_folder as the actual podcasts folder setting. In the preferences window, it's labelled slightly differently, suggesting it should be the _mount point_ of the device. So maybe that could cause some confusion.
> 
> But my usage of it has always been that I want to point it to a podcasts folder on my device - I'll usually have separate `/music` and `/podcasts` folders on my device.
> 
> So if /tmp is considered a removable device, and given the above assumptions about how it _should_ work, I would think your example absolute path of '/gposync/The Moth/The Moth Podcast_ Resolutions.mp3' is correct.

Okay with me. Indeed there is no way to guess what the real root is if the device root detection fails.

elelay:
Thank you for your code changes and for the testing.

> It would appear that `self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()` does not work correctly for me. I get the following error message and then go to the fallback code you posted:
> 
> ```
> 1739406744.544160 [gpodder.deviceplaylist] ERROR: find_enclosing_mount folder file:///Volumes/MAXMUZAK/test%20files/playlists failed: Containing mount for file /Volumes/MAXMUZAK/test files/playlists not found
> 1739406744.544207 [gpodder.deviceplaylist] WARNING: could not find mount point for MP3 player - using file:///Volumes/MAXMUZAK/test%20files as MP3 player root
> ```
> So my question is, given the above, are we expecting find_enclosing_mount().get_root() to succeed, or are we expecting to use the fallback code? Because if we are expecting find_enclosing_mount().get_root() to succeed, it does not. If we are expecting to need to use the fallback code, it does not appear to work as intended for me.
> 

good to know it doesn't work on macOS either. I hoped `find_enclosing_mount()` would work on more than Linux.
Thankfully `ismount()` works there. I haven't tried it on Windows yet but let's keep it as the preferred fallback.

> The fundamental assumption I am using is that the "device_sync.device_folder" setting does not _need_ to be the root of the device, despite how it is labelled in the preferences window. Is this the intended use case? Or should we revamp things so that it is clearly meant to be the root of the device, and then we can add some setting which sets the path from the root of the device to the desired podcasts folder?
> 

I think It would make the UI too complicated but we can try it in another PR.

> **Edit 2: After more research, I think I answered my own question - I think that the gio find_enclosing_mount() will only work on systems with dbus installed (i.e., linux only...?), is that correct? So MacOS and Windows will always use the fallback method?** If this is true, I might put a note or 2 in there to indicate that this is the case.

yes, please. Let's write down the usecases and the observed behaviour.

> Edit: I propose the following. This gives me the correct results for both absolute and relative playlist paths on MacOS:
>
> ```
>     def __init__(self, config, playlist_name):
>         self._config = config
>         self.linebreak = '\r\n'
>         self.playlist_file = (
>             util.sanitize_filename(playlist_name, self._config.device_sync.max_filename_length)
>             + '.' + self._config.device_sync.playlists.extension)
>         device_folder = util.new_gio_file(self._config.device_sync.device_folder)
>         self.playlist_folder = device_folder.resolve_relative_path(self._config.device_sync.playlists.folder)
>         self.playlist_to_device_relpath = os.path.relpath(device_folder, self.playlist_folder)
> 
>         self.mountpoint = None
>         try:
>             self.mountpoint = self.playlist_folder.find_enclosing_mount().get_root()
>         except GLib.Error as err:
>             logger.error('find_enclosing_mount folder %s failed: %s', self.playlist_folder.get_uri(), err.message)
>         # changed fallback code
>         if not self.mountpoint:
>             # ensure our path is a path, not a url - need to manually strip prefix
>             # this is largely so that os.path.ismount() will work correctly

Is there any chance that device_folder.get_path() would return the path directly instead of those 3 lines?
Tested under linux (file:/// and windows file:///: it works)
Otherwise I'm ok with this `__init__()` code.

>             drive_start = self._config.device_sync.device_folder.removeprefix('file:')
>             drive_start = drive_start.removeprefix('mtp:')
>             drive_start = request.url2pathname(drive_start)
> 
>             # find mount point, ensuring we don't end up locked up in the loop
>             while not os.path.ismount(drive_start) and drive_start != os.path.dirname(drive_start):
>                 drive_start = os.path.dirname(drive_start)
>             
>             # self.mountpoint must be a gio file
>             self.mountpoint = util.new_gio_file(drive_start)
> 
>             logger.warning('could not automatically find mount point for MP3 player - using %s as MP3 player root', self.mountpoint.get_uri())
>         self.playlist_absolute_filename = self.playlist_folder.resolve_relative_path(self.playlist_file)
> ```
> 
> 
> I also propose to change the setting label in the Preferences window from "Mountpoint:" to something like "Device podcasts folder:". How does that sound?

Good idea. Would *Folder containing podcasts on device* be good? Above the button, like the layout for choosing the playlist.
I don't know what the original intent was with the "podcast folder picker" button. Like you I used it to select a folder on the device, not the device root.

Let's do that in a separate PR. This way I can merge this one sooner ;-)

share/gpodder/extensions/rockbox_coverart.py
```diff
@@ -37,10 +50,196 @@ def on_episode_synced(self, device, episode):
...
+            if device_filetype.lower() not in ("jpeg", "jpg", "png"):
+                device_filetype = "jpg"
+
+            # sanitize for filetype checking - "jpg" will not match "jpeg"
+            if device_filetype.upper() == "JPG":
+                device_match_filetype = "JPEG"
+            else:
+                device_match_filetype = device_filetype.upper()
+
+            if (self.config.convert_and_resize_art is True):
```
elelay, line 67:
```suggestion
            if self.config.convert_and_resize_art:
```

share/gpodder/extensions/rockbox_coverart.py
elelay: it's the code after the `acquire()` that could throw and exception causing the lock not to be released.
I see that every branch has its own `try: except` so it's ok.

src/gpodder/gtkui/main.py
```diff
@@ -2008,7 +2008,8 @@ def do_save_episode(copy_from, copy_to):
                 if len(self.config.sendto.custom_file_format) > 0:
```
elelay, line 2008: We should always use `episode.sync_filename()` now.
Otherwise we don't take` self.config.device_sync.use_title_as_filename` into account when `custom_file_format` is empty.

elelay:
With mtpfs and gvfs-mtp installed and a gentle nudge (call mtpfs before running gPodder so it discovers the device) I get an Android mountpoint in the file browser to the folder on device:

screenshot-2025-02-15-003819

The selected path is `mtp://USB_Developer_Android_20080411/M%C3%A9moire%20de%20stockage%20interne/gPodder`

Gio.File.get_path() returns the hidden path to the real mountpoint:

```
>>> f = Gio.File.new_for_uri("mtp://USB_Developer_Android_20080411/M%C3%A9moire%20de%20stockage%20interne/gPodder")
>>> playlist_folder = f.resolve_relative_path("Playlists")
>>> playlist_folder.find_enclosing_mount().get_root().get_uri()
'mtp://USB_Developer_Android_20080411/'
>>> playlist_folder.find_enclosing_mount().get_root().get_path()
'/run/user/1000/gvfs/mtp:host=USB_Developer_Android_20080411'
```

The code in get_path_to_filename_for_playlist() works, but it produces:

```
#EXTM3U
#EXTINF:0,The Moth Podcast: Timeless Love
/Mémoire de stockage interne/gPodder/The Moth Podcast/CAD433540875.mp3
```
The correct path would have been `/gPodder/The Moth Podcast/CAD433540875.mp3` .

I don't see how to work around this in our code.
The user should not set `device_sync.playlists.use_absolute_path` in this case.

OK for you?

share/gpodder/extensions/rockbox_coverart.py
dconrad5: I'll put a comment in there saying to be careful when modifying that to catch all exceptions within there so that we always release the lock, or something to that effect.

src/gpodder/gtkui/main.py
dconrad5: Sure, I can do that. For that matter, what should we do if custom_file_format _is_ empty? Keep the fallback to the episode title, or fall back to the same filename as the computer? That seems like a reasonable check to do within episode.sync_filename() I think.

dconrad5:
Thanks for your patience with such a monster PR, I was selfish and wanted to use my changes while I waited for review!

> Is there any chance that device_folder.get_path() would return the path directly instead of those 3 lines? Tested under linux (file:/// and windows file:///: it works) Otherwise I'm ok with this `__init__()` code.

Looks like that does work for me on MacOS as well, so let's go with that. Simpler is better!

> > I also propose to change the setting label in the Preferences window from "Mountpoint:" to something like "Device podcasts folder:". How does that sound?
> > ```
> 
> Good idea. Would _Folder containing podcasts on device_ be good? Above the button, like the layout for choosing the playlist. I don't know what the original intent was with the "podcast folder picker" button. Like you I used it to select a folder on the device, not the device root.
> 
> Let's do that in a separate PR. This way I can merge this one sooner ;-)

Noted, will do!

> The selected path is `mtp://USB_Developer_Android_20080411/M%C3%A9moire%20de%20stockage%20interne/gPodder`
> 
> Gio.File.get_path() returns the hidden path to the real mountpoint:
> 
> ```
> >>> f = Gio.File.new_for_uri("mtp://USB_Developer_Android_20080411/M%C3%A9moire%20de%20stockage%20interne/gPodder")
> >>> playlist_folder = f.resolve_relative_path("Playlists")
> >>> playlist_folder.find_enclosing_mount().get_root().get_uri()
> 'mtp://USB_Developer_Android_20080411/'
> >>> playlist_folder.find_enclosing_mount().get_root().get_path()
> '/run/user/1000/gvfs/mtp:host=USB_Developer_Android_20080411'
> ```
> 
> The code in get_path_to_filename_for_playlist() works, but it produces:
> 
> ```
> #EXTM3U
> #EXTINF:0,The Moth Podcast: Timeless Love
> /Mémoire de stockage interne/gPodder/The Moth Podcast/CAD433540875.mp3
> ```
> 
> The correct path would have been `/gPodder/The Moth Podcast/CAD433540875.mp3` . _Edit_ : another correct path: `/mnt/sdcard/gPodder/The Moth Podcast/CAD433540875.mp3
> 
> I don't see how to work around this in our code. The user should not set `device_sync.playlists.use_absolute_path` in this case.
> 
> OK for you?

I don't have any MTP devices so it doesn't matter much to me at the moment :-P but what if we kept note of whether the mountpoint was found via Gio or fallback code? We could probably work around it then and go down one level. Would that be an acceptable solution?

src/gpodder/gtkui/main.py
dconrad5: Actually, now that I look at this, in this instance I think we want to remove use_title_as_filename from this particular usage. I didn't look closely enough the first time, this appears to be setting the filename for the local file, not the file on device.

dconrad5: Please see commits https://github.com/gpodder/gpodder/pull/1672/commits/33e83ae575b36a483355387f4a94dbd746902371 and https://github.com/gpodder/gpodder/pull/1672/commits/8625247f6a804fb2600282b7df94d64a95c58a49, I moved that check for zero-length custom string into `episode_sync_filename()` as well as removed that particular usage of `device_sync.use_title_as_filename`, I think it was being used in the wrong context there.

elelay: agreed.

src/gpodder/model.py
```diff
@@ -798,12 +798,15 @@ def was_downloaded(self, and_exists=False):
             return False
         return True
 
-    def sync_filename(self, use_custom=False, custom_format=None):
-        if use_custom:
+    def sync_filename(self, use_custom=False, custom_format=None, use_title=False):
+        if use_custom and len(custom_format) > 0:
```
elelay, line 802:
```suggestion
        if use_custom and custom_format:
```

elelay: Thank you!
I'm ready to merge with that last change.

dconrad5: Excellent! I'm glad I can contribute.

PR #1674, opened 2024-11-25
Fix docstring formatting

tpikonen:
Chip away at trying to make gPodder code mostly flake8 clean. Fix the formatting of docstrings to pass pydocstyle checks from flake8.

Also ignore spellcheck checks (SC*) at least for now.

The remaining flake8 errors are mostly about non-specific `except:` statements. I'm not sure if these should be fixed, or just ignored.

### DIFF
--- a/src/gpodder/model.py
+++ b/src/gpodder/model.py
@@ -77,8 +77,8 @@ class Feed:
         return None
 
     def get_new_episodes(self, channel, existing_guids):
-        """
-        Produce new episodes and update old ones.
+        """Produce new episodes and update old ones.
+
         Feed is a class to present results, so the feed shall have already been fetched.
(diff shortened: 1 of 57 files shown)


src/gpodder/model.py
```diff
@@ -489,9 +487,7 @@ def get_player(self, config):
         return player
 
     def can_play(self, config):
-        """
-        # gPodder.playback_episodes() filters selection with this method.
-        """
+        """# gPodder.playback_episodes() filters selection with this method."""
```
elelay, line 490:
```suggestion
        """gPodder.playback_episodes() filters selection with this method."""
```

elelay:
Thanks for this @tpikonen 
I'll merge once tests pass

PR #1680, opened 2024-12-14
Update Norwegian Nynorsk translation for next release

huftis:


huftis:
BTW, even though I have translated the three colour scheme strings (‘Light’, ‘Dark’ and ‘System’), they are still shown in English in the gPodder UI.

auouymous:
@tpikonen 92244d630 added translatable strings in `share/gpodder/ui/gtk/gpodderpreferences.ui`, but then used non-translatable strings in `src/gpodder/gtkui/desktop/preferences.py`. Do you know if it can use the translatable strings or do we need to translate them in `src/gpodder/gtkui/desktop/preferences.py`?

elelay:
Merged, thanks

PR #1681, opened 2024-12-14
Use locale-aware formatting in subscription statistics

huftis:
The average number of days for each podcast will now be shown using decimal commas in locales that use decimal commas.


elelay:
Merged, thanks for the attention to detail. Indeed it's also comma separated in French.

PR #1689, opened 2024-12-22
Move all dependencies to pyproject.toml

tpikonen:
While updating the flathub package of the latest release ([here](https://github.com/flathub/org.gpodder.gpodder/pull/33), please test if you can), I noticed the build system upgrade by @masoudd, which requires the 'build' and 'installer' python modules. These were not listed in the `build-system.requires` field of `pyproject.toml`, so add them there.

Also, now that we use `pyproject.toml` I think all dependencies should be listed there and only there. This means removing `tools/requirements.txt` and moving the modules there to either `dependencies` or `optional-dependencies` under `[project]`. Currently there is only a limited set of modules gPodder potentially uses in `dependencies` with some version skew. What do you think?


tpikonen:
The last two commits are a suggestion on how the optional dependencies could be organized.

masoudd:
Hello. Thank you for taking a look. I initially added `build` there and saw `build` installing `build` before building the wheel. like:

    $ python -m build
    * Creating isolated environment: venv+pip...
    * Installing packages in isolated environment:
      - build
      - setuptools>=64

That is just absurd in my mind. I keep reading the docs and going up down [the glossary](https://packaging.python.org/en/latest/glossary/) and see [the example](https://packaging.python.org/en/latest/tutorials/packaging-projects/) where they do call build at the end but don't add it in the `pyproject.toml` file and I'm just confused more.
If you need these added in the `pyproject.tom` to make the flathub maker work, then by all means.

>Also, now that we use pyproject.toml I think all dependencies should be listed there and only there.

I agree

tpikonen:
Installing `build` on a build target used by the same module indeed seems absurd.

The `build` and `installer` modules do not need to be listed in `pyproject.toml`, but they need to be included in the build manifest. The flatpak build essentially does `make install` in the source tree, which requires them to be available in the SDK environment.

If these dependencies should not be listed in `pyproject.toml`, is there any other place where they could be included, at least for documentation purposes?

tpikonen:
> If these dependencies should not be listed in `pyproject.toml`, is there any other place where they could be included, at least for documentation purposes?

They are listed in the `README.md`. I suppose this is enough.

pyproject.toml
```diff
@@ -17,15 +17,23 @@ classifiers = [
...
+    "podcastparser==0.6.10",
     "PyGObject",
+    "requests[socks]==2.32.3",
+    "urllib3==2.2.2",
 ]
 dynamic = ["version"]
 
 [project.optional-dependencies]
+full = [
+    "eyed3",
```
auouymous, line 31: > eyed3 is optional and pulls in a lot of dependencies, so disable by default

I assume yt-dlp must be specified to use it, could the same be done for eyed3?

auouymous (changes requested):

pyproject.toml
tpikonen: > I assume yt-dlp must be specified to use it, could the same be done for eyed3?

Probably, but TBH I'm not familiar enough with the Python build system to say how useful these build variants are.

The flatpak build manifest does `make install` which currently just calls `python -m build --wheel --no-isolation` without any other arguments. Could other installers (windows, mac) use the variants somehow?

auouymous: I also know nothing about python's build system, and have never used pip.

According to a quick search, `pip install '.[yt-dlp,eyed3]'` would install gpodder along with those optional deps, after running `make build` of course. I don't know if any linux distros make use of pip in their build system, but users could theoretically install gpodder this way. There is also `pip-gui`, and it or another frontend might extract optional deps and make them more discoverable.

The Windows and Mac builds manually install each dependency.

auouymous (approved):

auouymous:
The appveyor failure is most likely transient as other builds now have multiarch 16 to work with gdb 16.

@elelay Does this look good to you?

auouymous:
#1672 has been merged and requires `filelock` and `pillow` to use the `rockbox_coverart` extension. They were added to `tools/requirements.txt` but not to `pyproject.toml`. I assume both could be added as a single `rockbox-coverart` or `rockbox_coverart` dependency.

tpikonen:
I added a `rockbox-coverart` dependency group and rebased this to master.

pyproject.toml
```diff
@@ -17,15 +17,22 @@ classifiers = [
...
+    "urllib3==2.2.2",
 ]
 dynamic = ["version"]
 
 [project.optional-dependencies]
+eyed3 = ["eyed3"]
 gui = ["PySocks"]
+html5lib = ["html5lib==1.1"]
+mutagen = ["mutagen==1.47.0"]
+rocbox-coverart = ["filelock", "pillow"]
```
auouymous, line 34: Missing a 'k' in rockbox.

auouymous (changes requested):

pyproject.toml
tpikonen: Ah, thanks for spotting this.

elelay:
thanks for adding the coverart dependency.
Merging now

PR #1691, opened 2024-12-24
Change progress dialog to be non-modal.

auouymous:
The main window is unresponsive during long running actions on the main loop. The progress dialog appeared to notify the user of the action's progress.

However, if the action failed and opened another dialog, it would cause the progress dialog to appear due to exceeding the timeout. This caused a deadlock between two modal dialogs, where the error dialog couldn't be clicked and the progress dialog had no cancel button, or was maybe also not clickable.

Making the progress dialog non-modal should prevent this from happening. The user will be able to click the non-responsive UI behind the dialog, but that is better than gpodder locking up.

@elelay Do you see any reason not to do this? It should fix the startup problems on Windows.

### DIFF
--- a/src/gpodder/gtkui/interface/progress.py
+++ b/src/gpodder/gtkui/interface/progress.py
@@ -71,7 +71,6 @@ class ProgressIndicator(object):
     def _create_progress(self):
         self.dialog = Gtk.MessageDialog(self.parent,
                 0, 0, Gtk.ButtonsType.CANCEL, self.subtitle or self.title)
-        self.dialog.set_modal(True)
         self.dialog.connect('delete-event', self._on_delete_event)
         if self.cancellable:
             def cancel_callback(dialog, response):


auouymous:
I said it "should" fix the startup freezes but meant "might" as I don't know if anyone mentioned an error dialog also being open. It could have been hidden behind the progress dialog, or something else causing it. This is a problem though because I ran into it while trying to make database exceptions open an error dialog so users aren't presented with a blank gpodder window.

And now that I've thought more about it, this won't actually solve the issue unless the user moves the progress dialog to click on the hidden error dialog below it. We might need to place the progress indicator inside the gpodder window instead of it being a dialog. Or something like the following to move the progress dialog to top of gpodder window, so any error dialogs would be below it, centered on the gpodder window.

```patch
     def _create_progress(self):
         self.dialog = Gtk.MessageDialog(self.parent,
                 0, 0, Gtk.ButtonsType.CANCEL, self.subtitle or self.title)
+        ppos = self.parent.get_position()
+        dpos = self.dialog.get_position()
+        self.dialog.move(dpos.root_x, ppos.root_y)
-        self.dialog.set_modal(True)
         self.dialog.connect('delete-event', self._on_delete_event)
         if self.cancellable:
             def cancel_callback(dialog, response):
```

elelay:
Can you look into integrating it into the window? It looks to me like the best solution.

Was the modal progress dialog a way to prevent users from clicking on the UI while there is an operation in progress?

PR #1694, opened 2024-12-28
appdata: Replace developer_name tag with developer and name tags

tpikonen:
This fixes 'appstreamcli validate'.


elelay:
merged, thanks

PR #1695, opened 2024-12-28
Add an .editorconfig file

tpikonen:
The file has more or less sensible defaults for different file formats (IMO). The max_line_length is taken from `setup.cfg`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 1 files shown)


elelay:
Merged, thanks :-)

auouymous:
@tpikonen Won't the `[*] indent_style = space` cause problems in the makefile? There are also some files that use tabs that would get spaces mixed in with this, right?

tpikonen:
Grepping for `\t` in gPodder sources shows that files which contain tabs (other than the makefile) are shell scripts or XML files mostly under `tools`. I'm not a fan of tab-indentation on either of these file formats, but that's a matter of opinion.

I made a PR adding the tab indenting to Makefile editor configuration: #1699. Other formats can be added to the hard-tabs group there if needed.

PR #1696, opened 2024-12-28
Break lines to under 80 characters on markdown docs

tpikonen:
The .editorconfig file has 80 chars as the line length for documentation files, so break lines on the markdown docs to that length.


elelay:
Merged, thanks

PR #1697, opened 2024-12-30
Update gpodder.desktop.in

DinizBortolotto:
Desktop entry file has more than one 'Main Category'.

If multiple Main Categories are included in a single desktop entry file, the entry may appear more than once in the menu. I believe that gPodder belongs to 'Feed' category, so the more accurate classification of gPodder is a Network/Feed application.

References:
https://specifications.freedesktop.org/menu-spec/latest/category-registry.html https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html


tpikonen:
Thanks, this looks like a correct classification.

elelay:
Merged, thanks :-)

PR #1699, opened 2025-01-03
editorconfig: Add indent_style = tab for Makefiles

tpikonen:
Adds correct settings for tabs in Makefiles.


elelay:
Thanks, and thanks to @auouymous for the catch

PR #1712, opened 2025-01-18
Fallback format by default

Paculino:
Lately I've been noticing that videos (from various sources) are not available except in video only or audio only formats, which isn't a big deal, since they can be easily merged. This is a fallback for the yt-dlp -f option to include, as last resorts, a combination of the best video and best audio, or the best available alone. This may often end up being a larger than expected file, but it is a fallback, and if a problem, an after-download command could be calling ffmpeg to scale the video, or the time until deletion could be reduced.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -366,6 +366,7 @@ class gPodderYoutubeDL(download.CustomDownloader):
             return info, opts
 
     def fetch_video(self, info, opts):
+        self.add_format(self.gpodder_config, opts, fallback="bv*+ba/b")
         with youtube_dl.YoutubeDL(opts) as ydl:
             return ydl.process_video_result(info, download=True)
 


auouymous:
The problem is that some formats can become available before others. If someone can only afford to download small formats, this could result in costly automated downloads.

We could probably change all of the default formats to include best for the given size. That way you still get what you want or lesser. And if you only want MP4, you won't get WEBM.

https://github.com/gpodder/gpodder/blob/master/src/gpodder/youtube.py#L42
```
...
mp4_720 = [22, '136+140', 'bestvideo[mp4][width<=1280][height<=720][fps<60]+bestaudio']
...
webm_1080 = [46, '248+251', 'bestvideo[webm][width<=1920][height<=1080][fps<60]+bestaudio'] + mp4_1080
...
```

tpikonen:
I just tried to use gpodder to download audio-only content through yt-dlp and failed. My config (in Video->Preferred YouTube format) had just one of the default formats, which do not allow audio-only downloads. There is a warning (in bold) in the youtube-dl prefs page about this not working, but I think we can do better than that.

This PR, if updated to the latest version of the youtube-dl extension fixes this, but may download very large files.

Should we add fallbacks to all the formats in https://github.com/gpodder/gpodder/blob/master/src/gpodder/youtube.py#L42 like @auouymous suggested above, or is there another way?

Here's this PR ported to current master for reference:

```diff
--- a/share/gpodder/extensions/10_youtube-dl.py
+++ b/share/gpodder/extensions/10_youtube-dl.py
@@ -372,9 +372,9 @@ class gPodderYoutubeDL(download.CustomDownloader):
         if gpodder.config._proxies:
             opts['proxy'] = gpodder.config._proxies['http']
             logger.debug(f"Setting proxy from network setting proxy: {opts['proxy']}")
 
-        opts['format'] = self.get_ydl_format(self.gpodder_config)
+        opts['format'] = self.get_ydl_format(self.gpodder_config, fallback='bv*+ba/b')
         with youtube_dl.YoutubeDL(opts) as ydl:
             info = ydl.extract_info(url, download=False)
             return info, opts
```

PR #1701, opened 2025-01-04
tagging.py coverart rework

dconrad5:
Allow using episode-specific artwork, resizing artwork, converting filetypes, preferring episode-specific or channel artwork.

Depends on PR ~~#1672.~~

Documentation at https://github.com/gpodder/gpodder.github.io/pull/57

Everything is working well as far as I can tell.

MP4 coverart extraction not implemented for lack of available video podcast with embedded artwork.

PR #1715, opened 2025-01-25
Rework Device preferences tab

dconrad5:
Add a dropdown for device filename method (same as local, episode title, custom syncname). Make this grey out unless it's actually available (filesystem-based devices).

Add a custom syncname field to the device tab, which is only available when using custom syncname.

Added max_filename_length, one_folder_per_podcast, use_absolute_path. Decided to omit playlists.extension. Reordered things in that tab as well.

### DIFF
--- a/share/gpodder/ui/gtk/gpodderpreferences.ui
+++ b/share/gpodder/ui/gtk/gpodderpreferences.ui
@@ -21,6 +21,13 @@
     <property name="step-increment">1</property>
     <property name="page-increment">1</property>
   </object>
+  <object class="GtkAdjustment" id="adjustment_max_filename_length">
+    <property name="lower">30</property>
+    <property name="upper">150</property>
+    <property name="value">60</property>
+    <property name="step-increment">5</property>
+    <property name="page-increment">5</property>
+  </object>
   <object class="GtkDialog" id="gPodderPreferences">
     <property name="can-focus">False</property>
     <property name="title" translatable="yes">Preferences</property>
@@ -993,6 +1000,21 @@
                               </packing>
                             </child>
                             <child>
+                              <object class="GtkCheckButton" id="checkbutton_one_folder_per_podcast">
+                                <property name="label" translatable="yes">Create separate folders for each podcast</property>
+                                <property name="visible">True</property>
+                                <property name="can-focus">True</property>
+                                <property name="receives-default">False</property>
+                                <property name="halign">start</property>
+                                <property name="draw-indicator">True</property>
+                              </object>
+                              <packing>
+                                <property name="expand">False</property>
+                                <property name="fill">True</property>
+                                <property name="position">2</property>
+                              </packing>
+                            </child>
+                            <child>
                               <object class="GtkSeparator">
                                 <property name="visible">True</property>
                                 <property name="can-focus">False</property>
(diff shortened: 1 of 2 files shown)


dconrad5:
I realized there are a couple more settings that would be good to have in the preferences window: 
- max_filename_length
- one_folder_per_podcast
- playlists.extension (maybe)
- use_absolute_path

So I think I will add these to this PR as well.

dconrad5:
> * max_filename_length
> 
> * one_folder_per_podcast
> 
> * playlists.extension (maybe)
> 
> * use_absolute_path

Added max_filename_length, one_folder_per_podcast, use_absolute_path. Decided to omit playlists.extension. Reordered things in that tab as well.

src/gpodder/util.py
```diff
@@ -1632,7 +1632,8 @@ def sanitize_filename(filename, max_length):
         filename = filename.decode('utf-8')
 
     # see #361 - at least slash must be removed
-    filename = re.sub(r"[\"*/:<>?\\|]", "_", filename)
+    # add additional common punctuation
+    filename = re.sub(r"[\"\'\‘\’\“\”&\[\]!@#$%^()+=,;\{\}*/:<>?\\|]", "_", filename)
```
auouymous, line 1636: I don't like whitespace or puntuation (other than periods, hyphens and underscores) in my filenames but this might be excessive. It should only remove characters that are known to cause problems on some OS.

Caused #1716.

auouymous (changes requested):

src/gpodder/util.py
dconrad5: I can certainly back off on that. I was chasing down why some podcast playlists didn't work and some did, I think. Now that Rockbox's default codepage is utf-8, I think most of that can be reverted. I'll play with it when I get home today.

Kind of chuffed someone was using my patchset, to be honest!

dconrad5: Ok, I can hardly call my testing exhaustive, but for the podcasts I listen to, rockbox's change to utf-8 as the default codepage seems to have fixed my playlist issues that caused me to do this commit. I'll go ahead and revert it.

dconrad5: Both 4d78ccc and 4086ee0 should now be fully reverted.

elelay:
you can wait until #1701 is merged or rebase on master now

auouymous:
Ignore my last comment, I didn't realize the commit was reverted instead of being removed.

dconradSukup:
> Ignore my last comment, I didn't realize the commit was reverted instead of being removed.

Yeah, sorry about that. I actually just yesterday learned about "interactive rebasing", which is probably what I should have been doing all along.

dconrad5:
> Ignore my last comment, I didn't realize the commit was reverted instead of being removed.

Yeah, sorry about that. I actually just yesterday learned about interactive rebasing, which I probably should have been doing all along with squashing and fixups and the like. That would have cleaned things up immensely.

I should be able to rebase this today or tomorrow, by the way.

dconrad5:
Now rebased on top of PR #1701

dconrad5:
Now rebased on top of current master

elelay:
Merged, thanks

PR #1721, opened 2025-02-03
Rename download directory after overriding title from OPML file.

auouymous:
Channel title from OPML file overrides title provided by feed. However, the download directory is set prior to that and must be renamed.

This is the same way gpo handles it.


elelay:
Merged, thanks for the fix!

PR #1722, opened 2025-02-05
model.py: fix typos in references to PodcastEpisode.published_formatted()

dconrad5:
Looks like a typo caused the "episode.sortdate" field to break.


elelay:
Good catch, thank you @dconrad5 :+1:

Issue #1688, opened 2024-12-22
Error on startup with latest installer version on w7


Snippet24816:
Hi, I'm getting this: The procedure entry point CopyFile2 could not be located in the dynamic link library KERNEL32.dll.
Most likely is related to Windows 7 or some service, in case is useful to someone else, this appears to be the solution 
https://www.youtube.com/watch?v=uZddkN0iFcE
I will try and report back

masoudd:
I found https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfile2 says minimum is windows 8. gpodder uses NSIS to make the installer and [here they claim it supports all windows versions](https://nsis.sourceforge.io/Features).

Snippet24816:
Ok didn't work out as it seems CopyFile2  is available from W8 onwards, this time no workaround seems to be possible so I'm closing this thread.

Snippet24816:
I will reopen the thread just in case...

masoudd:
I'm not sure about which version of windows gpodder wants to support, but I think windows 7 should still be supported if at all possible. I'm still not sure why the installer uses things that are win8 and above. I'll try to find out if we can avoid it

Snippet24816:
Thank you

elelay:
Hi @Snippet24816 @masoudd,
to clarify: the installer runs fine on windows 7 (tried it myself), but gpodder.exe or python.exe use CopyFile2.
We use python.exe from msysgit.
The portable version is also broken.

And gpodder 3.11.1 is the last version working with win7.
gpodder 3.11.2 to 3.11.4 complain of missing api-ms-win-core-path-l1-1-0.dll see #1540

elelay:
Closing now because it is an "issue" with msys2 dropping windows 7 support, that we can't work around.

Snippet24816:
Hi that can be worked around see here please: https://github.com/gpodder/gpodder/issues/1540#issuecomment-2221780643
What I cant workaround is the use CopyFile2 instead of CopyEx. I had been using 3.11.4 happily with that.

elelay:
Indeed, we can't work around CopyFile2.

Snippet24816:
May I ask for the latest build before using copyFile2, I know I been using some build in the 11.4 but I can't find it :(

auouymous:
[...]

PR #1729, opened 2025-03-02
Add support for updating in youtube-dl extension preferences

masoudd:
This pull request adds two buttons to the youtube-dl extension preferences:

yt-dlp

The "Check for update" button runs `pip index versions yt-dlp` and attempts to parse its output for the line `LATEST: ...` and updates the label to show it.
<details><summary><code>pip index versions yt-dlp</code> output</summary>

```
$ pip index versions yt-dlp
WARNING: pip index is currently an experimental command. It may be removed/changed in a futur
e release without prior warning.
yt-dlp (2025.2.19)
Available versions: 2025.2.19, 2025.1.26, 2025.1.15, 2025.1.12, 2024.12.23, 2024.12.13, 2024.
12.6, 2024.12.3, 2024.11.18, 2024.11.4, 2024.10.22, 2024.10.7, 2024.9.27, 2024.8.6, 2024.8.1,
 2024.7.25, 2024.7.16, 2024.7.9, 2024.7.8, 2024.7.7, 2024.7.2, 2024.7.1, 2024.5.27, 2024.5.26
, 2024.4.9, 2024.3.10, 2023.12.30, 2023.11.16, 2023.11.14, 2023.10.13, 2023.10.7, 2023.9.24,
2023.7.6, 2023.6.22, 2023.6.21, 2023.3.4, 2023.3.3, 2023.2.17, 2023.1.6, 2023.1.2, 2022.11.11
, 2022.10.4, 2022.9.1, 2022.8.19, 2022.8.14, 2022.8.8, 2022.7.18, 2022.7.17, 2022.6.29, 2022.
6.22.1, 2022.6.22, 2022.5.18, 2022.4.8, 2022.3.8.2, 2022.3.8.1, 2022.3.8, 2022.2.4, 2022.2.3,
 2022.1.21, 2021.12.27, 2021.12.25, 2021.12.1, 2021.11.10.1, 2021.11.10, 2021.10.22, 2021.10.
10, 2021.10.9, 2021.9.25, 2021.9.2, 2021.9.1, 2021.8.10, 2021.8.2, 2021.7.24, 2021.7.21, 2021
.7.7, 2021.6.23, 2021.6.9, 2021.6.1, 2021.5.20, 2021.5.11, 2021.4.22, 2021.4.11, 2021.4.3, 20
21.3.24.1, 2021.3.21, 2021.3.15, 2021.3.7, 2021.3.3.2, 2021.3.1, 2021.2.24, 2021.2.19, 2021.2
.15, 2021.2.9, 2021.2.4, 2021.1.29, 2021.1.24.post1, 2021.1.24, 2021.1.20, 2021.1.16
  INSTALLED: 2025.2.19
  LATEST:    2025.2.19

``` 

</details> 

The "Update" button runs `pip install --upgrade yt-dlp`.

I'm asking for your thoughts about this approach. Is this layout ok?

Possible improvements:
- Should this be shown in all gpodder installs? Or should it be limited to the windows and mac versions? This won't work for gpodder installed from linux repositories that manage `yt-dlp` themselves.
- Is it possible to re-import the updated `yt-dlp`  module? Currently you need to restart gpodder to use the updated module.
- Currently the UI freezes when running the pip commands

See initial discussion: https://github.com/gpodder/gpodder/issues/1688#issuecomment-2667378738

Let me know if anything needs to change.

### DIFF
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -596,7 +596,7 @@ class gPodderExtension:
                 self.latest_version = match.group(1)
                 success = True
             else:
-                logger.Error("Could not find LATEST version in pip output")
+                logger.Error("Could not find LATEST version in pip output:\n%s", output)
                 self.latest_version = None
         except Exception as e:
             logger.Error("Error checking for latest version: %r", e)
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -520,6 +520,7 @@ class gPodderExtension:
         self.infobar = None
         self.latest_version = _('Not checked')
         self.update_label = None
+        self.update_btn = None
 
     def on_load(self):
         self.ytdl = gPodderYoutubeDL(self.container.manager.core.config, self.container.config)
@@ -581,20 +582,34 @@ class gPodderExtension:
             self.container.config.embed_subtitles = False
 
     def check_for_update(self, widget):
+        success = False
         try:
             output = subprocess.check_output([sys.executable, '-m', 'pip', 'index', 'versions', program_name]).decode('utf-8')
             logger.debug(f"pip index version {program_name}: {output}")
             match = re.search(r'LATEST:\s*(\S*)', output)
             if match:
                 self.latest_version = match.group(1)
+                success = True
             else:
                 logger.Error("Could not find LATEST version in pip output")
-                self.latest_version = _('Error')
+                self.latest_version = None
         except Exception as e:
             logger.Error(e)
-            self.latest_version = _('Error')
+            self.latest_version = None
 
-        self.update_label.set_text('%s %s' % (_('Available version:'), self.latest_version))
+        if success:
+            # Need to remove zeros from single digit days and months
+            # Because pip returns versions like 2021.2.9 but the module reports it like 2021.02.09
+            version = ".".join([str(int(i)) for i in youtube_dl.version.__version__.split('.')])
+            if self.latest_version != version:
+                self.update_btn.set_sensitive(True)
+                self.update_label.set_text('%s %s' % (_('Available version:'), self.latest_version))
+            else:
+                self.update_btn.set_sensitive(False)
+                self.update_label.set_text(_('No update available'))
+        else:
+            self.update_btn.set_sensitive(False)
+            self.update_label.set_text(_('Error checking for update'))
 
     def do_update(self, widget):
         try:
@@ -645,21 +660,24 @@ class gPodderExtension:
         self.infobar = infobar
         box.pack_end(infobar, False, False, 0)
 
-        box.pack_start(Gtk.HSeparator(), False, False, 0)
+        # Do not show the pip update buttons in linux
+        if gpodder.ui.win32 or gpodder.ui.osx:
+            box.pack_start(Gtk.HSeparator(), False, False, 0)
 
-        note_label = Gtk.Label(_("Note: You need to restart gpodder after updating %s") % (program_name))
-        box.pack_start(note_label, False, False, 0)
-        update_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, spacing=10)
-        check_for_update_btn = Gtk.Button(_('Check for update'))
-        check_for_update_btn.connect('clicked', self.check_for_update)
-        self.update_label = Gtk.Label('%s %s' % (_('Available version:'), self.latest_version))
-        do_update_btn = Gtk.Button(_('Update'))
-        do_update_btn.connect('clicked', self.do_update)
+            note_label = Gtk.Label(_("Note: You need to restart gpodder after updating %s") % (program_name))
+            box.pack_start(note_label, False, False, 0)
+            update_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, spacing=10)
+            check_for_update_btn = Gtk.Button(_('Check for update'))
+            check_for_update_btn.connect('clicked', self.check_for_update)
+            self.update_label = Gtk.Label('%s %s' % (_('Available version:'), self.latest_version))
+            self.update_btn = Gtk.Button(_('Update'))
+            self.update_btn.connect('clicked', self.do_update)
+            self.update_btn.set_sensitive(False)  # Only enable it if check update finds newer version
 
-        update_box.pack_start(self.update_label, False, False, 0)
-        update_box.add(check_for_update_btn)
-        update_box.add(do_update_btn)
-        box.pack_start(update_box, False, False, 0)
+            update_box.pack_start(self.update_label, False, False, 0)
+            update_box.add(check_for_update_btn)
+            update_box.add(self.update_btn)
+            box.pack_start(update_box, False, False, 0)
 
         box.show_all()
         infobar.hide()
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -584,8 +584,12 @@ class gPodderExtension:
     def check_for_update(self, widget):
         success = False
         try:
-            output = subprocess.check_output([sys.executable, '-m', 'pip', 'index', 'versions', program_name]).decode('utf-8')
-            logger.debug(f"pip index version {program_name}: {output}")
+            output = subprocess.check_output(
+                    [sys.executable, '-m', 'pip', 'index', 'versions', program_name],
+                    stderr=subprocess.STDOUT,
+                    encoding='utf-8',
+                    timeout=60)
+            logger.debug("pip index version %s: %s", program_name, output)
             match = re.search(r'LATEST:\s*(\S*)', output)
             if match:
                 self.latest_version = match.group(1)
@@ -594,7 +598,7 @@ class gPodderExtension:
                 logger.Error("Could not find LATEST version in pip output")
                 self.latest_version = None
         except Exception as e:
-            logger.Error(e)
+            logger.Error("Error checking for latest version: %r", e)
             self.latest_version = None
 
         if success:
@@ -603,7 +607,7 @@ class gPodderExtension:
             version = ".".join([str(int(i)) for i in youtube_dl.version.__version__.split('.')])
             if self.latest_version != version:
                 self.update_btn.set_sensitive(True)
-                self.update_label.set_text('%s %s' % (_('Available version:'), self.latest_version))
+                self.update_label.set_text(_('Available version: %(version)s') % {'version': self.latest_version})
             else:
                 self.update_btn.set_sensitive(False)
                 self.update_label.set_text(_('No update available'))
@@ -613,9 +617,15 @@ class gPodderExtension:
 
     def do_update(self, widget):
         try:
-            subprocess.check_call([sys.executable, '-m', 'pip', 'install', '--upgrade', program_name])
+            subprocess.check_output(
+                    [sys.executable, '-m', 'pip', 'install', '--upgrade', program_name],
+                    stderr=subprocess.STDOUT,
+                    encoding='utf-8',
+                    timeout=120)
+        except subprocess.CalledProcessError as e:
+            logger.error("Error running %s: exit code %s, output: %s", e.cmd, e.returncode, e.output)
         except Exception as e:
-            logger.Error(e)
+            logger.exception("Error updating %s: %r", program_name, e)
 
     def show_preferences(self):
         box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=10)
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -588,6 +588,7 @@ class gPodderExtension:
                     [sys.executable, '-m', 'pip', 'index', 'versions', program_name],
                     stderr=subprocess.STDOUT,
                     encoding='utf-8',
+                    close_fds=True,
                     timeout=60)
             logger.debug("pip index version %s: %s", program_name, output)
             match = re.search(r'LATEST:\s*(\S*)', output)
@@ -621,6 +622,7 @@ class gPodderExtension:
                     [sys.executable, '-m', 'pip', 'install', '--upgrade', program_name],
                     stderr=subprocess.STDOUT,
                     encoding='utf-8',
+                    close_fds=True,
                     timeout=120)
         except subprocess.CalledProcessError as e:
             logger.error("Error running %s: exit code %s, output: %s", e.cmd, e.returncode, e.output)
--- a/share/gpodder/extensions/youtube-dl.py
+++ b/share/gpodder/extensions/youtube-dl.py
@@ -7,6 +7,7 @@
 import logging
 import os
 import re
+import subprocess
 import sys
 import time
 from collections.abc import Iterable
@@ -517,6 +518,8 @@ class gPodderExtension:
         self.container = container
         self.ytdl = None
         self.infobar = None
+        self.latest_version = _('Not checked')
+        self.update_label = None
 
     def on_load(self):
         self.ytdl = gPodderYoutubeDL(self.container.manager.core.config, self.container.config)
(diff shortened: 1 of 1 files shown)


masoudd:
Sorry for the force pushes, forgot to run linter.

auouymous:
Good job!


> I'm asking for your thoughts about this approach. Is this layout ok?

LGTM. After checking, if latest version is same as `youtube_dl.version.__version_`, it should set it to no update available or something, and maybe disable the update button.


> Should this be shown in all gpodder installs?

It should always be shown for Mac and Windows, Linux is questionable. You can use `gpodder.ui.win32 or gpodder.ui.osx` to determine if not linux. If Linux, it could check for pip and show a warning to only use it if yt-dlp was installed via pip. However, a Linux user who installed yt-dlp via pip, and maybe also gpodder, knows how to manually update yt-dlp. So it might be best to never show it for Linux users, to avoid any "accidents".


> Is it possible to re-import the updated yt-dlp module?

Yes, https://stackoverflow.com/a/1254379, but it won't kill the user to restart gpodder, and avoid any bugs that might be caused by reloading the module.

tpikonen:
Having this enabled for Linux is probably not a good idea, especially since this happens when trying to install anything with pip on Debian and probably other distros as well:

```
$ pip install --user yt-dlp
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.13/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

pipx likely also does not work, since we want the yt-dlp module, not only the command.

masoudd:
OK this will only show the update buttons on win32 and osx, Also the update button is disabled by default and only gets enabled if the check version button finds a different version. I only checked equality between the two versions, assuming the version reported by pip is >= from the installed version

computerfreak105:
Hi, what a nice thing! Thanks for making this possible!

Do I understand correctly that we can configure these options in the config.json as well? This would be perfect.

masoudd:
> Hi, what a nice thing! Thanks for making this possible! Do I understand correctly that we can configure these options in the config.json as well? This would be perfect.

Hi. I'm not sure. This just adds buttons that basically run command-line commands when clicked. It is meant as a convenience for users that are not familiar with command-line to be able to update the yt-dlp package. What aspect of it do you want to configure?

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,42 @@ def toggle_embed_subtitles(self, widget):
         else:
             self.container.config.embed_subtitles = False
 
+    def check_for_update(self, widget):
+        success = False
+        try:
+            output = subprocess.check_output([sys.executable, '-m', 'pip', 'index', 'versions', program_name]).decode('utf-8')
```
elelay, line 587:
```suggestion
            output = subprocess.check_output([sys.executable, '-m', 'pip', 'index', 'versions', program_name], stderr=subprocess.STDOUT, encoding='utf-8', timeout=60)
```

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,42 @@ def toggle_embed_subtitles(self, widget):
...
+            logger.debug(f"pip index version {program_name}: {output}")
```
elelay, line 588: donn't generate the debug string if not needed:
```suggestion
            logger.debug("pip index version %s: %s", program_name, output)
```

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,42 @@ def toggle_embed_subtitles(self, widget):
...
+            match = re.search(r'LATEST:\s*(\S*)', output)
+            if match:
+                self.latest_version = match.group(1)
+                success = True
+            else:
+                logger.Error("Could not find LATEST version in pip output")
+                self.latest_version = None
+        except Exception as e:
+            logger.Error(e)
```
elelay, line 597:
```suggestion
            logger.error("Error checking for latest version: %r", e)
```

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,42 @@ def toggle_embed_subtitles(self, widget):
...
+            self.latest_version = None
+
+        if success:
+            # Need to remove zeros from single digit days and months
+            # Because pip returns versions like 2021.2.9 but the module reports it like 2021.02.09
+            version = ".".join([str(int(i)) for i in youtube_dl.version.__version__.split('.')])
+            if self.latest_version != version:
+                self.update_btn.set_sensitive(True)
+                self.update_label.set_text('%s %s' % (_('Available version:'), self.latest_version))
```
elelay, line 606: makes it easier to translate:
```suggestion
                self.update_label.set_text(_('Available version: %(version)s' % {'version': self.latest_version}))
```

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,42 @@ def toggle_embed_subtitles(self, widget):
...
+                self.update_label.set_text(_('No update available'))
+        else:
+            self.update_btn.set_sensitive(False)
+            self.update_label.set_text(_('Error checking for update'))
+
+    def do_update(self, widget):
+        try:
+            subprocess.check_call([sys.executable, '-m', 'pip', 'install', '--upgrade', program_name])
+        except Exception as e:
+            logger.Error(e)
```
elelay, line 618: I'd capture the output for exception logging:

```suggestion
            subprocess.check_output([sys.executable, '-m', 'pip', 'install', '--upgrade', program_name], stderr=subprocess.STDOUT, encoding='utf-8', timeout=120)
        except subprocess.CalledProcessError as e:
            logger.error("Error running %s: exit code %s, output: %s", e.cmd, e.returncode, e.output)
        except Exception as e:
            logger.exception("Error updating %s: %r", program_name, e)
```

elelay (changes requested):
LGTM.
Let's see how it fares on diverse user installs :-)

computerfreak105:
Hi Masoud, it would be great if the parameters of these buttons could be set in the settings.json as well under the youtube-dl section.

 �

The section looks like this:

 �

    "youtube-dl": {

      "embed_subtitles": false,

      "manage_channel": false,

      "manage_downloads": true

    }

  },

 �

So, there would be additional entries like

      "update": 

 �

When set to true Gpodder would do the update command before downloading or maybe once after starting up gpodder.

For blind users the GUI isn’t accessible anymore. This is the reason why I’d like to have the possibility of doing the same without having to click these buttons.

 �

Your addition allows to update yt-dlp also under windows, right? If yes, this would be really huge because this wasn’t possible for the average user until now. If this is completely about something else which I missed, then please forgive my excitement and just ignore it :-)

 �

 �

auouymous:
@computerfreak105 is asking for a command in gpo to perform the update. I don't know if extensions can register commands, but I'll look into it.

auouymous:
#1730 will be needed to add gpo support.

share/gpodder/extensions/youtube-dl.py
auouymous: It was planned to use a progress dialog and show error output in a dialog, this is a rough draft. Logging the update and error would be a good idea though.

share/gpodder/extensions/youtube-dl.py
masoudd: Should I `%` the string input to `_()` or `%` its output?
`set_text(_('Available version: %(version)s' % {'version': self.latest_version}))` or
`set_text(_('Available version: %(version)s') % {'version': self.latest_version})`

auouymous: `set_text(_('Available version: %(version)s') % {'version': self.latest_version})`

Always apply `%` to the result of `_()`.

masoudd:
@computerfreak105 
>For blind users the GUI isn’t accessible anymore.

I'm sorry to hear that. What do you mean "anymore"? Was it accessible before? Can we make it accessible again?
I admit I'm not sure how you use gpodder. Do you use a screen reader? If so what is its name? I want to replicate what you use to see if I can make it work.

>This is the reason why I’d like to have the possibility of doing the same without having to click these buttons.

That is certainly possible. I will try to find a way. If you use gpo, #1730 can help. I will investigate further.

>Your addition allows to update yt-dlp also under windows, right? If yes, this would be really huge because this wasn’t possible for the average user until now

Yes. That is my goal.

> If this is completely about something else which I missed, then please forgive my excitement and just ignore it :-)

Thank you for your input. The accessibility aspects of a software is often forgotten by developers. Please do no hesitate to chime in in the future.

share/gpodder/extensions/youtube-dl.py
elelay: ah yes, you're right. Thanks @auouymous

computerfreak105:
Hi Masoud

Thanks, that’s really kind of you!

 �

GPodder for windows was accessible for hte most part, but this was maybe ten years ago. Today there isn’t much a blind person can do anymore other than let it run and prepare everything beforehand by importing .opml subscribtions from GPO CLI. If the GUI would become accessible again I could just grab a youtube channel and add it in gpodder. But therefore the menu would have to be made accessible again.

And also the list of podcasts and youtube channels. Blind people need the possibility to navigate such lists with the arrow keys of a computer keyboard. Ideally arrow up and down would jump from podcast/youtube channel. Arrow right would expand a selected entry so that the episodes can be navigated as well, arrow left would close the entry again. Maybe you get the idea? Pressing the right context menu key or Shift+F10 would open the context menu with options for removing the selected entry and what else is possible there.

 �

I use the screen reader JAWS for windows.

 �

https://support.freedomscientific.com/Downloads/JAWS

 �

Sadly they only provide 40 minutes of demo until one has to restart the computer in order to run it for another 40 minutes. This is hardly enough for getting a real feel for the screen reader.

But if you are really inclined and have the time, then please read this first:

 �

https://www.boia.org/blog/jaws-screen-reader-an-overview-for-developers-and-content-creators

 �

I’m afraid I’m not really useful when it comes to coding and such, but I can test things pretty well. So, if you’ll ever need me for a feedback I’ll gladly test gpodder for you with JAWS. I think there is at least another blind person here who might benefit from it and help you as well.

 �

Thanks again and warm regards

 �

 �

 �

 �

 �

 �

auouymous:
> GPodder for windows was accessible for hte most part, but this was maybe ten years ago.

I wonder if it was the gtk3 port that broke it.


> Ideally arrow up and down would jump from podcast/youtube channel. Arrow right would expand a selected entry so that the episodes can be navigated as well, arrow left would close the entry again.

I use a patch that makes left and right arrows move between channel, episode and progress lists. It prevents moving cursor in shownotes, or shift selecting with arrows there. It also prevents using those arrows in other widgets, some of which the patch detects and does the right thing. So overriding arrow behavior would cause usability issues if we miss any exclusions or add new widgets in the future.

Tab currently jumps between the 4-5 visible widgets. But it would be easy to add ctrl combos or function keys to jump to specific lists. Would that help any or is splitting channels and episodes into separate tabs mandatory?


> Pressing the right context menu key or Shift+F10 would open the context menu with options for removing the selected entry and what else is possible there.

The menu key already opens context menus for selected channel, episode or progress task.

computerfreak105:
Hi Auouymous

 �

Sounds like an usable patch you have. I’d love to try it out some time and see if the JAWS Screen Reader responds to it correctly.

So, if I understand correctly the arrow up and down keys would go through all channels. If I’d press arrow right the focus would be in the list of episodes  which I could also navigate with arrow down and up. And if I would press arrow right a third time I’d find myself in the progress bar where all the episodes are shown which are currently downloading? This would be very nice and further shortcuts to the widgets and whatever the GUI offers would be great.

 �

The menu would have to become accessible with keyboard navigation again. Maybe it still responds to the alt and arrow keys. But JAWS doesn’t react at all . Normally when I press the alt key in accessible applications the cursor jumps to the menu and jaws speaks all entries when I press the arrow keys.

 �

 �

auouymous:
> Sounds like an usable patch you have. I’d love to try it out some time and see if the JAWS Screen Reader responds to it correctly.

I just looked at the patch and gpodder already partially supports left and right arrows. You must first focus the channel or episode list using the tab key, then left arrow goes to channel and right goes to episode. Not as useful as my patch but it isn't buggy. ;)


> So, if I understand correctly the arrow up and down keys would go through all channels. If I’d press arrow right the focus would be in the list of episodes  which I could also navigate with arrow down and up.

The up and down arrows currently do move through channels or episodes, once a list is focused with the tab key.


> And if I would press arrow right a third time I’d find myself in the progress bar where all the episodes are shown which are currently downloading?

Yes, my patch includes progress list, and wraps around if you keep pressing left or right arrows.


> The menu would have to become accessible with keyboard navigation again. Maybe it still responds to the alt and arrow keys. But JAWS doesn’t react at all . Normally when I press the alt key in accessible applications the cursor jumps to the menu and jaws speaks all entries when I press the arrow keys.

The menu bar puts underscores below a letter in each item when alt is held down, but alt+arrows do nothing. alt+p opens podcasts menu, alt+s opens subscriptions, alt+e opens episodes, alt+x opens extras and alt+v opens view. The up and down arrows work once a menu is open. And if you tab to the channel or episode lists, the menu key will open a context menu for the selected channel or episode, and up/down arrows work in them too.

masoudd:
From `share/gpodder/examples/hello_world.py`:
```
# Concurrency Warning: use gpodder.util.Popen() instead of subprocess.Popen()
#
# When using subprocess.Popen() to spawn a long-lived external command,
# such as ffmpeg, be sure to include the "close_fds=True" argument.
#
# https://docs.python.org/3/library/subprocess.html#subprocess.Popen
#
# This is especially important for extensions responding to
# on_episode_downloaded(), which runs whenever a download finishes.
#
# Otherwise that process will inherit ALL file descriptors gPodder
# happens to have open at the moment (like other active downloads).
# Those files will remain 'in-use' until that process exits, a race
# condition which prevents gPodder from renaming or deleting them on Windows.
#
# Caveat: On Windows, you cannot set close_fds to true and also
# redirect the standard handles (stdin, stdout or stderr). To collect
# output/errors from long-lived external commands, it may be necessary
# to create a (temp) log file and read it afterward.
```

If I'm not mistaken, this means if gpodder finishes downloading a file while one of the pip commands are running it would cause a race condition?
So I'd have to rewrite the subprocess calls to use `gpodder.util.Popen` and use a temp file to parse the output?

elelay:
> If I'm not mistaken, this means if gpodder finishes downloading a file while one of the pip commands are running it would cause a race condition? So I'd have to rewrite the subprocess calls to use `gpodder.util.Popen` and use a temp file to parse the output?

yes. Good point! Totally forgotten about that

masoudd:
Reading into the docs for `Popen` I found:

https://docs.python.org/3.13/library/subprocess.html#popen-constructor

>Changed in version 3.7: On Windows the default for close_fds was changed from [False](https://docs.python.org/3.13/library/constants.html#False) to [True](https://docs.python.org/3.13/library/constants.html#True) when redirecting the standard handles. It’s now possible to set close_fds to [True](https://docs.python.org/3.13/library/constants.html#True) when redirecting the standard handles.

And since even [the current debian stable (Bookworm) ships python 3.11](https://wiki.debian.org/Python#Supported_Python_Versions), It's safe to assume we can use `close_fds`. It's actually defaulting to `True` now but I'll specify it explicitly.

masoudd:
This is not tested on mac. I have tested on windows 10 and it seems to work.

elelay:
great, thanks.
Let's review it one last time this week-end and merge.

share/gpodder/extensions/youtube-dl.py
```diff
@@ -577,6 +581,54 @@ def toggle_embed_subtitles(self, widget):
...
+                    encoding='utf-8',
+                    close_fds=True,
+                    timeout=60)
+            logger.debug("pip index version %s: %s", program_name, output)
+            match = re.search(r'LATEST:\s*(\S*)', output)
+            if match:
+                self.latest_version = match.group(1)
+                success = True
+            else:
+                logger.Error("Could not find LATEST version in pip output")
```
elelay, line 599:
```suggestion
                logger.Error("Could not find LATEST version in pip output:\n%s", output)
```

elelay (approved):
Looks good to me. Just please put the non-matching pip index output in the log so that users can report them.

elelay:
merged, thanks :+1:

PR #1731, opened 2025-03-03
Port MPRIS listener extension to GDBus, remove org.gpodder.player Dbus interface

tpikonen:
Replace the Dbus communication from the MPRIS listener extensions to gPodder with actions, which allows removing the org.gpodder.player Dbus interface.

Port the MPRIS listener code in the extension to GDBus.

Seems to work OK, but more testing would be appreciated.


elelay:
I depended on the org.gpodder.player interface in #1343.
Surely it can be adapted, but I'd rather look at it again before merging (or would you like to look into it?)

tpikonen:
> I depended on the org.gpodder.player interface in #1343.

The `enqueue_in_mediaplayer` extension and #1343 seem to depend on the MPRIS player API, not on the org.gpodder.player Dbus interface, which probably precedes MPRIS. 

I could not rebase #1343 on top of this for unrelated reasons, but at least `enqueue_in_mediaplayer` as it is now works fine with this PR.

If we want to remove the dbus-python dependency from gPodder, then the `MPRISResumer` class needs to be ported to GDBus though.

elelay:
> The `enqueue_in_mediaplayer` extension and #1343 seem to depend on the MPRIS player API, not on the org.gpodder.player Dbus interface, which probably precedes MPRIS.

Agreed: enqueue-in-mediaplayer will be fine.

mpris-listener is a bridge between the org.gpodder.player DBus interface and MPRIS: it filters MPRIS events and properties  from media players and adapts them to the simpler gpodder player api. You changed it to use actions instead of dbus.

I admit my head hurts each time I try to understand what object is presented on the bus by which part of the code:

screenshot-2025-03-04-222806

the org.gpodder.player interface is implemented by the mpris-listener extension. `player.py` in gPodder core will listen to signals emitted by any object implementing this interface, in process or out of process.

By removing the dbus api we prevent another player from interfacing with gPodder like that. Since [panucci](https://github.com/gpodder/panucci) has not been worked on for 12 years I'd say it is fair.

What is your reasoning for removing the interface, though? Simplification?


Back to #1343: in the PR I've added a `Seek()` method (the first method) to the API, for gPodder to tell the player to seek to a position. the mpris extension will call the MPRIS equivalent on the right player (vlc, ...) using dbus. Note that it is a hack, with a hardcoded id, preventing an outside player from directly implementing it.

How would you implement calling the extension from gPodder without dbus, though? Add a `on_seek` extension endpoint?


> I could not rebase #1343 on top of this for unrelated reasons, but at least `enqueue_in_mediaplayer` as it is now works fine with this PR.

no need to.


> If we want to remove the dbus-python dependency from gPodder, then the `MPRISResumer` class needs to be ported to GDBus though.

yes

elelay:
What about `dbusproxy.py`?
Currently there are 2 bus names: org.gpodder (the `dbusproxy.py`) and org.gpodder.gpodder (the gui). Should/Could it be merged?

tpikonen:
I was looking at a very old version of the chapters UI MR when writing the previous comments. I see that the latest version has a lot more involved interface.

> What is your reasoning for removing the [org.gpodder.player] interface, though? Simplification?

The player interface duplicates MPRIS and is not in practice implemented by anything nowadays, except some code inside gPodder.

> How would you implement calling the extension from gPodder without dbus, though? Add a on_seek extension endpoint?

Extensions are not flexible enough to do player control, so it would make sense to replace the `MediaPlayerDBusReceiver` class (the `org.gpodder.player` interface) with an MPRIS receiver with Seek() etc. method support in core gPodder.

> What about dbusproxy.py?

Are there any users of the `org.gpodder.interface` and `org.gpodder.podcasts` interfaces?

I'd say there are 3 options on how to handle them if the rest of the gPodder is ported to GDBus:

 1. Just remove the interfaces. Some of the functionality is maybe still be available from the `org.gtk.Actions` interface autogenerated by GLib / GTK, but the actions are not designed to be a remote API.
 2. Port them to GDBus. This is at least a moderate amount of work.
 3. Make the interfaces optional. Move the fake-dbus module from `tools` to `src` and make sure gPodder works on all platforms without dbus-python. If real dbus-python is there, then the DBus interfaces are available.

Option 3 is probably not much work and is the least disruptive for now.

PR #1733, opened 2025-03-04
Focus lists using F keys.

auouymous:
This allows jumping directly to the channel, episode or progress lists using F1, F2, or F3, respectively.

Unless someone has a better choice of keys.


computerfreak105:
Hmm, maybe F1 is needed for help? If not, then I think F1 to F3 are wonderful key commands. Or you could use it with Ctrl. This would leave F1 free in case it’s needed for help topics ...

 �

 �

 �

From: auouymous ***@***.***> 
Sent: Tuesday, March 4, 2025 11:51 AM
To: gpodder/gpodder ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [gpodder/gpodder] Focus lists using F keys. (PR #1733)

 �

This allows jumping directly to the channel, episode or progress lists using F1, F2, or F3, respectively.

Unless someone has a better choice of keys.

  _____  


You can view, comment on, or merge this pull request online at:


 � �https://github.com/gpodder/gpodder/pull/1733


Commit Summary


*	b8b08b9 <https://github.com/gpodder/gpodder/pull/1733/commits/b8b08b971a00d220447ea77b4ba911663977a2f2>  Focus lists using F keys.


File Changes 


(1 <https://github.com/gpodder/gpodder/pull/1733/files>  �file)

*	M src/gpodder/gtkui/main.py <https://github.com/gpodder/gpodder/pull/1733/files#diff-b704897959982473c4ba54fe4c20429f6d501ad24e1908052e51e8c4992f4ec4>  (16) 


Patch Links:


*	https://github.com/gpodder/gpodder/pull/1733.patch
*	https://github.com/gpodder/gpodder/pull/1733.diff

auouymous:
> maybe F1 is needed for help?

Ya, F1 is generally used for help, maybe F2,F3,F4 could be used instead, and F1 could be made to open help.


> Or you could use it with Ctrl.

That was my original thought, however, there aren't any good keys for them. The "CDEFKLNPQRST" and "0123" ctrl combos are in use. It could use ctrl+shift combos with "CEP", but those are harder for some to type.

Alt combos for 'P' and 'E' are also taken, so those are out too. Alt+123 are available, but I'm not sure if they would work on Mac with its option key. It looks like gtk might send the command key as alt on Mac. I have changed the patch to use F1 for help, F234 to switch lists and added alt+123 as alternates. Could someone on Mac could let me know if the command key or option key works with 123?


Also, none of these keys would be discoverable in the client, only via the online help. We could add a section documenting these keys and the left and right arrow keys, which are not standard gtk movements. Along with the alt combos to open menus and other keyboard navigation tips.

computerfreak105:
Cool, that’s wonderful. But it might be an unnecessary doubble if we have F1-F4 and Alt+F1-F4 do the same. I’d be ok if it’s just F1 to F4. So we can keep it simple for everyone and don’t have to worry about windows and mac.

 �

I’d love to have also a shortcut to add new subscriptions. Maybe Ctlr+N? Or even simpler F5?

 �

Another thing which comes to my mind: Is it possible to write the name of a podcast/youtube channel and jump directly to it? Let’s say I’m in a list of hundreds of youtube channels and I would like to go to a channel named “sports and games”. So if I would write a few letters like “spo” the cursor would jump to this channel.

 �

Thanks for considering.

 �

 �

auouymous:
> But it might be an unnecessary doubble if we have F1-F4 and Alt+F1-F4 do the same.

Ya, I'm just tossing out options. I have thumb keys and layers so all keys are easy to reach without moving my hands. The F keys are further away for most users but easier to press with one finger. While the alt+123 are closer and similar difficulty as ctrl combos.


> I’d love to have also a shortcut to add new subscriptions.

Ctrl+L already works, as in the "location" of the feed.


> Another thing which comes to my mind: Is it possible to write the name of a podcast/youtube channel and jump directly to it? Let’s say I’m in a list of hundreds of youtube channels and I would like to go to a channel named “sports and games”. So if I would write a few letters like “spo” the cursor would jump to this channel.

Ctrl+shift+F opens a search box in channel list, typing filters the channel list, escape key closes the search box and restores the channel list. While a search is active, you can press shift+tab or the new F2 to jump to channel list and use arrows to move through it.

Ctrl+F does the same for episode list, and will filter based on title or description or channel. The episode list also supports special queries inside parentheses, such as `(since <= 7)` to only show episodes released in last 7 days.

Both searches will filter against ALL channels or episodes, no matter what the current view settings are showing.

PR #1734, opened 2025-03-04
fix dbusproxy.py to return str instead of bytes

elelay:
See https://dbus.freedesktop.org/doc/dbus-python/PY3PORT.html#bytes-vs-strings in python3 we can return str and it will be utf-8 encoded by the C library.

Reason for this commit is a bug in safe_first_line: converts str to bytes and then applies a str regex to it.

Tested by forcing a podcast description to "vive le vélo" using sqlite3 and then calling the get_podcasts() method on gPodder via d-spy and checking that the string displayed contains "vélo" correctly decoded.

### DIFF
--- a/src/gpodder/dbusproxy.py
+++ b/src/gpodder/dbusproxy.py
@@ -28,15 +28,7 @@ import gpodder
 from gpodder import util
 
 
-def safe_str(txt):
-    if txt:
-        return txt.encode()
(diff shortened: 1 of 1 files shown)


elelay:
Won't be necessary if #1731 includes gdbus port of dbusproxy.py

PR #1737, opened 2025-03-08
enqueue_in_mediaplayer: Port to GDBus

tpikonen:
Removes dbus-python usage from enqueue_in_mediaplayer extension and replaces it with GDBus from GLib.

Also makes some small fixes and adds mpv resumer.


elelay:
this is #1731 without removing the player interface and with extra goodies, right?

elelay:
Sorry for the noise: this is a different extension.

elelay:
merged, thanks :-)

PR #1739, opened 2025-03-13
Fix not being able to find absolute file paths from device POV

dconrad5:
With the creation of real absolute paths on devices, we need to strip the root "/" from the beginning of the path in order to check if the file still exists (for deletion in gPodder). In other words, change from the device POV to the computer POV.


dconrad5:
Here's a question, should we depend on the current state of the absolute paths setting, or should we try to infer whether they are absolute paths or relative paths based on the presence of a beginning slash? I could see an argument either way. I think I lean towards trusting the presence (or not) of a leading slash. Thoughts?

dconrad5:
> Here's a question, should we depend on the current state of the absolute paths setting, or should we try to infer whether they are absolute paths or relative paths based on the presence of a beginning slash? I could see an argument either way. I think I lean towards trusting the presence (or not) of a leading slash. Thoughts?
> 
> Edit: The more I think about it, the more I like the idea of trusting whether a leading slash exists here, because it could very well be that the setting changed between last sync and now.

I would like to detect when the setting has changed from last sync to this sync and keep the functionality working in that case, but the way that episode_dict[] (ref. ln253) is set up, we would need to convert the path from one style to the other for the episode keys to match, and that's proving to be a bit too confusing for my level of caring about this corner case at the moment. So I will leave it be.

dconrad5:
This is ready to review - I don't think the failure to build the windows executable is due to anything I have done here.

dconrad5:
Force pushed to trigger build

dconrad5:
> > Here's a question, should we depend on the current state of the absolute paths setting, or should we try to infer whether they are absolute paths or relative paths based on the presence of a beginning slash? I could see an argument either way. I think I lean towards trusting the presence (or not) of a leading slash. Thoughts?
> > Edit: The more I think about it, the more I like the idea of trusting whether a leading slash exists here, because it could very well be that the setting changed between last sync and now.
> 
> I would like to detect when the setting has changed from last sync to this sync and keep the functionality working in that case, but the way that episode_dict[] (ref. ln253) is set up, we would need to convert the path from one style to the other for the episode keys to match, and that's proving to be a bit too confusing for my level of caring about this corner case at the moment. So I will leave it be.

@elelay Would the corner case described here (if it makes any sense) bother you, or would it be okay? Handling it will require a little bit more complication than this current PR, and not sure if I should pursue it right now or not.

dconrad5:
> @dconrad5 sorry I don't have time at the moment to review the PR. Hopefully this week-end. Thanks

No worries, just was wondering if this is sufficient or not. Take your time!

dconrad5:
> > Here's a question, should we depend on the current state of the absolute paths setting, or should we try to infer whether they are absolute paths or relative paths based on the presence of a beginning slash? I could see an argument either way. I think I lean towards trusting the presence (or not) of a leading slash. Thoughts?
> > Edit: The more I think about it, the more I like the idea of trusting whether a leading slash exists here, because it could very well be that the setting changed between last sync and now.
> 
> I would like to detect when the setting has changed from last sync to this sync and keep the functionality working in that case, but the way that episode_dict[] (ref. ln253) is set up, we would need to convert the path from one style to the other for the episode keys to match, and that's proving to be a bit too confusing for my level of caring about this corner case at the moment. So I will leave it be.

This corner case is fixed with the newest commit push. There are still some settings which we cannot account for changing between last sync and this sync, but we can at least account for the changing of use_absolute_path.

Note that now the lookup in the episode dictionary now uses the file basename (with extension) rather than the whole path. I do not believe this will be an issue.

Please take a look and tell me if this is acceptable or not!

PR #1747, opened 2025-03-28
Add support for webkit2 version 4.1, Used for showing html podcast notes

### LINKED ISSUE #1740: Add support for webkit2 version 4.1
Please consider supporting webkit2 4.1 in addition or as a replacement to webkit2 4.0 to help with the transition from libsoup 2 to libsoup 3. 4.1 is a drop in replacement for 4.0 for projects that do not use libsoup with the only difference being the version of libsoup internally used by webkit2.

masoudd:
This PR adds a try..except chain to first try the 4.1 version if present. If not then it falls back to 4.0 version.

Closes #1740


elelay:
Merged, thanks

Issue #1662, opened 2024-10-01
Fix depcheck.py to reduce install size on Windows


auouymous:
The depcheck.py step was commented out in #1657 due to it failing. The build works without it and it is more important to have a working build to get a release out. However, it might be responsible for removing unused DLLs and the installation size might now require more space, see https://github.com/gpodder/gpodder/pull/1657#issuecomment-2385038001.

```
Traceback (most recent call last):                                                               
  File "C:/projects/gpodder/tools/win_installer/misc/depcheck.py", line 27                       
    repo.require(namespace, version, 0)                                                          
        Typelib file for namespace 'GIRepository', version '3.0' not found (0)                   
```

Using the following code snippets before calling depcheck.py did not work.

```
 mkdir -p lib/girepository-1.0/                                                     
 cp /mingw64/lib/girepository-1.0/GIRepository-2.0.typelib lib/girepository-1.0/    
```

```
export GI_TYPELIB_PATH="${MINGW_ROOT}"/lib/girepository-1.0                          
```

```
export GI_TYPELIB_PATH=/usr/local/lib/girepository-1.0                               
export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH                               
```

masoudd:
Seems like there is two parallel girepository. One is [gobject-introspection](https://gitlab.gnome.org/GNOME/gobject-introspection) and the other (seemingly newer one) is included in [glib](https://gitlab.gnome.org/GNOME/glib/-/tree/main/girepository).

|  | gobject-introspection | glib girepository |
|--------|--------|--------|
| latest version on msys2 mingw32 | 1.82 | 2.82 |
| dll file provided in /mingw32/bin/ | libgirepository-1.0-1.dll | libgirepository-2.0-0.dll |
| typelib file in /mingw32/lib/girepository-1.0/ | GIRepository-2.0.typelib | GIRepository-3.0.typelib |

I found [https://docs.gtk.org/girepository/migrating-gi.html](https://docs.gtk.org/girepository/migrating-gi.html)
> As part of this move, the API version has been bumped from 1.0 to 2.0, and some fairly straightforward API changes have been made. Note that the GIR version was previously 2.0, and is now 3.0 — so the new version of libgirepository uses girepository-2.0.pc and GIRepository-3.0.typelib.

Which suggests that this situation is temporary and the old one is supposed to be faded out.

[...]

PR #1748, opened 2025-03-31
Fix windows build, depcheck.py: Add try..except to try GIRepository 3.0 then 2.0

### LINKED ISSUE #1742: depcheck.py used for windows build is failing again
This script: `tools/win_installer/misc/depcheck.py`
From the [ci build that failed](https://ci.appveyor.com/project/elelay/gpodder/builds/51703184):

```
>>>> cleanup_after <<<<
rm: cannot remove '/home/appveyor/_gpodder_build_root/mingw32/bin/__pycache__': Is a directory
  File "C:/projects/gpodder/tools/win_installer/misc/depcheck.py", line 23, in <module>
    gi.require_version("GIRepository", "2.0")  # isort:skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/msys64/home/appveyor/_gpodder_build_root/mingw32/lib/python3.12/site-packages/gi/__init__.py", line 132, in require_version
    raise ValueError('Namespace %s not available for version %s' %
ValueError: Namespace GIRepository not available for version 2.0
Command exited with code 1
```
From `NEWS` in [pygobject git repo](https://gitlab.gnome.org/GNOME/pygobject/-/blob/main/NEWS?ref_type=heads):
```
3.52.0 - 2025-03-09
-------------------

From this version onwards girepository-2.0 is used for introspection.
```

I believe the temporary situation I mentioned in [here](https://github.com/gpodder/gpodder/issues/1662#issuecomment-2521570019) is resolved,
`pygobject` has switched to the right column of the table in my linked comment,
And we need to switch to `gi.require_version("GIRepository", "3.0")`

masoudd:
Hopefully this fixes the windows build ci.

Due to the differences in the API between the versions I have added an if..else to call the right method for the version '3.0' vs '2.0'.

Any input is appreciated.

See also: #1662
Closes #1742

### DIFF
--- a/tools/win_installer/misc/depcheck.py
+++ b/tools/win_installer/misc/depcheck.py
@@ -20,7 +20,15 @@ from functools import cache
 from multiprocessing import Process, Queue
 
 import gi  # isort:skip
-gi.require_version("GIRepository", "2.0")  # isort:skip
+
+girepository_version = 3
+try:
+    gi.require_version("GIRepository", "3.0")  # isort:skip
+    girepository_version = 3
+except ValueError as e:
+    girepository_version = 2
+    gi.require_version("GIRepository", "2.0")  # isort:skip
+
 from gi.repository import GIRepository  # isort:skip
 
 
@@ -28,7 +36,18 @@ def _get_shared_libraries(q, namespace, version):
     repo = GIRepository.Repository()
     try:
         repo.require(namespace, version, 0)
-        lib = repo.get_shared_library(namespace)
+        if girepository_version == 3:
+            ret = repo.get_shared_libraries(namespace)
+            if ret:
+                lib = ret[0]
+            else:
+                lib = None
+        elif girepository_version == 2:
+            lib = repo.get_shared_library(namespace)
+        else:
+            lib = None
+            logging.error("GIRepository version is not 3 or 2")
+
         q.put(lib)
     except Exception as e:
         logging.exception(e)
@@ -148,6 +167,7 @@ def main(argv):
     if "--debug" in argv[1:]:
         logging.getLogger().setLevel(logging.DEBUG)
 
+    logging.debug("GIRepository being used: %s", girepository_version)
     libs = get_things_to_delete(sys.prefix)
 
     if "--delete" in argv[1:]:
--- a/tools/win_installer/misc/depcheck.py
+++ b/tools/win_installer/misc/depcheck.py
@@ -33,6 +33,7 @@ from gi.repository import GIRepository  # isort:skip
 
 
 def _get_shared_libraries(q, namespace, version, loglevel=logging.WARNING):
+    """put a list of libraries into q, regardless of girepository_version."""
     import multiprocessing
     logger = multiprocessing.log_to_stderr(level=loglevel)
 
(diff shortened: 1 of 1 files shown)


tools/win_installer/misc/depcheck.py
```diff
@@ -20,15 +20,34 @@
...
 
 def _get_shared_libraries(q, namespace, version):
     repo = GIRepository.Repository()
     try:
         repo.require(namespace, version, 0)
-        lib = repo.get_shared_library(namespace)
+        if girepository_version == 3:
+            ret = repo.get_shared_libraries(namespace)
+            if ret:
+                lib = ret[0]
```
elelay, line 42: I think you have to return all libraries.
[get_shared_library](https://lazka.github.io/pgi-docs/GIRepository-2.0/classes/Repository.html#GIRepository.Repository.get_shared_library) returns a single string with paths separated with ','.
[get_shared_libraries](https://docs.gtk.org/girepository/method.Repository.get_shared_libraries.html) returns the list already split.

tools/win_installer/misc/depcheck.py
```diff
@@ -20,15 +20,34 @@
...
+            else:
+                lib = None
+        elif girepository_version == 2:
+            lib = repo.get_shared_library(namespace)
```
elelay, line 46: Maybe you can do the `lib.lower().split(",")` here:
```suggestion
            lib = repo.get_shared_library(namespace)
            if lib:
              lib = lib.split(","))
```

and rename lib as libs.

elelay (changes requested):
you will need to fully adapt to the changed interface.
I propose returning libs from _get_shared_libraries

tools/win_installer/misc/depcheck.py
masoudd: Yes. Sorry I don't know why I assumed it will always be a list of length 1

masoudd:
Now `_get_shared_libraries` always puts a list in the queue and `get_shared_libraries` returns a list

elelay:
Perfect!
Merged, thanks.

auouymous:
@masoudd Should https://github.com/gpodder/gpodder/blob/master/tools/win_installer/misc/depcheck.py#L24 be setting it to `3` or something else that would trigger the `else` branch at https://github.com/gpodder/gpodder/blob/master/tools/win_installer/misc/depcheck.py#L53 ? And if 3.0 isn't found wouldn't it always set girepository_version to 2, and then throw an uncaught exception if 2.0 isn't found?

masoudd:
> @masoudd Should https://github.com/gpodder/gpodder/blob/master/tools/win_installer/misc/depcheck.py#L24 be setting it to `3` or something else that would trigger the `else` branch at https://github.com/gpodder/gpodder/blob/master/tools/win_installer/misc/depcheck.py#L53 ? And if 3.0 isn't found wouldn't it always set girepository_version to 2, and then throw an uncaught exception if 2.0 isn't found?

Yes. I'm sorry I'm not sure why I missed that. It should be initialized to something other than 2 or 3. maybe 0. And it should assign to it after a successful call to `require_version()`, not before.
Like:
```
girepository_version = 0
try:
    gi.require_version("GIRepository", "3.0")  # isort:skip
    girepository_version = 3
except ValueError as e:
    gi.require_version("GIRepository", "2.0")  # isort:skip
    girepository_version = 2
    
```

auouymous:
Okay, hopefully #1758 builds fine, does it look good to you?

PR #1757, opened 2025-04-13
added or fixed translations

pieska:


po/de.po
```diff
@@ -2637,11 +2638,11 @@ msgstr "Falscher Benutzername/Passwort"
 msgid ""
 "Proxies given in environment variables (will be used only if proxy above is "
 "disabled):"
-msgstr ""
+msgstr "Proxy-Typ:"
```
auouymous, line 2641: This is changing line 2640 which is for `msgid "Proxies given in environment variables (will be used only if proxy above is "disabled):"`. The correct line should be 2618.

auouymous (changes requested):

po/de.po
pieska: oops, you're right, fixed it.

pieska:
I noticed some texts in the context menu not getting tanslated, for instance "Update channel":
```
$ grep -A1 "Update channel" po/de.po
msgid "Update channel"
msgstr "Aktualisierungsintervall:"
$ grep "Update channel" share/gpodder/ui/gtk/menus.ui 
        <attribute name="label" translatable="yes">Update channel</attribute>
        <attribute name="label" translatable="yes">Update channel</attribute>
```

"Open download folder" is getting translated, so it works for some entries and for some not.

auouymous:
Thanks for the translations.

auouymous:
I don't know what the fuzzy msgid is or if it has anything to do with it, since "Open download folder" does not have fuzzy. @elelay Any ideas?

```
#: share/gpodder/ui/gtk/menus.ui.h:53
#, fuzzy
#| msgid "Update interval:"
msgid "Update channel"
msgstr "Aktualisierungsintervall:"
```

PR #1758, opened 2025-04-14
Fix depcheck GIRepository versioning issue introduced in #1748.

auouymous:


masoudd:
This looks good to me. Sorry for missing this edge case in the previous PR

elelay:
Thanks @masoudd :+1:

elelay:
And thanks @auouymous to catch this. It produces a better error message by catching the ValueError

PR #1760, opened 2025-04-18
Set published timestamp on downloaded files.

### LINKED ISSUE #1745: New Feed Download - Keep Original File Date
When a new feed is added and the initial download of past episodes starts, all files receive the current date, not the original date of the file. For feeds that do not use episode numbers in their titles, you cannot tell the original order of the cast.

auouymous:
Downloaded episode files receive the current time on disk, this changes them to the published timestamp, if available. Episodes downloaded with yt-dlp already do this.

Fixes #1745.

### DIFF
--- a/src/gpodder/download.py
+++ b/src/gpodder/download.py
@@ -381,6 +381,13 @@ class DefaultDownload(CustomDownload):
                             url)
                     continue
                 raise
+
+        # Podcastparser defaults published to zero if pubDate is not specified in feed,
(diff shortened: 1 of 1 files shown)


elelay:
would you like to enforce max = current time so that files don't appear modified in the future?

auouymous:
Oh ya, done.

elelay:
Merged, thanks :-)

PR #1762, opened 2025-04-20
Preferences: add buttons to open system and user extensions folders

dconrad5:
Adds buttons in the preferences window to open both the system extensions directory and the user extensions directory.

I would like to make these hidden except when the Extensions tab is open. They also probably need margins adjusted, etc. to make them look correct.

Thoughts?

Screenshot 2025-04-20 at 11 11 14 AM


thp:
> I would like to make these hidden except when the Extensions tab is open. They also probably need margins adjusted, etc. to make them look correct.

Why not put it inside the extensions tab? :)

dconrad5:
> > I would like to make these hidden except when the Extensions tab is open. They also probably need margins adjusted, etc. to make them look correct.
> 
> Why not put it inside the extensions tab? :)

I certainly could, may need to make the extension pane shorter vertically is all. If that would be acceptable it should be pretty easy.

dconrad5:
> > > I would like to make these hidden except when the Extensions tab is open. They also probably need margins adjusted, etc. to make them look correct.
> > 
> > 
> > Why not put it inside the extensions tab? :)
> 
> I certainly could, may need to make the extension pane shorter vertically is all. If that would be acceptable it should be pretty easy.

Done:

Screenshot 2025-04-20 at 2 32 33 PM

PR #1765, opened 2025-05-01
enqueue_in_mediaplayer: Also test unquoted URLs from metadata

tpikonen:
Some players (mpv-mpris) do incomplete URL quoting, so also try to match unquoted player URLs when receiving metadata.


tpikonen:
Rebased to master, fixing the conflict.

PR #1783, opened 2025-06-26
Improve DownloadTask.DONE status icon

tpikonen:
Set DownloadTask.DONE icon to `selection-mode-symbolic`. This icon is easier to distinguish from the `go-down` icon used for the DOWNLOADING state.

Also add some horizontal padding to the icon.


auouymous:
I like the padding, not so much the icon. I assume something is forcing the symbolic variant of go-down which lost its trail of dots and would definitely be a problem. Do you know of another icon that could replace go-down since the symbolic variant doesn't look like its downloading? If not, I am fine with this change.

Issue #1763, opened 2025-04-21
Improvement: Reflect extension run order in Preferences


dconrad5:
~With PR #1698 merged, the Preferences Extensions tab should probably be made to print the extensions in priority order (within their respective categories). Could also print out the numeric priority.~

After discussions in #1791, rather than sorting by priority, we will just display it in one or two places.

Just putting this here to write it down.

PR #1791, opened 2025-08-20
Issue #1763: Preferences: List extensions by priority within categories

dconrad5:
As title states. Sorted by category, then priority, then metadata title.

Numerical priority is not displayed to the user - if they need to know it, they'll see the filenames anyway.

### DIFF
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -507,7 +507,7 @@ class gPodderPreferences(BuilderWidget):
 
         def key_func(pair):
             category, container = pair
-            return (category, container.metadata.title)
+            return (category, container.priority, container.metadata.title)
 
         def convert(extensions):
             for container in extensions:


elelay:
I understand and the code is fine.
I find it strange that items are not sorted by visible title anymore, though.
Please other people comment for or against it :-)

auouymous:
It makes it harder to find an extension. Displaying the priority before title or description would help to make it not look like we forgot to sort the entries. However, simply displaying a number could be confusing, so it would need to be labeled, which puts too much info before the title and description. Placing it after can result in priority being clipped for some extensions on narrow windows. I'm not sure how to properly display it, but sorting might be better done with a checkbox that defaults to off. Maybe the checkbox displays priority number before description `%s\n<small>%d — %s</small>`?

We should also display priority in the `extension info` dialog opened by right clicking an extension.

dconrad5:
> I understand and the code is fine. I find it strange that items are not sorted by visible title anymore, though. Please other people comment for or against it :-)

> It makes it harder to find an extension. Displaying the priority before title or description would help to make it not look like we forgot to sort the entries. However, simply displaying a number could be confusing, so it would need to be labeled, which puts too much info before the title and description. Placing it after can result in priority being clipped for some extensions on narrow windows. I'm not sure how to properly display it, but sorting might be better done with a checkbox that defaults to off. Maybe the checkbox displays priority number before description `%s\n<small>%d — %s</small>`?
> 
> We should also display priority in the `extension info` dialog opened by right clicking an extension.

I see - to be honest, before working on this it never even really occurred to me as a user that they were sorted in the list alphabetically.

I'll see about maybe making it a configurable setting (sort by alphabetical vs. sort by priority) like you say. Or... what if we left the sorting as alphabetical, but _only_ display the priority in the extension info dialog?

tpikonen:
> what if we left the sorting as alphabetical, but only display the priority in the extension info dialog?

I'd say this is reasonable. Also, maybe add a 'Prio: xx' prefix to the the extension description in the list. Configurable sorting for the extension list is IMO overkill.

Anyway, that's my 0.02 EUR towards building this bikeshed :)

dconrad5:
> > what if we left the sorting as alphabetical, but only display the priority in the extension info dialog?
> 
> I'd say this is reasonable. Also, maybe add a 'Prio: xx' prefix to the the extension description in the list. Configurable sorting for the extension list is IMO overkill.
> 
> Anyway, that's my 0.02 EUR towards building this bikeshed :)

> > if we left the sorting as alphabetical, but only display the priority in the extension info dialog?
> 
> yes, please. I would be against an option for sorting by priority

Sounds good - that's certainly easier in any case.

PR #1792, opened 2025-08-24
Port dbusproxy to gdbus

elelay:
This ports the podcasts and gui objects to glib dbus:

podcasts:
- `get_episode_title(ep_url)`
- `get_episodes(podcast_url)`
- `get_podcasts()`
- `play_or_download_episode()`

gui:
- `show_gui_window()`
- `offer_new_episodes(podcast_urls)`
- `subscribe_to_url(url)`
- `mark_episode_played(filename)`

### DIFF
--- a/src/gpodder/dbusproxy.py
+++ b/src/gpodder/dbusproxy.py
@@ -139,7 +139,7 @@ class DBusPodcastsProxy:
             title = episode.title
             url = episode.url
             description = first_line(episode._text_description)
-            filename = or_empty_str(episode.download_filename)  # None when not downloaded
+            filename = or_empty_str(episode.local_filename(False, check_only=True))  # None when not downloaded
             file_type = episode.file_type()
             is_new = (episode.state == gpodder.STATE_NORMAL and episode.is_new)
             is_downloaded = episode.was_downloaded(and_exists=True)
--- a/src/gpodder/dbusproxy.py
+++ b/src/gpodder/dbusproxy.py
@@ -67,6 +67,21 @@ xml = f"""
           <arg name='message' type='s' direction='out'/>
       </method>
   </interface>
+  <interface name='{gpodder.dbus_interface}'>
+      <method name='show_gui_window'>
+      </method>
+      <method name='offer_new_episodes'>
+          <arg name='channel_urls' type='as' direction='in'/>
+          <arg name='new_episodes' type='b' direction='out'/>
+      </method>
+      <method name='subscribe_to_url'>
+          <arg name='url' type='s' direction='in'/>
+      </method>
+      <method name='mark_episode_played'>
+          <arg name='filename' type='s' direction='in'/>
+          <arg name='success' type='b' direction='out'/>
+      </method>
+  </interface>
 </node>
 """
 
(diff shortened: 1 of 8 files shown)


tpikonen (changes requested):
Seems to work, except for `offer_new_episodes` method.

gPodder has had two bus names, `org.gpodder` from dbusproxy and `org.gpodder.gpodder` derived from the app-id by GLib and GTK. Unless you want to be 100% backwards compatible, I would suggest using just `org.gpodder.gpodder` for everything. The bus name is defined at least in `src/gpodder/__init__.py` and `share/dbus-1/services/org.gpodder.service.in`.

The `/gui` and `/podcasts` bus objects are also redundant and could be just added to the default `/org/gpodder/gpodder` object, since the methods are under their own interfaces.

Also, maybe squash the `remove -gdbus suffix from GDBus implementation` commit?

src/gpodder/dbusproxy.py
```diff
@@ -36,27 +38,101 @@ def first_line(txt):
         return lines[0]
 
 
-class DBusPodcastsProxy(dbus.service.Object):
+def or_empty_str(txt):
+    """can't marshal the None value as a string, so use this wrapper to be safe"""
```
tpikonen, line 42: This style of docstring triggers the flake8 linter (D400 and D403). Docstrings should be capitalized and end with a period.

src/gpodder/gtkui/main.py
```diff
@@ -4033,12 +4024,11 @@ def iconify_main_window(self):
         if not self.is_iconified():
             self.gPodder.iconify()
 
-    @dbus.service.method(gpodder.dbus_interface)
     def show_gui_window(self):
+        """for dbusproxy only"""
```
tpikonen, line 4028: Docstring formatting is off also here.

src/gpodder/gtkui/main.py
```diff
@@ -4052,22 +4042,14 @@ def subscribe_to_url(self, url):
                 add_podcast_list=self.add_podcast_list,
                 preset_url=url)
 
-    @dbus.service.method(gpodder.dbus_interface)
-    def mark_episode_played(self, filename):
-        if filename is None:
+    def mark_episode_played(self, episode):
+        """for dbusproxy only at the moment"""
```
tpikonen, line 4046: Docstring formatting.

elelay:
> Seems to work, except for `offer_new_episodes` method.

sorry. Now all methods work.

> gPodder has had two bus names, `org.gpodder` from dbusproxy and `org.gpodder.gpodder` derived from the app-id by GLib and GTK. Unless you want to be 100% backwards compatible, I would suggest using just `org.gpodder.gpodder` for everything. The bus name is defined at least in `src/gpodder/__init__.py` and `share/dbus-1/services/org.gpodder.service.in`.

ok

> The `/gui` and `/podcasts` bus objects are also redundant and could be just added to the default `/org/gpodder/gpodder` object, since the methods are under their own interfaces.
> 

ok

> Also, maybe squash the `remove -gdbus suffix from GDBus implementation` commit?

Let's squash after review.

PR #1793, opened 2025-08-25
add flake8 linter

elelay:
currently non blocking.

### DIFF
--- a/.github/workflows/linttest.yml
+++ b/.github/workflows/linttest.yml
@@ -24,7 +24,7 @@ jobs:
         run: |
           sudo apt-get update -q
           sudo apt-get install intltool desktop-file-utils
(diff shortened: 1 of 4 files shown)


tpikonen:
Currently flake8 produces almost 200 warnings, which is a bit too much to be useful. Some the warnings are easy to fix and some could be ignored in `setup.cfg` though.

The most common warning seems to `B901 blind except: statement` (and B902), which are maybe hard to fix without introducing unhandled exceptions. Maybe we should just ignore them?

elelay:
yes, I would ignore B901, but it's E722 on my machine

    # flake8 --version
    7.2.0 (mccabe: 0.7.0, pycodestyle: 2.14.0, pyflakes: 3.3.2) CPython 3.13.5 on Linux

elelay:
so: do we merge as-is?

PR #1794, opened 2025-08-26
main: Fix 'Open download folder' on non-downloaded episodes

tpikonen:
When the selected episode was not downloaded, None was submitted as arg to pathlib.Path, resulting in an unhandled exception in on_open_episode_download_folder().

Raise GError when episode file does not exist, the folder is then opened with util.gui_open() in the exception handler.


elelay:
merged, thanks :-)

PR #1795, opened 2025-08-26
Port mpris-listener to gdbus; remove player.py dbus interface; player.PlayerControl in registry.player_control

elelay:
Builds on #1792.
Adopts almost all changes from #1731 except gpodder.player deletion and using gtkui/main actions.

It now defines `PlayerControl(ObservableService)` to be implemented by extensions providing media player monitoring
and control.
At this time only mpris-listener.py is envisioned to fill that role.

PlayerControl has as `seek(file_uri, position)` method that can be used to jump to a chapter.

FIXME:

- [x] PlayerController should listen to Extension (un)load to re-register for events
- [x] can't be integrated in `./bin/gpo` because gpo doesn't run the glib event loop. This has to be refactored...
- ~~add a mpris-listener extension setting to not register to player_control, so that another implementation can be registered~~
  (PlayerControl and PlayerController merged => the mpris-listener can be disabled if not wanted)

### DIFF
--- a/share/gpodder/extensions/mpris-listener.py
+++ b/share/gpodder/extensions/mpris-listener.py
@@ -61,6 +61,7 @@ class CurrentTrackTracker(object):
     of a single MPRIS-enabled player (an running instance of vlc, ...).
     player_id is the identity of the MPRIS-enable player on d-bus.
     """
+
     def __init__(self, notifier, player_id):
         self.uri = None
         self.length = None
@@ -104,8 +105,7 @@ class CurrentTrackTracker(object):
         return False
 
     def update(self, **kwargs):
-        """ callback from MPRISDBusReceiver """
-
+        """Call this from MPRISDBusReceiver on events."""
         player_id = kwargs.pop('player_id', self.player_id)
         if self.player_id != player_id:
             raise Exception("CurrentTrackTracker[%s] received update for %s" % (self.player_id, player_id))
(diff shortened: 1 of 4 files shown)


elelay:
Sorry, I broke the code during refactoring in multiple commits. It's now fixed.

elelay:
PlayerController is a bit cleaner now.

About `PlayerController`: I find it good to have episode update and mypgo_client calling out of `gtkui/main.py`.

I'm still not happy about the `on_episode_status_changed()` callback, because it's crafted specially for `gtkui/main.py` to refresh its buttons and channels.

Also PlayerController refines the playback signals for potentially different extensions (finding the episode, keeping track of currently playing ones) but it doesn't emit any event (except  `on_episode_status_changed()`). This can make it less useful than intended for something like the playbar: the playbar will needs to register for PlayerControl signals, and then do its book-keeping itself to update the ui. So `PlayerController.currently_playing` would not be used.

Maybe `PlayerController` should be absorbed into `PlayerControl` (update episode, mygpo upload, keep a dict of currently playing) and `mpris-listener.py` (filter some events).

Then, the `PlayerControl` callbacks could include the episode, already updated. This way `gtkui/main.py` can register a lambda to call its `episode_list_status_changed` on Playback events. `gtkui/main.py` would observe `registry.player_control` and register its mygpo_client to the new PlayerControl when it appears.

Would you prefer code without `PlayerController`?

tpikonen:
I unfortunately do not have the time to really think about this right now, but a couple of random thoughts:

I found the two classes in player.py somewhat surprising / complicated, so probably a single abstract class deriving from `ObservableService` would be enough.

A minor problem, but the name `PlayerControl` is a bit misleading, since it's mostly about listening to start and stop events (OK, there's now the `seek` method). Maybe `PlayerInterface` would be better?

elelay:
@tpikonen is it ok to merge?

PR #1799, opened 2025-09-21
makefile: Don't print all files on codespell invovation

tpikonen:
The `codespell` invocation in `make lint` currently prints out all gPodder files known to git. Reduce the noise by rewriting the command in makefile.


elelay:
It was indeed a bit noisy.
Merged, thanks!

PR #1805, opened 2025-09-24
youtube-dl: Use youtube-dl config file to set initial opts

### LINKED ISSUE #1307: Modify the yt-dlp command
I'm wondering whether it is possible to modify the yt-dlp command that is run when downloading podcasts with the Youtube-dl extension. Right now one can only modify the audio format that is being downloaded. Having the ability to add arguments to yt-dlp would be incredibly useful for applications like Sponsorblock.

tpikonen:
Use youtube_dl.parse_options() to get an initial _ydl_opts dict read from the youtube-dl / yt-dlp config file and update that with our options.

The dict values which are dicts or lists are updated or extended to preserve e.g. the postprocessors set in the config file.

Seems to work, but needs more testing.

Fixes #1307.


share/gpodder/extensions/10_youtube-dl.py
```diff
@@ -357,10 +359,22 @@ def fetch_info(self, url, tempname, reporthook):
 
         # Need the proxy_url from src/gpodder/config.py:get_proxies_from_config()
         if gpodder.config._proxies:
-            opts['proxy'] = gpodder.config._proxies['http']
-            logger.debug(f"Setting proxy from network setting proxy: {opts['proxy']}")
+            fetch_opts['proxy'] = gpodder.config._proxies['http']
+            logger.debug(f"Setting proxy from network setting proxy: {fetch_opts['proxy']}")
+
+        opts = self._ydl_opts.copy()
+        # Update or extend existing dict and list values
```
auouymous, line 366: Why is this needed when the other two sites could get away with something like `opts.update(fetch_opts)`?

auouymous (changes requested):

share/gpodder/extensions/10_youtube-dl.py
auouymous: Thank you, now they use the same syntax as the conditional keys. We can now move the `cachedir` lines together and group them between full line comments before keys instead of EOL comments after some of them.

Give me a few days to test it.

share/gpodder/extensions/10_youtube-dl.py
```diff
@@ -298,10 +298,10 @@ def __init__(self, gpodder_config, my_config, force=False):
...
         cachedir = os.path.join(gpodder.home, 'youtube-dl')
         os.makedirs(cachedir, exist_ok=True)
-        self._ydl_opts = {
-            'cachedir': cachedir,
-            'noprogress': True,  # prevent progress bar from appearing in console
-        }
+        # Read options from youtube-dl config file
+        _parser, _opts, _urls, self._ydl_opts = youtube_dl.parse_options()
+        self._ydl_opts['cachedir'] = cachedir
+        self._ydl_opts['noprogress'] = True,  # prevent progress bar from appearing in console
```
auouymous, line 304: The trailing comma creates a list value.

share/gpodder/extensions/10_youtube-dl.py
```diff
@@ -342,25 +342,23 @@ def add_format(self, gpodder_config, opts, fallback=None):
...
-            'writesubtitles': subs,
-            'subtitleslangs': ['all'] if subs else [],
-            'postprocessors': [{'key': 'FFmpegEmbedSubtitle'}] if subs else [],
-        }
+        opts = self._ydl_opts.copy()
+        opts['paths'].update({'home': os.path.dirname(tempname)})
+        # Postprocessing in yt-dlp breaks without ext
+        opts['outtmpl'] = ({'default': os.path.basename(tempname) + '.%(ext)s'} if program_name == 'yt-dlp'
+                           else tempname + '.%(ext)s')
+        opts['nopart'] = True,  # don't append .part (already .partial)
```
auouymous, line 350: One more missed comma.

auouymous (changes requested):

tpikonen:
> Do you think https://github.com/gpodder/gpodder/pull/1805 should have a checkbox to disable config loading

So ticking this checkbox in the youtube-dl tab would enable loading the yt-dlp configuration file? Should it be set to off by default to be backwards compatible?

auouymous:
Yes and yes. Anyone who checks the box hopefully knows how to uncheck it if something breaks.

tpikonen:
The last commit adds a checkbox toggling the config file reading in the preferences tab.

PR #1807, opened 2025-10-01
fix #1800 remove_finished() in episodeselector.py called with wrong args

### LINKED ISSUE #1800: remove_finished() in episodeselector.py called with wrong args
#1792 removed `episode_new_status_changed()` in main.py and replaced it with  `episode_list_status_changed()` in `gPodderEpisodeSelector` constructor. The first function takes a list of URLs and the second takes a list of episodes, so things break.

This can be fixed in the only call site of `remove_finished()` in episodeselector.py by calling the function with a list of episodes.

elelay:
following refactoring in #1792

Exception occurs when marking episodes as old from the episode selector dialog after refresh


tpikonen:
Looks good.

Issue #1675, opened 2024-12-01
Another issue with "Retry-After" header: impossible to update podcast until the Retry-After expires


NannoSilver:
I am running gPodder 3.11.4 (2023-10-11) portable version, downloaded from https://portableapps.com/apps/internet/gpodder_portable

This is another issue on how gPodder is handling the "Retry-After" header, similar to the issue reported at https://github.com/gpodder/gpodder/issues/1661

While trying to update a podcast, gPodder got from the server "Error 503" with a "Retry-After" header.
After that, gPodder disabled the "Update podcast" in the right menu until the Retry-After time expired. That was to expect, but the problem is that gPodder disabled the "Update podcast" for all podcasts, even those hosted in different websites, and not only for the one that got the Error 503 and the Retry-After header.

![gpodder B](https://github.com/user-attachments/assets/30446f8b-095e-4376-84de-44cd06b2fa29)

![gpodder A](https://github.com/user-attachments/assets/70ac9e31-8ff4-4dc8-a06e-6acb92411b9e)

elelay:
This makes sense, since this iteration of the update loop has not finished, but it is bad ux.

PR #1808, opened 2025-10-01
Fix #1627 #1661 #1675 by storing cache-control and not-before from server

### LINKED ISSUE #1627: Does gPodder implement Last-Modified / If-Modified-Since (conditional GET)?
I see one client in my logs downloading the podcast feed apparently unconditionally every 20 minutes.  New content is added to the feed less than once a month, so several MB of traffic and hundreds of polls to achieve that is a little excessive IMHO.

    [29/Apr/2024:13:24:54 +0000] "GET /rss/podcast.rss HTTP/1.1" 200 12856 "-" "gPodder/3.11.1 (+http://gpodder.org/) Linux"
    [29/Apr/2024:13:44:55 +0000] "GET /rss/podcast.rss HTTP/1.1" 200 12856 "-" "gPodder/3.11.1 (+http://gpodder.org/) Linux"
    [29/Apr/2024:14:05:12 +0000] "GET /rss/podcast.rss HTTP/1.1" 200 12856 "-" "gPodder/3.11.1 (+http://gpodder.org/) Linux"
    [29/Apr/2024:14:25:01 +0000] "GET /rss/podcast.rss HTTP/1.1" 200 12856 "-" "gPodder/3.11.1 (+http://gpodder.org/) Linux"

I have not seen a 304 (Not Modified) response to it for over a week, which may be to do with my turning off ETags to work around a 2008-vintage Apache server bug that I discovered interefering with other clients' conditional requests.  (They are now working better.)

Do you support If-Modified-Since, and if not could you please add support for it?

This traffic would also be mitigated if gPodder would use the Cache-Control or Expires headers that I supply (of 4h+), or indeed the skipHours in the RSS feed itself!

Rgds

Damon

### LINKED ISSUE #1661: Impossible to cancel audio download if server send "Retry-After" header

elelay:
Fixes #1627 (Support Cache-Control/Expires on feed refresh)
Fixes #1661 #1675 (Issues with letting requests handle the retry-after)

WARNING: requires a DB schema change. Don't run on your default GPODDER_HOME!

I've checked with example feed from #1627: indeed the feed doesn't refresh a few minutes after first time.

Adds a force-refresh action, that is also useful after the user raises the episode limit or other situations where we previously would recommend removing and re-adding the feed. It is hidden as Shift+Click and Ctrl-Shift-R, though.


elelay:
Regarding

> surface the "not updated due to cache-control" to the feed's ui?

I wen't ahead and added an extra icon on the cover in the channel + a message to explain.

elelay:
Episodes:

- [x] tested and fixed episode download: saves the not_before time and surfaces the error message in the UI (no change)
- [x] when asking to download an episode with not_before in the future, the download is paused
- [x] it is possible to force-download by right-clicking the task

src/gpodder/gtkui/model.py
```diff
@@ -589,6 +590,7 @@ def __init__(self, cover_downloader):
         self.icon_theme = Gtk.IconTheme.get_default()
         self.ICON_DISABLED = 'media-playback-pause'
         self.ICON_ERROR = 'dialog-warning'
+        self.ICON_NOT_REFRESHED = 'power-profile-power-saver-symbolic'  # weather-clear-night-symbolic preferences-system-time-symbolic
```
elelay, line 593: alternate viable icons in comment

PR #1811, opened 2025-10-03
Remove payment links pointing to flattr.com from extensions

### LINKED ISSUE #1798: Remove Flattr links from extensions
[Flattr](https://en.wikipedia.org/wiki/Flattr) has been shut down for a while now, so I think we should remove the broken links to it from extension sources.

The payment links only contain links to Flattr and nowhere else, so there would be nothing using the 'Support the author' functionality after removing flattr links.

tpikonen:
Flattr has been shut down.

Fixes #1798.


thp:
@brot If you have updated donation links, feel free to chime in.

PR #1813, opened 2025-10-04
Tweak Extensions categories

dconrad5:
I think the extensions categories list could use a little adjustment -

Currently, this PR adds "Device Sync", with:

- Rockbox Cover Art Sync
- Run A Command On Sync (Perhaps this could be renamed "Run A Command After Sync"?)

I also propose (but have not done yet in this PR) to add a "Download" counterpart to "Post Download", and put in it:

- Filter Episodes
- youtube-dl
- (maybe) Subtitle Downloader for TED Talks?

Thirdly (and finally), I wonder if "Run A Command On Download" should be renamed to "Run A Command After Download". This has not been done yet in this PR either.

Opinions?


dconrad5:
Done, please take a look.

PR #1819, opened 2025-10-18
Import AppIndicator3 with a version to avoid warning

tpikonen:
Currently enabling the `ubuntu_appindicator` extension results in this warning:

```
PyGIWarning: AppIndicator3 was imported without specifying a version first. Use gi.require_version('AppIndicator3', '0.1') before import to ensure that the right version gets loaded.
```

This MR does as instructed.


elelay:
Merged, thanks :-)

PR #1821, opened 2025-10-23
logger.Error -> logger.error  in 10_youtube-dl.py

masoudd:
The logger object methods are all lowercase.
Sorry for not catching this earlier. I will make sure to be more thorough in the future.


auouymous:
Thanks for the fix.

PR #1829, opened 2025-12-22
two-way-sync: fix for on-device absolute paths

dconrad5:
This PR fixes two-way-sync when absolute paths are used on-device. This also adds a note describing the choice to allow behavior to be quirky when the use_absolute_path setting is changed.

When the use_absolute_path setting is changed, the deleted episodes will simply be re-added to the device because a match cannot be made in the episode dictionary. This quirk is acceptable because the alternate may be too broad of matching. Add a note stating as much.


dconrad5:
Bumping this. It's a simple enough fix, is there any chance it can get merged?

elelay:
Merged, thanks.
Sorry it fell into the cracks...

PR #1831, opened 2025-12-25
model: Fix docstring capitalization etc. Fixes flake8 D400 and D403

tpikonen:
After these changes gPodder is flake8-clean, and `make lint` could be made to fail on failing checks.


auouymous:
Thanks for the fix, and all of your other flake8 fixes.

PR #1832, opened 2025-12-25
Extensions: Automatically show extension info in preferences tab, add some extensions preferences

dconrad5:
Automatically show extension info in the preferences tab for an extension if the extension provides a widget for preferences. This requires adding a container return value to on_preferences(), which this PR does for all extensions which have a preferences tab.

Also add several extension preferences tabs.

Example:

image


elelay (commented):
Code looks good, thanks.
Untested, though...

share/gpodder/extensions/20_audio_converter.py
```diff
@@ -140,3 +145,55 @@ def _convert_episodes(self, episodes):
         # which one is being converted and nothing prevents from clicking convert twice.
         for episode in episodes:
             self._convert_episode(episode)
+
+    def on_context_menu_toggled(self, widget):
+        self.container.config.context_menu = widget.get_active()
+
+    # return (use_opus, use_ogg)
+    def outputtype_to_options(self, type):
+        if type.upper() == 'OPUS':
```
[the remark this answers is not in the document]
dconrad5: It is not strictly necessary, but I thought it would be good to sanitize the values here. The one-liner works for me, see the newest commit.

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -463,8 +463,24 @@ def inject_extensions_preferences(self, init=False):
...
         if result:
-            for label, callback in result:
+            for label, callback, container in result:
                 page = callback()
+
+                title_display = Gtk.Label(use_markup=True, label='<b><big>{}</big></b>'.format(container.metadata.title))
+                title_display.set_halign(Gtk.Align.CENTER)
+                desc_display = Gtk.Label(use_markup=True, label=container.metadata.description)
+                desc_display.set_property('xalign', 0.0)
+                metadata_display = Gtk.Label(use_markup=True, label='\n'.join('<b>{}</b>: {}'.format(html.escape(key), html.escape(value))
```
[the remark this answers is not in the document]
elelay: Looks good. Just a couple extra colons in the parent string:

screenshot-2026-01-02-222750

auouymous:
Wouldn't it be easier to just add an info button on each extension page? It clutters the top of each page requiring users to scan down further to find the actual settings, which might require scrolling on a narrow window.

The photo displays an extension preferences page from another PR but combined with this PR it displays the same information twice, just slightly different.

The embedded info is not separated from the page contents, see youtube-dl extension.

It doubled the width of my preferences window. @tpikonen is trying to reduce the width of windows for narrow displays.

This duplicates the extension info code which leads to inconsistencies, as has already happened by renaming the field titles. It is also missing fields that appear in extension info.

I am seeing a missing ffmpeg warning on the youtube-dl page which does not appear without this patch. Is that hiding in this patch somewhere or is the patch some how triggering the warning (I have ffmpeg)?

elelay:
> Wouldn't it be easier to just add an info button on each extension page? It clutters the top of each page requiring users to scan down further to find the actual settings, which might require scrolling on a narrow window.

On the contrary I find it useful to have this information here. Also the link to the doc in a prominent manner is useful with our more complicated extensions.

> The photo displays an extension preferences page from another PR but combined with this PR it displays the same information twice, just slightly different.

The one-line description is indeed redundant with a paragraph from the help text. The help text should be adapted to not be redundant.


> The embedded info is not separated from the page contents, see youtube-dl extension.
> 
> It doubled the width of my preferences window. @tpikonen is trying to reduce the width of windows for narrow displays.

Indeed: not sure what sets a minimal width like this...

> This duplicates the extension info code which leads to inconsistencies, as has already happened by renaming the field titles. It is also missing fields that appear in extension info.

It's non obvious that a right-click should be done on the list item to get access to the "Information" popup.
Also the translated field titles and clickable doc link are an improvement.
Again, having the help link next to the preferences seems an improvement to me.

> 
> I am seeing a missing ffmpeg warning on the youtube-dl page which does not appear without this patch. Is that hiding in this patch somewhere or is the patch some how triggering the warning (I have ffmpeg)?

I'm also seeing the warning. Something must be broken...

dconrad5:
I would like to get this pull request across the finish line. My whole goal with this was not to re-vamp how the "Extensions" page displays anything, my goal was to allow a way to automatically display an extensions tab with the extension title and some useful information rather than it needing to be fully defined within the extension. Therefore, what if we limit the scope of this pull request? How about this:

  - This PR does not and will not alter the "Extensions" tab
  - I will rework this PR so that each extension tab displays the following:
    - The title of the extension
    - The existing extension description
    - the extension info via an "Extension Info" button, which will produce the existing extension info window, avoiding duplicating that code.
  - I will remove the extra "whatisthis" explanations in each of the extensions which this PR modifies

Is that an acceptable scope of this pull request? Like I said, I would really like to get this over the line, and I think the above is an achievable level of work for me at this moment.

dconrad5:
> I would like to get this pull request across the finish line. My whole goal with this was not to re-vamp how the "Extensions" page displays anything, my goal was to allow a way to automatically display an extensions tab with the extension title and some useful information rather than it needing to be fully defined within the extension. Therefore, what if we limit the scope of this pull request? How about this:
> 
>     * This PR does not and will not alter the "Extensions" tab
> 
>     * I will rework this PR so that each extension tab displays the following:
>       
>       * The title of the extension
>       * The existing extension description
>       * the extension info via an "Extension Info" button, which will produce the existing extension info window, avoiding duplicating that code.
> 
>     * I will remove the extra "whatisthis" explanations in each of the extensions which this PR modifies
> 
> 
> Is that an acceptable scope of this pull request? Like I said, I would really like to get this over the line, and I think the above is an achievable level of work for me at this moment.

Please test out the current state of this pull request, I believe it now achieves this scope of work. Is this acceptable? I can polish as requested.

elelay:
@dconrad5 thank you for reviving this PR. I would suggest #1879 instead of modifying the return value of `on_preferences` on all extensions.

dconrad5:
> @dconrad5 thank you for reviving this PR. I would suggest #1879 instead of modifying the return value of `on_preferences` on all extensions.

That certainly looks like it will work! Much cleaner than my solution too. I have rebased my commits on top of your pull request #1879 and tried to clean up my commits a lot. Hopefully it's much easier to review now.

As far as I can tell, all the extensions tabs that this commit touches now work at the minimum window width as well. I tried to make sure all labels wrap, and made some text/number boxes narrower to achieve this.

elelay:
Thank you @dconrad5.
- I've merged #1879 so please rebase again (should have not been necessary but commits in your branch are slightly different.
- Please rename `_container` to `container` in `inject_extensions_preferences`: the underscore prefix was to indicate it was unused, but it is used now ;-)

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -463,8 +463,21 @@ def inject_extensions_preferences(self, init=False):
         # add preferences buttons for all extensions
         result = gpodder.user_extensions.on_preferences()
         if result:
-            for label, callback in result:
+            for (label, callback), _container in result:
                 page = callback()
+
+                title_display = Gtk.Label(use_markup=True, wrap=True, label='<b><big>{}</big></b>'.format(_(_container.metadata.title)))
```
elelay: Please escape data coming from extensions:

```suggestion
                title_display = Gtk.Label(use_markup=True, wrap=True, label='<b><big>{}</big></b>'.format(GLib.markup_escape_text(_(_container.metadata.title))))
```

src/gpodder/gtkui/desktop/preferences.py
```diff
@@ -463,8 +463,21 @@ def inject_extensions_preferences(self, init=False):
...
+                title_display.set_halign(Gtk.Align.CENTER)
+                desc_display = Gtk.Label(use_markup=True, wrap=True, xalign=0.0, label=_(_container.metadata.description))
```
elelay: Also escape here:

```suggestion
                desc_display = Gtk.Label(use_markup=True, wrap=True, xalign=0.0, label=GLib.markup_escape_text(_(_container.metadata.description)))
```

src/gpodder/gtkui/desktop/preferences.py
elelay: (and import GLib from gi.repository)

dconrad5: You prefer this over html.escape()? I will have to add a GLib import to this file.

elelay: Yes please: `GLib.markup_escape_text` is mentioned in https://docs.gtk.org/gtk3/class.Label.html#markup-styled-text

share/gpodder/extensions/80_rename_download.py
```diff
@@ -10,13 +10,17 @@
...
+from gi.repository import Gtk
+
 logger = logging.getLogger(__name__)
 
 _ = gpodder.gettext
 N_ = gpodder.ngettext
 
 __title__ = _('Rename episodes after download')
-__description__ = _('Rename episodes to "<Episode Title>.<ext>" on download')
+__description__ = _('Rename episodes to "&lt;Episode Title&gt;.&lt;ext&gt;" on download')
```
elelay: please revert this: escaping is done in preferences.py (the description of the extension in the list doesn't use markup)

src/gpodder/gtkui/desktop/preferences.py
dconrad5: Can do. Please take a look at the newest 2 commits.

share/gpodder/extensions/80_rename_download.py
dconrad5: done

src/gpodder/gtkui/desktop/preferences.py
dconrad5: done

elelay:
Looks good to me, thank you @dconrad5.
I appreciate you used shortened labels to avoid very long items in the preference tab names.
Just fix the linting error due to the overly long lines in latest iteration and I'll approve.

@auouymous can chime in with anything I missed.

dconrad5:
> Looks good to me, thank you @dconrad5. I appreciate you used shortened labels to avoid very long items in the preference tab names. Just fix the linting error due to the overly long lines in latest iteration and I'll approve.
> 
> @auouymous can chime in with anything I missed.

Looks like it passes linting now; sorry for the noise, I can never get the linting and import sorting right the first time.

elelay:
> Looks like it passes linting now; sorry for the noise, I can never get the linting and import sorting right the first time.

No worries. Though `make lint` that can be run locally before pushing ;-)

elelay (approved):
LGTM, thanks

dconrad5:
> > Looks like it passes linting now; sorry for the noise, I can never get the linting and import sorting right the first time.
> 
> No worries. Though `make lint` that can be run locally before pushing ;-)

_That's_ what the command is!! I couldn't remember how to do it...

PR #1833, opened 2025-12-26
pyproject.toml: Update dependencies

tpikonen:
gPodder lags one minor release behind in podcastparser and requests. urllib3 has also seen several releases with the latest one containing some security fixes.

I'm not sure why requests and urllib3 are set to exact versions, this PR sets them to the latest versions above the current ones.


auouymous:
Thank you, I agree with not pinning those packages.

PR #1835, opened 2025-12-31
Replace intltool by gettext in build

tpikonen:
Intltool is de-facto deprecated and has been removed from Gnome flatpak SDK, so it now needs to built and installed before gPodder can be built for flathub.

Fortunately gettext (which the build currently depends on via intltool) now has the same functionality as intltool, so replacing it is straightforward.

I don't use a localized gPodder but the generated `po/messages.pot` at least looks OK to me. I also can't run the slightly changed windows installer scripts, so this needs testing.


elelay:
> The broken Mac build is a separate issue, I was just wondering if the localization works in Linux with this PR.

yes, make messages succeeds and the app is localized in French when launched afterwards.

tpikonen:
The Mac build is now fixed.

PR #1837, opened 2026-01-13
Narrow dialogs

tpikonen:
The 'Welcome' dialog and the 'Extension info' dialog do not currently fit narrow screens. 

In the Welcome dialog this is caused by non-wrapping button labels. Set them to wrap.

In the Extension info dialog the cause is potentially long window titles and a too large minimum width of the ScrolledWindow inside the 'show_message_details' dialog. Allow setting a shorter window title and set the minimum width smaller.

### DIFF
--- a/src/gpodder/gtkui/interface/common.py
+++ b/src/gpodder/gtkui/interface/common.py
@@ -102,9 +102,12 @@ class BuilderWidget(GtkBuilderWidget):
         widget.connect("icon-press", _on_press)
         widget.connect("populate-popup", _on_popup)
 
-    def show_message_details(self, title, message, details):
+    def show_message_details(self, title, message, details, window_title=None):
+        if window_title is None:
+            window_title = title
+
         dlg = Gtk.MessageDialog(self.main_window, Gtk.DialogFlags.MODAL, Gtk.MessageType.INFO, Gtk.ButtonsType.OK)
-        dlg.set_title(title)
+        dlg.set_title(window_title)
         dlg.set_property('text', title)
         dlg.format_secondary_text(message)
 
@@ -123,7 +126,7 @@ class BuilderWidget(GtkBuilderWidget):
         tv.set_buffer(tb)
         tv.set_property('expand', True)
         sw = Gtk.ScrolledWindow()
-        sw.set_size_request(400, 200)
+        sw.set_size_request(320, 200)
         sw.set_property('shadow-type', Gtk.ShadowType.IN)
         sw.add(tv)
         sw.show_all()
--- a/src/gpodder/gtkui/desktop/preferences.py
+++ b/src/gpodder/gtkui/desktop/preferences.py
@@ -610,7 +610,9 @@ class gPodderPreferences(BuilderWidget):
                          for key, value in container.metadata.get_sorted()
                          if key not in ('title', 'description'))
 
-        self.show_message_details(container.metadata.title, container.metadata.description, info)
+        self.show_message_details(
+            container.metadata.title, container.metadata.description, info,
+            window_title=_("Extension info"))
 
     def open_weblink(self, w, url):
         util.open_website(url)


src/gpodder/gtkui/interface/common.py
```diff
@@ -102,9 +102,9 @@ def _on_popup(_widget, menu):
         widget.connect("icon-press", _on_press)
         widget.connect("populate-popup", _on_popup)
 
-    def show_message_details(self, title, message, details):
+    def show_message_details(self, window_title, title, message, details):
```
elelay:
```suggestion
    def show_message_details(self, title, message, details, window_title=None):
      if window_title is None:
        window_title = title
```
this could be less awkward for invocations with identical `title` and `window_title`

auouymous:
There is a `message-border` style property but I'm not sure how to set it. Or we could replace the MessageDialog with a normal Dialog as I've done below and configure it however we want. Thoughts?

```python
    def show_message_details(self, title, message, details, window_title=None):
        if window_title is None:
            window_title = title

        dlg = Gtk.Dialog(window_title, self.main_window, Gtk.DialogFlags.MODAL)
        dlg.set_resizable(True)
        dlg.add_button(_('_Close'), Gtk.ResponseType.OK)

        box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
        box.set_border_width(5)
        box.set_spacing(10)

        if title:
            lbl = Gtk.Label('')
            lbl.set_markup('<span weight="bold" size="larger">%s</span>' % title)
            lbl.set_line_wrap(True)
            lbl.set_halign(Gtk.Align.START)
            lbl.set_selectable(True)
            box.add(lbl)

        if message:
            lbl = Gtk.Label(message)
            lbl.set_line_wrap(True)
            lbl.set_halign(Gtk.Align.START)
            lbl.set_selectable(True)
            box.add(lbl)

        tv = Gtk.TextView()
        tv.set_wrap_mode(Gtk.WrapMode.WORD_CHAR)
        tv.set_border_width(10)
        tv.set_editable(False)
        tb = Gtk.TextBuffer()
        tb.insert_markup(tb.get_start_iter(), details, -1)
        tv.set_buffer(tb)
        tv.set_property('expand', True)
        sw = Gtk.ScrolledWindow()
        sw.set_size_request(400, 200)
        sw.set_property('shadow-type', Gtk.ShadowType.IN)
        sw.add(tv)
        sw.show_all()
        box.add(sw)

        dlg.vbox.pack_start(box, True, True, 0)
        dlg.vbox.show_all()
        dlg.get_widget_for_response(Gtk.ResponseType.OK).grab_focus()
        dlg.run()
        dlg.destroy()
```

tpikonen:
> Or we could replace the MessageDialog with a normal Dialog 

That also works, but the width of the ScrolledWindow needs to be set smaller. The `set_size_request()` call sets the minimum size, so the dialog with a 400 px ScrolledWindow will overflow narrow screens. A typical width of a smartphone screen is 360 px (Gtk doubles the pixels) and this PR has the width set to 320 to allow some margin.

Personally I like the wide margins on the MessageDialog better on desktop, and with this PR it also (just) fits a 360 px screen.

auouymous:
> width of the ScrolledWindow needs to be set smaller.

I didn't know what size you were targeting so left it at current.


> The `set_size_request()` call sets the minimum size

It also only grows to match title width. I have a debug extension that uses this dialog and it is really cramped at 320px. The resizable flag I added here will help, but it would be nice if it could default to a little bigger on larger screens. We should make the get_screen_size function in src/gpodder/gtkui/config.py global and use it for widgets that set a size request.

```python
screen_width, screen_height = get_screen_size(Gdk.Display.get_default())
width = screen_width if screen_width < 640 else 640
sw.set_size_request(width - 20, 200)
```

We just need to pick a good upper bound around 640 or less (640 feels kind of big). The `20` represents the width of margins we choose. The MessageDialog margins are controlled by theme and therefore an unknown. We could use the custom Dialog code above to simulate the MessageDialog but give us control over the margins.

The height could also adapt but the title and message heights make that more difficult to get right.


> A typical width of a smartphone screen is 360 px (Gtk doubles the pixels)

Are you saying the phone is 720px and gtk does some hidpi to simulate 360px? Just wondering because I haven't seen a 360px device in the last 20 years. I started using gpodder on an 800x480 nokia n810 in 2007. :)

auouymous:
> Should I add the width setting per screen width to this PR?

Ya, it can all be done here, but we really should also use the custom dialog so we are in control of the margin width. Simply using a constant value that approximately represents the margin for you and I won't work for all themes. The other option is figuring out how to override that `message-border` style property for the dialog.

PR #1852, opened 2026-03-19
Add background processing to the Audio Converter extension

clayote:
The Audio Converter extension wasn't very practical to use on long episodes, because it was totally synchronous--gPodder would be unresponsive until *all* episodes queued for conversion were finished. This PR fixes it to use a thread pool to monitor the `Popen`ed subprocesses. It won't start any redundant conversion operations. The context menu entry will be hidden for episodes already queued for conversion.

I've also changed the compression options for the Opus format to be more efficient, using `-application voip`.

When the extension fails to convert a file, it will show the `ffmpeg` error in the toast notification, not only in the log file.

Removed a couple of pointless `global` lines from `soundcloud.py` to make the linter happy.

### DIFF
--- a/share/gpodder/extensions/20_audio_converter.py
+++ b/share/gpodder/extensions/20_audio_converter.py
@@ -198,7 +198,7 @@ class gPodderExtension:
             logger.warning("Error converting audio file: %s / %s", stdout, stderr)
             gpodder.user_extensions.on_notification_show(
                 _("Conversion failed"),
-                "%s: %s" % (episode.title, stderr[stderr.rfind("Error") :]),
+                "%s: %s" % (episode.title, stderr[stderr.rfind("Error"):]),
             )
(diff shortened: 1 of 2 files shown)


clayote:
A comment left by the original extension author, Bernd Schlapsi, said that he'd left out background processing on purpose, so that the user would not start the same conversion operation again. I feel that this is adequately addressed by not showing the context menu entry for the converter when the user right-clicks an episode that is already queued for conversion.

clayote:
That CI failure doesn't really look related. I've been using this extension with no issues. Merge at will.

auouymous:
The Windows build is broken, but there are a few linting errors in this PR.

clayote:
Ah, that's better. Thanks @auouymous, I didn't realize there were other checks yet to run

PR #1855, opened 2026-04-11
minor fixes

pieska:
- safeguard against empty or nonexistent entries
- use sys.executable instead of hardcoded executable name


auouymous:
Thanks for the fixes.

PR #1735, opened 2025-03-05
add config option cookiesfrombrowser

pieska:
adds a config option cookiesfrombrowser which is the same as in yt-dlp. the list is converted to a tupel with "None" translated to None.

I'm no python expert, feel free to improve it.

PR #1856, opened 2026-04-11
more options needed

pieska:
add ytdlp options (cookiesfrombrowser,force_ipv4,js_runtime,extracto_args,remote_components) needed for ytdlp-pot-providers like bgutil-ytdlp-pot-provider, otherwise downloads fail with HTTP/403


auouymous:
This was #1735, right?

#1805 added the ability to load the config file used by the yt-dlp executable, and is explained at https://github.com/gpodder/gpodder.github.io/blob/master/docs/extensions/youtubedl.md?plain=1#L70. This PR doesn't appear to do anything that can't be done in the config file, right?

pieska:
yes, it was #1735 but it's obsoleted by #1805 now

PR #1861, opened 2026-05-06
Bump github actions.

auouymous:


auouymous:
@elelay It appears msys has begun removing i686 packages such as gtk and librsvg. There are no binaries listed on https://packages.msys2.org/base/mingw-w64-gtk3 or https://packages.msys2.org/base/mingw-w64-librsvg.

The MSYS2_ARCH most likely needs to be x86_64, but MSYSTEM can be MINGW64, UCRT64 or CLANG64. According to https://www.msys2.org/news/, it appears MINGW64 is now deprecated. Should we use it until the build breaks or try one of the others?

And do you have any ideas on how to fix the mingw64-make not found error?

Issue #1753, opened 2025-04-06
Does not sync with iPod anymore AttributeError: 'iPodDevice' object has no attribute 'destination'


user82622:
GPodder suddenly stopped syncing to my iPod. Libgpod is installed, tested on Fedora 41, Fedora 40, 39 and 42 in a Toobox.

```
[gpodder.log] ERROR: Uncaught exception: Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/gpodder/syncui.py", line 108, in on_synchronize_episodes
    if not device.open():
           ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/gpodder/sync.py", line 307, in open
    self.ipod = libgpod_ctypes.iPodDatabase(self.mountpoint)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/gpodder/libgpod_ctypes.py", line 352, in __init__
    for track in glist_foreach(self.podcasts_playlist[0].members, ctypes.POINTER(Itdb_Track))]
                               ~~~~~~~~~~~~~~~~~~~~~~^^^
ValueError: NULL pointer access

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/gpodder/gtkui/main.py", line 4107, in on_sync_to_device_activate
    self.sync_ui.on_synchronize_episodes(self.channels, episodes, force_played)
  File "/usr/lib/python3.12/site-packages/gpodder/syncui.py", line 118, in on_synchronize_episodes
    device.destination.get_uri(), err.message)
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'iPodDevice' object has no attribute 'destination'

Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/gpodder/syncui.py", line 108, in on_synchronize_episodes
    if not device.open():
           ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/gpodder/sync.py", line 307, in open
    self.ipod = libgpod_ctypes.iPodDatabase(self.mountpoint)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/gpodder/libgpod_ctypes.py", line 352, in __init__
[...]

PR #1867, opened 2026-06-11
iPod sync: Allow empty / NULL string values, use ffmpeg to get track length

tpikonen:
I tested syncing to a video iPod and did not succeed with current gPodder master. This fixes the immediate problems I had, namely the exceptions caused by empty strings and wrong track lengths, if mplayer or eyed3 are not installed (but ffmpeg is).

iPod sync now works somehow, but seems to have other bugs. Unfortunately gtkpod has been dropped from recent linux distros, so fixing the FS in the iPod is tricky.

This may help with #1753, since it also seems to have something to do with NULL pointers.

### DIFF
--- a/src/gpodder/libgpod_ctypes.py
+++ b/src/gpodder/libgpod_ctypes.py
@@ -289,11 +289,17 @@ class iPodTrack(object):
         self.db = db
         self.track = track
 
-        self.episode_title = track[0].title.decode()
-        self.podcast_title = track[0].album.decode()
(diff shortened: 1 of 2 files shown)


tpikonen:
> Unfortunately gtkpod has been dropped from recent linux distros, so fixing the FS in the iPod is tricky.

There is a [snap package of gtkpod](https://snapcraft.io/gtkpod), so deleting buggy tracks etc. is still possible.

PR #1873, opened 2026-07-27
Fix ytdlp channel parsing.

auouymous:
Youtube no longer includes the unix timestamp in video listings. Instead, it uses a relative textual timestamp such as "1 day ago". Yt-dlp makes no attempt to resolve the string to a timestamp, therefore, Gpodder must fetch the video page to get the date without time and parse it to a semi-accurate timestamp assuming 00:00 for the time.

The description in video listings was clipped, but is no empty. While fetching timestamp, gpodder must also read the full description from video page.

Gpodder was not updating existing episodes with yt-dlp. This caused premiered videos, which were added without a total time, to never show one after they were eventually released. These videos are now properly updated.

### DIFF
--- a/share/gpodder/extensions/10_youtube-dl.py
+++ b/share/gpodder/extensions/10_youtube-dl.py
@@ -236,18 +236,24 @@ class YoutubeFeed(model.Feed):
         return None
 
     def get_new_episodes(self, channel, existing_guids):
+        # TODO: timestamp is not available and thus entries are unlikely to be sorted
         # entries are already sorted by decreasing date
         # trim guids to max episodes
         entries = [e for i, e in enumerate(self._ie_result['entries'])
                    if not self._max_episodes or i < self._max_episodes]
         all_seen_guids = {e['guid'] for e in entries}
+
         # only fetch new ones from youtube since they are so slow to get
-        new_entries = [e for e in entries if e['guid'] not in existing_guids]
+        # changes to existing episodes will not be fetched
+        # however, if total_time was zero (premieres), include episode for update
+        new_entries = [e for e in entries if e['guid'] not in existing_guids or existing_guids[e['guid']].total_time == 0]
+
         logger.debug('%i/%i new entries', len(new_entries), len(all_seen_guids))
         self._ie_result['entries'] = new_entries
         self._downloader.refresh_entries(self._ie_result)
+
         # episodes from entries
-        episodes = []
+        new_episodes = []
         for en in self._ie_result['entries']:
             guid = video_guid(en['id'])
             if en.get('ext'):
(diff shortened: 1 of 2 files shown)


share/gpodder/extensions/10_youtube-dl.py
```diff
@@ -259,23 +269,50 @@ def get_new_episodes(self, channel, existing_guids):
             else:
                 filesize = sum(int(f.get('filesize') or 0)
                                for f in en.get('requested_formats', []))
+            if en.get('thumbnails') and en.get('thumbnails')[0].get('url'):
+                thumbnail = re.sub(r'[?].*', '', en.get('thumbnails')[0].get('url'))
```
elelay:
```suggestion
            if thumbnail := en.get('thumbnails', [{}])[0].get('url'):
                thumbnail = re.sub(r'[?].*', '', thumbnail)
```

share/gpodder/extensions/10_youtube-dl.py
```diff
@@ -236,18 +236,28 @@ def get_http_last_modified(self):
...
         # only fetch new ones from youtube since they are so slow to get
-        new_entries = [e for e in entries if e['guid'] not in existing_guids]
+        # changes to existing episodes will not be fetched
+        # however, if total_time was zero (premieres), include episode for update
+        def get_existing_total_time(guid, existing_guids):
+            # Detect (and update) existing episode based on GUIDs
+            existing_episode = existing_guids.get(guid, None)
+            return existing_episode.total_time if existing_episode else 0
+        new_entries = [e for e in entries if get_existing_total_time(e['guid'], existing_guids) == 0]
+
```
elelay, line 250:
```suggestion
        # however, if total_time was zero (premieres), include episode for update
        new_entries = [e for e in entries if e['guid'] not in existing_guids or existing_guids[e['guid']].total_time == 0]
```
or you could also omit the existing_guids param in `get_existing_total_time`

src/gpodder/youtube.py
```diff
@@ -234,6 +247,102 @@ def youtube_get_new_endpoint(vid):
...
+            return timestamp, description
+        t = get(primary_v, 'dateText', 'simpleText')
+        if t is None:
+            t = get(primary_v, 'relativeDateText', 'simpleText')
+            if t is None:
+                return timestamp, description
+
+        # "Jan 1, 1970"
+        # "Premiered Jan 1, 1970"
+        m = re.search(r'([A-Z][a-z]{2})\s+([0-9]+),\s+([0-9]+)', t)
```
elelay:
```suggestion
        m = re.search(r'(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s+([0-9]+),\s+([0-9]+)', t)
```
let's avoid a bad month resulting in an exception in `months[m[1]]`

auouymous: The day and year values can also be invalid and hopefully cause `datetime` to throw, but the outer try catch should silently catch it and return current time. I've added your suggestion and tightened the day and year, although they can still be invalid.

share/gpodder/extensions/10_youtube-dl.py
auouymous: Thanks, it is much cleaner without the function.

share/gpodder/extensions/10_youtube-dl.py
auouymous: That is cleaner but it also changes behavior. `thumbnail` is None if the branch is not taken, and this sets it to an empty string. I changed it to use `t` for the temporary variable to prevent this.

src/gpodder/youtube.py
elelay: ah, yes. My fix was a bit foolish then. Thank you for tightening up day and year.

share/gpodder/extensions/10_youtube-dl.py
elelay: thumbnail is not initialized if the branch is not taken, causing an `UnboundLocalError: cannot access local variable 'thumbnail' where it is not associated with a value` exception (I've forced the exception by changing `en.get('thumbnails'` into `en.get('NOT_THERE'`).

auouymous: Oops, I don't know why I thought it would default to None. What do you think about the following to force None if ''?

```python
if thumbnail := en.get('thumbnails', [{}])[0].get('url'):
    thumbnail = re.sub(r'[?].*', '', thumbnail)
...
'episode_art_url': thumbnail or None,
```

elelay: looks good to me :-)

auouymous: Another oops, I don't know why I thought your assignment would return an empty string if not found, it clearly returns None. It now uses your change without my additional `or None`.

elelay: perfect

PR #1874, opened 2026-08-07
Don't use deprecated app menu

City-busz:
This ensures that the menu is visible when using client-side decoration.

When client-side decoration is used, the app menu appears on the title bar instead on the menu bar, but only if the `appmenu` is added to the window button layout in user's settings.

### DIFF
--- a/src/gpodder/gtkui/app.py
+++ b/src/gpodder/gtkui/app.py
@@ -178,7 +178,7 @@ class gPodderApplication(Gtk.Application):
                 self.add_accelerator(accel, action, None)
 
         else:
-            self.set_app_menu(self.app_menu)
+            menubar.prepend_submenu(_("_Application"), self.app_menu)
 
         Gtk.Window.set_default_icon_name('gpodder')
 


elelay:
Hi, thanks for the patch.
It is not clear to me how it is handled eg on macOS where the application menu is special.

tpikonen:
> This ensures that the menu is visible when using client-side decoration.

When I run gPodder with this patch and with `GTK_CSD=0` env var set I can see the "Application" menu in the menubar. Without this patch and `GTK_CSD=0` there seems to be no way to access the menu (in gnome-shell).

So, should the description here and in the commit message be "This ensures that the menu is visible when _not_ using client-side decoration."?

tpikonen:
> It is not clear to me how it is handled eg on macOS where the application menu is special.

If macOS can use a separate app menu, maybe we can add another branch for it to the code in question. As it is now, this patch fixes the non-visible app menu on Linux without CSD.

tpikonen:
Gtk3 has [`gtk_application_prefers_app_menu`](https://docs.gtk.org/gtk3/method.Application.prefers_app_menu.html) which has some words about macOS app menus.

elelay:
> > This ensures that the menu is visible when using client-side decoration.
> 
> When I run gPodder with this patch and with `GTK_CSD=0` env var set I can see the "Application" menu in the menubar. Without this patch and `GTK_CSD=0` there seems to be no way to access the menu (in gnome-shell).
> 
> So, should the description here and in the commit message be "This ensures that the menu is visible when _not_ using client-side decoration."?

Also please note that a non-gnome window manager, not using CSD shows the application menu (gPodder) without the patch.

PR #1880, opened 2026-09-19
Apply episode per feed limit also on gpo update + apply limit even if feed not modified

elelay:
It was applied in the GUI version but not in `gpo`.

Also apply the limit if the feed has not been modified, to apply a lowering of the limit even if the feed didn't change


elelay:
> Also apply the limit if the feed has not been modified, to apply a lowering of the limit even if the feed didn't change

@auouymous @tpikonen what do you think about this?

I found it when testing the gpo update fix, but maybe it's a _feature_ that dormant feeds are kept untouched?
For instance: when the feed has not been updated but we raise the episode limit, no old/new episodes will appear.
This is a separate bug and I will fix it by force-refresh ignoring "NOT MODIFIED" in #1808.

